PR #34, opened 2017-11-17
Support for python3 in eosdata

daritter:
The print statement is not compatible with python3 so switch to print
function and import it from the future for python 2. This should work in
python 2.6+.


dvandyk:
Manually rebased and pushed. Also prefixed the commit message. Thanks Martin!

Issue #40, opened 2018-02-06
GSL integration


fredRos:
- [x] Rename existing `integrate` to `integrate1D` to faciliate refactoring
- [x] Add interface for integrations with GSL for a given relative and absolute precision
```c++
template <typename Method_>
integrate(const std::function<double (const double&)>&, const double&a, const double& b, const Method_::Config& config);
integrate<GSL::QNG>(f, a, b, config);
```
where `Method_` is one of `GSL::QNG` (for Gauss-Kronrod) or `GSL::QAGS` (for adaptive
integration and unknown singular points). 

- [x] Convert all calls to our numerical integration to the new interface to GSL.
- [x] (Potentially) use Config class pattern
```c++
Config().relprecision(1e-3).absprecision(1e-5)....
```

fredRos:
I saw instances of calls to `gsl_monte_miser` in three dimensions. Unfortunatly cubature-based multidimensional integration is not implemented in cuba but this might be much more efficient. Example libraries that I have used for this purpose are cuba and cubature. They typically require much smaller number of evaluations for a better precision in 3D.

fredRos:
We still use the old integration routines for array integration...

fredRos:
Cubature is coming in #60. This issue was solved with #44

PR #44, opened 2018-02-13
Move to GSL integration methods

fredRos:
This is to address #40 . It's not done yet but @dvandyk may review already. In particula, I'm happy about where to put what.

While implementing, I noticed that multithreading can be an issue. Right now the integration is thread safe if the integrand is. But every call to the adaptive integration reserves and frees memory. This memory could be shared if EOS is not used in multiple threads. Else one would need a pool of gsl_memory that one thread could reuse. These are performance issues that don't afffect correctness. Profiling a real use case will show if such compilcations are called for.

### DIFF
--- a/eos/utils/integrate.cc
+++ b/eos/utils/integrate.cc
@@ -29,7 +29,7 @@ namespace eos
     using std::real;
     using std::imag;
 
-    double integrate(const std::function<double (const double &)> & f, unsigned n, const double & a, const double & b)
+    double integrate1D(const std::function<double (const double &)> & f, unsigned n, const double & a, const double & b)
     {
         if (n & 0x1)
             n += 1;
@@ -84,13 +84,13 @@ namespace eos
             std::cerr << "Q2 = " << Q2 << std::endl;
             std::cerr << "Reintegrating with twice the number of data points" << std::endl;
 #endif
-            result = integrate(f, 2 * n, a, b);
+            result = integrate1D(f, 2 * n, a, b);
         }
 
         return result;
     }
 
-    complex<double> integrate(const std::function<complex<double> (const double &)> & f, unsigned n, const double & a, const double & b)
+    complex<double> integrate1D(const std::function<complex<double> (const double &)> & f, unsigned n, const double & a, const double & b)
     {
         if (n & 0x1)
             n += 1;
@@ -145,7 +145,7 @@ namespace eos
             std::cerr << "Q2 = " << Q2 << std::endl;
             std::cerr << "Reintegrating with twice the number of data points" << std::endl;
 #endif
-            result = integrate(f, 2 * n, a, b);
+            result = integrate1D(f, 2 * n, a, b);
         }
 
         return result;
--- a/eos/utils/integrate.cc
+++ b/eos/utils/integrate.cc
@@ -20,8 +20,19 @@
 #include <eos/utils/integrate.hh>
 #include <eos/utils/matrix.hh>
 
+#include <gsl/gsl_errno.h>
+
 #include <limits>
 #include <vector>
+
+namespace
+{
+    double gsl_function_adapter(double x, void *params)
+    {
+        const auto& f = *static_cast<eos::GSL::fdd*>(params);
+        return f(x);
+    }
+}
 
 namespace eos
 {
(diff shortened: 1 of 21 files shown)


eos/utils/integrate.cc
```diff
@@ -150,4 +163,110 @@ namespace eos
...
+        gsl_function F;
+        F.function = &gsl_function_adapter;
+        F.params = (void*)&f;
+
+        // TODO possible memory leak if exception in gsl_integration_qag
+        auto ws = gsl_integration_workspace_alloc(config.limit());
+        auto status = gsl_integration_qag(&F, a, b, config.epsabs(), config.epsrel(),
+                                          config.limit(), config.key(), ws,
+                                          &result, &abserr);
+        gsl_integration_workspace_free(ws);
```
dvandyk: Allocating and freeing the workspace everytime this is called seems a bit of a waste of CPU cycles. I was wondering if one can modify ```GSL::QAGS``` as follows:
``` C++
class QAGS
{
    class Workspace
    {
         gsl_integration_workspace * _work_space;
         Workspace() :
             _work_space(gsl_integration_workspace_alloc(...))
         {
             if (! _work_space) throw IntegrationError("Could not allocate ...");
         }
         ~Workspace() { gsl_integration_workspace_free(_work_space); }
         operator=(const Workspace &) = delete;
         operator=(Workspace &&) = delete;
         operator gsl_integration_workspace * () const { return _work_space; }
    };
    static thread_local Workspace work_space;
}
```

You can then use ```QAGS::work_space``` instead of ```ws``` without any explicit allocations.

eos/utils/integrate.hh
```diff
@@ -43,6 +44,69 @@ namespace eos
...
+        class Config : public QNG::Config
+        {
+            public:
+                Config();
+
+                size_t limit() const;
+                Config& limit(const size_t&);
+
+                int key() const;
+                Config& key(const int&);
```
dvandyk: I think the ```key``` should be known at compile time. An elegant solution would be to use:
``` c++
namespace impl { class QAGSConfig { /* without the key */... }; }
template <unsigned points_> struct QAGS
{
    using Config = impl::QAGSConfig;
    static const key;
};
static const QAGS<15>::key = GSL_INTEG_GAUSS15;
...
static const QAGS<61>::key = GSL_INTEG_GAUSS61;
```

What do you think?

eos/utils/integrate.cc
```diff
@@ -20,14 +20,27 @@
...
+        return f(x);
+    }
+}
+
 namespace eos
 {
     using std::abs;
     using std::real;
     using std::imag;
+    using namespace GSL;
```
dvandyk: Why?

eos/utils/integrate.cc
```diff
@@ -20,14 +20,27 @@
...
+#include <gsl/gsl_integration.h>
+
 #include <limits>
 #include <vector>
 
+namespace
+{
+    double gsl_function_adapter(double x, void *params)
+    {
+        const auto& f = *static_cast<eos::GSL::fdd*>(params);
```
dvandyk: Are you using eos::GSL::fdd anywhere else? How about:
``` C++
template <typename Result_, typename ... Arguments_>
Result_ std_function_adapter(Arguments_ ... args, void * params)
{
    const auto & f = *static_cast<std::function<Result_ (Arguments_ ...)> *>(params);

    return f(args ...);
}
```

Completely untested!

eos/utils/integrate.cc
fredRos: Yup. `static thread_local` is the perfect fit in this case. And RAII of course solves the exception-safety issue.

eos/utils/integrate.hh
fredRos: That would work. Or with an enum

eos/utils/integrate.cc
fredRos: No, but I think we should have this type. But it doesn't have to be in `GSL` because it's more widely used. It hurts to see the same type `std::function<double (const double &)> ` repeated all over the place, e.g. in `integrate.hh`, `integrate_TEST.cpp` etc.

eos/utils/integrate.cc
fredRos: A relict... after removing, I only had to prefix `GSL::fdd`

eos/utils/integrate.hh
dvandyk: Enum is also fine, yes.

eos/utils/integrate.cc
dvandyk: I agree, the repetition is not nice.

PR #51, opened 2018-03-05
Merge changes from hackathon with Fred (March 2nd)

dvandyk:
@fredRos This includes all the changes from your branch 'for-danny'. I have adjusted the commits in the following way:

 - The way we used AC_CHECK_LIB was slightly bogus. Fixed in 3456d5d810cc2e0bf0ed1eb1f945cb06d04f49b1.
 - I introduced another layer of subdirectories by moving all Python files from ```python/``` to ```python/eos```. While I still agree that this looks nasty, it has practical use. At the time of writing this, we did not run the Python unit test, which requires setting the PYTHONPATH to the as-of-yet uninstalled modules. Of course it could not find the ```eos``` module this way. Fixed in 845515e3501649ece71f27029fdaea19eb06a211.
 - You did not add the contents of the ```python/plot``` subdirectory in the your latest commit. Can you please do that? I will then update this PR.


fredRos:
> You did not add the contents of the python/plot subdirectory in the your latest commit. Can you please do that? I will then update this PR.

I pushed to `for-danny`. Can't judge the rest of the PR now

dvandyk:
@fredRos Please merge this PR whenever you feel like it.

manual/doc/installation.tex
```diff
@@ -31,7 +31,7 @@ \section{Installing the Dependencies}
 If you intend to use the Python \cite{Python} interface to the EOS library, you will need
 to additionally install
 \begin{description}
-    \item[python3-dev] python3 interpreter and required header files,
+    \item[python3-dev] python3 interpreter in version 3.3 or higher, and required header files,
```
fredRos: I would not support 3.3. That's very old and many systems like conda forge don't support it either (I think for numpy reasons). Let's go with 3.4 or 3.5, the most recent is 3.6. In any case python 3 may be a stumbling block for mac users where the stock python is 2.7. These users would find it easiest to use conda with python 3, so we need to make sure eos plays well with conda

fredRos: Just checked, 3.5 should be enough

python/eos/plot/config.py
```diff
@@ -0,0 +1,43 @@
...
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+# Place, Suite 330, Boston, MA  02111-1307  USA
+
+import matplotlib
+from matplotlib import rcParams
+
+matplotlib.use('pgf')
```
fredRos: You should add a note in the manual like this: Import eos before you do something like `import matplotlib.pyplot as plt` because eos sets its default plot style and a matplotlib backend. All options (except the backend) can be overwritten by updating `matplotlib.rcParams[...]`. Note that the default plots use latex to create labels and math expressions, so for this to work latex needs to be available on your system.

src/scripts/eos-plot
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
fredRos: You told me that python3 is not available on fedora because `python` is python 3.x and 2.x is not available any more. Would the code break on fedora?

manual/doc/installation.tex
dvandyk: Ok:

- CentOS 7 still has only Python 2.7 in the base repository. Requires a new installation anyway. And I still see problems with libboost_python in this case.
- Python 3.{4,5,6} is available pre-built for MacOS X.

So I see no problem to go with >=3.5 here.

python/eos/plot/config.py
dvandyk: There is no section yet in the manual that discusses the Python interface at all.
I just made a new issue #54 out of your comment.

src/scripts/eos-plot
dvandyk: Sorry, I meant CentOS, see above. I do not know of anybody getting EOS to build with python support on CentOS yet. Still trying myself.

eos/Makefile.am
```diff
@@ -5,7 +5,7 @@ MAINTAINERCLEANFILES = Makefile.in
 
 SUBDIRS = parameters utils optimize statistics form-factors rare-b-decays b-decays .
 
-AM_CXXFLAGS = -I$(top_srcdir) -std=c++0x -Wall -Wextra -pedantic
+AM_CXXFLAGS = -I$(top_srcdir) -std=c++14 -Wall -Wextra -pedantic
```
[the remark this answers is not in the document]
dvandyk: Done, see 6652b7063e781cf93f9b05ae7acd3b1b996b8ef2.

configure.ac
```diff
@@ -1,6 +1,6 @@
 dnl vim: set sw=8 sts=8 noet ft=config foldmethod=marker foldmarker={{{,}}}:
 
-AC_INIT([EOS], [0.0])
+AC_INIT([EOS], [0.1])
```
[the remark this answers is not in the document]
dvandyk: Yes. I promised 1.0 to the DFG at some point. Baby steps...

.travis/build-and-test.bash
```diff
@@ -12,6 +12,8 @@ function build_and_test_ubuntu() {
         --prefix=/usr \
         || exit 1
     make distcheck -j2 DISTCHECK_CONFIGURE_FLAGS="--enable-pmc --enable-python --prefix=/usr" VERBOSE=1 || exit 1
+    make install || exit 1
```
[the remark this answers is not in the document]
dvandyk: Done, see 7a1843896e1ded9e090857afe020261ffefb1a8f.

dvandyk:
@fredRos I think I handled all your requests. Can you check?

dvandyk:
Hm, my fix using export doesn't seem to work as I imagined. ```./autogen.bash``` complains. Stay tuned.

fredRos:
This question is related, did you try that approach? https://stackoverflow.com/questions/43827664/using-same-cflags-on-makefile-am-in-each-subdirectories-sub-make

dvandyk:
The problem was modifying ```AM_CXXFLAGS``` using ```+=``` when it had been inherited from the toplevel export. The offending line was problematic anyway, and has been removed in  1e1b29f3f5deabeadb9c1ec68330514721d4655c. Now waiting for the checks to run through.

dvandyk:
Ok, ```export``` did not work as I thought. Using your suggested solution now. Waiting for the checks.

dvandyk:
Tests pass. Coverage allegedly decreases in ```eos/utils/thread_pool.cc```. I did not touch that file,
nor its test case.

README.md
```diff
@@ -2,54 +2,69 @@ EOS - A HEP Program for Flavor Observables
...
+ 3. sampling process-specific probability density functions.
 
-EOS is written in C++, and relies on as a small set of dependencies:
+An up-to-date list of EOS related publications can be found [here](https://eos.github.io/publications/).
 
- - GNU Scientific Library
- - Hierarchical Data Format v5 (HDF5)
- - Minuit2 v5 (or greater)
- - libPMC (optional)
+EOS is written in C++14, with an optional interface to Python, and depends on
```
fredRos: You write C++14 but in `configure.ac` we still check for C++0x. Then it seems it's not C++14. Can you please check?

dvandyk: ```configure.ac``` has already been changed to check for and use ```-std=c++14```.

Issue #55, opened 2018-03-09
cubature integration


fredRos:
https://github.com/stevengj/cubature
* 2D or 3D integration
* only h adaptive, p seems to require too much memory
* copy into `utils/integrate-cubature.hh` etc. 
* another template specialization of `integrate<>(...)`

dvandyk:
Resolved through 5a135bf7686dda38b805d6931aef66bfc8dfa372 .

PR #60, opened 2018-03-15
Cubature

fredRos:
First working version fixing #55 . I am happy about suggestions

### DIFF
--- a/eos/utils/integrate-cubature.hh
+++ b/eos/utils/integrate-cubature.hh
@@ -1,0 +1,117 @@
+/* Adaptive multidimensional integration of a vector of integrands.
+ *
+ * Code copied over with minor adjustments from the cubature project
+ * https://github.com/stevengj/cubature/releases/tag/v1.0.3
+ *
+ * Copyright (c) 2005-2013 Steven G. Johnson
+ *
+ * Portions (see comments) based on HIntLib (also distributed under
+ * the GNU GPL, v2 or later), copyright (c) 2002-2005 Rudolf Schuerer.
+ *     (http://www.cosy.sbg.ac.at/~rschuer/hintlib/)
+ *
+ * Portions (see comments) based on GNU GSL (also distributed under
+ * the GNU GPL, v2 or later), copyright (c) 1996-2000 Brian Gough.
+ *     (http://www.gnu.org/software/gsl/)
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+ *
+ */
+
+#ifndef CUBATURE_H
+#define CUBATURE_H
+
+#include <stdlib.h> /* for size_t */
+
+/* USAGE: Call hcubature or pcubature with your function as described
+   in the README file. */
+
+/* a vector integrand - evaluates the function at the given point x
+   (an array of length ndim) and returns the result in fval (an array
+   of length fdim).   The void* parameter is there in case you have
+   to pass any additional data through to your function (it corresponds
+   to the fdata parameter you pass to cubature).  Return 0 on
+   success or nonzero to terminate the integration. */
+typedef int (*integrand) (unsigned ndim, const double *x, void *,
+                          unsigned fdim, double *fval);
+
+/* a vector integrand of a vector of npt points: x[i*ndim + j] is the
+   j-th coordinate of the i-th point, and the k-th function evaluation
+   for the i-th point is returned in fval[i*fdim + k].  Return 0 on success
+   or nonzero to terminate the integration. */
+typedef int (*integrand_v) (unsigned ndim, size_t npt,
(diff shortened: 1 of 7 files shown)


eos/utils/integrate_TEST.cc
```diff
@@ -73,18 +73,29 @@ class IntegrateTest :
...
+            q4 = integrate<GSL::QAGS>(f4obj, 1.0, std::exp(1), config_QAGS);
             std::cout << "\\int_0.0^exp(1) f4(x) dx = " << q4 << ", eps = " << std::abs(i4 - q4) / q4 << " with QAGS" << std::endl;
             TEST_CHECK_RELATIVE_ERROR(i4, q4, eps);
+
+            auto config_cubature = cubature::Config().epsrel(eps);
+            auto f4lam = [](const std::array<double, 1> &args) -> double {
+                return f4(args[0]);
+            };
+            constexpr std::array<double, 1> a { 1.0 };
+            constexpr std::array<double, 1> b { std::exp(1) };
```
dvandyk: This is incorrect. While g++ treats ```std::exp``` as ```constexpr```, the C++ standard does not. This
causes the build failure with ```clang``` as seen by TravisCI.

eos/utils/integrate-cubature.hh
```diff
@@ -0,0 +1,123 @@
...
+
+#ifndef CUBATURE_H
+#define CUBATURE_H
+
+#include <stdlib.h> /* for size_t */
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif /* __cplusplus */
```
dvandyk: Since we do not and will not use C in the EOS source code, this (and the corresponding closing brace)
is dead weight.

eos/utils/integrate-impl.hh
```diff
@@ -114,6 +117,45 @@ namespace eos
...
     }
+
+    namespace cubature
+    {
+
+        template <size_t dim_>
+        int integrand(unsigned /* ndim */, const double *x, void *data,
+                      unsigned /* fdim */, double *fval)
+        {
+            auto& f = *static_cast<cubature::fdd<dim_> *>(data);
```
dvandyk: Since a vector-valued integrand woudl require returning e.g. an ```std::array```, this
 - should be re-labelled as ```scalar_integrand```, and
 - should accept ```fdim``` as a parameter and assert that it is indeed equal to 1,
 - should accept ```ndim``` and assert that it is equal to ```dim_```.

eos/utils/integrate-impl.hh
```diff
@@ -114,6 +117,45 @@ namespace eos
...
+            // TODO use std::experimental::to_array once it's not experimental anymore
+            std::array<double, dim_> args;
+            std::copy(x, x + dim_, args.data());
```
dvandyk: Something like an ```array_view``` would be nice here. But, alas, not before C++20...

eos/utils/integrate-impl.hh
dvandyk: I just realize that this is a hot path. Please put the checks in anyway. We can still improve performance with ```#ifdef DEBUG``` later on.

eos/utils/integrate.cc
```diff
@@ -281,6 +282,48 @@ namespace eos
         return result;
     }
 
+    namespace cubature
+    {
+        Config::Config() :
+            _qng(),
```
dvandyk: Hrm. I know why you're doing this, but it still rubs me in a wrong way. Can we rename ```QNG::Config``` to something like ```PrecisionGoal```, and use a composite for a new ```QNG::Config```? I think that makes more sense.

eos/utils/integrate.hh
```diff
@@ -120,12 +120,46 @@ namespace GSL
...
+        GSL::QNG::Config _qng;
+        size_t _maxeval;
+    };
+}
+
+    /*!
+     * Numerically integrate functions of one or more than one variable with
+     * cubature methods.
+     */
+    /// @}
```
dvandyk: Why does the group end here and not after the function? Do I misremember how doxygen works?

eos/utils/integrate_TEST.cc
```diff
@@ -73,18 +73,29 @@ class IntegrateTest :
             std::cout << "\\int_0.0^10.0 f3(x) dx = " << q3 << ", eps = " << std::abs(i3 - q3) / q3 << " over 16 points" << std::endl;
             TEST_CHECK_RELATIVE_ERROR(i3, q3, eps);
 
-            double q4 = integrate1D(std::function<double (const double &)>(&f4), 16, 1.0, std::exp(1)), i4 = 1.0;
+            auto f4obj = std::function<double (const double &)>(&f4);
+            double q4 = integrate1D(f4obj, 16, 1.0, std::exp(1)), i4 = 1.0;
```
dvandyk: Can you separate this and the following two cleanups from the cubature part, i.e., in a separate commit?

eos/utils/integrate.cc
fredRos: Sure, and it feels slightly odd. I'll refactor

eos/utils/integrate-impl.hh
fredRos: Oh, I didn't know about it. Too bad that didn't make it into C++17

eos/utils/integrate_TEST.cc
fredRos: Ah, a nasty one. Oversight by C++ standards committee? I'll fix

eos/utils/integrate-cubature.hh
fredRos: Ok. It was just copied over

eos/utils/integrate-impl.hh
fredRos: If you insist, I put in the checks. But they will be called thousands of times

eos/utils/integrate.hh
fredRos: I don't know why I put a group there, just removed it

fredRos:
I resolved all comments and pushed

PR #66, opened 2018-03-24
Get rid of templated MultivariateGaussianBlock

dvandyk:
@fredRos I would love if you could review these two commits. Tests are passing locally.


eos/statistics/log-likelihood.cc
```diff
@@ -879,87 +881,106 @@ namespace eos
...
-                        this->chol[i][j] = (j>i) ? 0 : gsl_matrix_get(chol, i, j);
+                        gsl_matrix_set(_chol, i, j, 0.0);
                     }
                 }
-                gsl_matrix_free(chol);
             }
 
             virtual ~MultivariateGaussianBlock()
             {
+                gsl_matrix_free(_covariance_inv);
```
fredRos: RAII. I can see that you don't want to start wrapping lots of GSL structs. But why don't you put them into a `unique_ptr` with a custom deleter?

eos/statistics/log-likelihood.cc
```diff
@@ -972,106 +993,94 @@ namespace eos
...
-                return L;
+                // compute inverse matrix from cholesky
+                gsl_linalg_cholesky_invert(_covariance_inv);
             }
 
+
             virtual LogLikelihoodBlockPtr clone(ObservableCache cache) const
             {
-                std::array<ObservableCache::Id, n_> indices;
+                auto k = _mean->size;
```
fredRos: Copying your nitpicking on @akokulu7 PR: why don't you declar `k const`?

eos/statistics/log-likelihood.cc
```diff
@@ -972,106 +993,94 @@ namespace eos
...
+
+                std::vector<ObservableCache::Id> ids;
 
                 // add observables to cache
-                for (auto i = 0u ; i < n_ ; ++i)
+                for (auto i = 0u ; i < k ; ++i)
                 {
-                    indices[i] = cache.add(this->cache.observable(this->ids[i])->clone(cache.parameters()));
+                    ids.push_back(cache.add(this->_cache.observable(this->_ids[i])->clone(cache.parameters())));
```
fredRos: Since you renamed `cache` to `_cache`, you can remove `this->`

eos/statistics/log-likelihood.cc
```diff
@@ -1081,43 +1090,36 @@ namespace eos
...
+                // generate standard normals in observables
+                for (auto i = 0u ; i < k ; ++i)
+                {
+                    gsl_vector_set(_observables, i, gsl_ran_ugaussian(rng));
+                }
 
-                // transform
-                random_samples = chol * random_samples;
+                // transform: observables2 <- _chol * observables
+                gsl_blas_dgemv(CblasNoTrans, 1.0, _chol, _observables, 0.0, _observables_2);
```
fredRos: Here and elsewhere: The Cholesky decomposition is a triangular matrix. Why don't you use `gsl_blas_dtrmv`?

eos/statistics/log-likelihood.cc
```diff
@@ -1362,169 +1345,19 @@ namespace eos
...
-
-            for (auto j = i + 1 ; j < n_ ; ++j)
-            {
-                covariance[i][j] = covariance[j][i] = correlation[i][j] * std::sqrt(variances[i] * variances[j]);
-            }
+            indices.push_back(cache.add(o));
         }
 
-        return MultivariateGaussian(cache, observables, mean, covariance, number_of_observations);
+        return LogLikelihoodBlockPtr(new implementation::MultivariateGaussianBlock(cache, std::move(indices), mean, covariance, number_of_observations));
```
fredRos: Nice `move`

eos/statistics/log-likelihood.cc
```diff
@@ -1362,169 +1345,19 @@ namespace eos
...
-                                             const std::array<double, 12> & mean, const std::array<std::array<double, 12>, 12> & covariance,
-                                             const unsigned & number_of_observations = 12u);
-    template LogLikelihoodBlockPtr LogLikelihoodBlock::MultivariateGaussian<16>(ObservableCache cache, const std::array<ObservablePtr, 16> & observables,
-                                             const std::array<double, 16> & mean, const std::array<std::array<double, 16>, 16> & covariance,
-                                             const unsigned & number_of_observations = 16u);
-    template LogLikelihoodBlockPtr LogLikelihoodBlock::MultivariateGaussian<18>(ObservableCache cache, const std::array<ObservablePtr, 18> & observables,
-                                             const std::array<double, 18> & mean, const std::array<std::array<double, 18>, 18> & covariance,
-                                             const unsigned & number_of_observations = 18u);
-    template LogLikelihoodBlockPtr LogLikelihoodBlock::MultivariateGaussian<19>(ObservableCache cache, const std::array<ObservablePtr, 19> & observables,
```
fredRos: I'm glad all these explicit instantiations are gone. They were a pain to add

eos/statistics/log-likelihood.hh
```diff
@@ -255,12 +259,22 @@ namespace eos
...
              *
-             * @note For every dimension, this template and the corresponding implementation
-             *       have to be instantiated explicitly.
+             * @param cache         The Observable cache from which we draw the predictions.
+             * @param observables   The Observables whose distribution we model.
+             * @param mean          The vector of means.
+             * @param covariance    The covariance matrix
+             */
+            static LogLikelihoodBlockPtr MultivariateGaussian(ObservableCache cache, const std::vector<ObservablePtr> & observables,
+                    gsl_vector * mean, gsl_matrix * covariance, const unsigned & number_of_observations);
```
fredRos: I don't like this style `gsl_vector * mean`, it looks like a product to me. I'd suggest `gsl_vector* mean` or `gsl_vector *mean`. But that's general thing in the code

eos/statistics/log-likelihood.cc
dvandyk: I have commit for this somewhere. Going to recycle it.

eos/statistics/log-likelihood.hh
dvandyk: I prefer it actually. Keeping it :-)

eos/statistics/log-likelihood.cc
dvandyk: Ok, it's messier than expected. Either I'd have one header that pulls in basically all of GSL, or many separste ones. I will make this a separate issue to discuss the solution.

eos/statistics/log-likelihood.cc
dvandyk: Done.

eos/statistics/log-likelihood.cc
dvandyk: I like the additional ```this->``` to make perfectly clear what is the present instance and what is not.

eos/statistics/log-likelihood.cc
dvandyk: In principal a good idea. However, present GSL deprecated the ```gsl_linalg_cholesky_decomp``` function, and I cannot find out easily right now what the correct output is (upper/lower diagonal).
Might be a good idea to open a new issue to keep track of this #70.

eos/statistics/log-likelihood.cc
dvandyk: Me too!

eos/statistics/log-likelihood.cc
fredRos: The output should be lower triangular. But ok, we follow up on this in #70

PR #71, opened 2018-03-27
Preparing to load constraints from YAML files

dvandyk:
@fredRos this my plan to implement loading constraints from YAML files, similar to how we currently handle parameters.

With this first version of the PR, we produce sensible output for Gaussian constraints, which looks like
this:
``` YAML
Lambda_b->Lambdamu^+mu^-::BR[15.0,20.0]@LHCb-2015B:
    type: Gaussian
    observable: Lambda_b->Lambdall::BR@LowRecoil
    kinematics: {s_min: 15, s_max: 20}
    options: {l: mu}
    mean: 6e-07
    sigma-stat: {hi: 4.5e-08, lo: -4.5e-08}
    sigma-sys: {hi: 4.5e-08, lo: -4.5e-08}
    dof: 1
```
All other constraints currently produce this:
``` YAML
B_s->K^*::V@MILC-2013A:
    type: unimplemented
```

My plan is to 
- [x] implement serialization for all existing constraint types,
- [x] implement deserialization for ```Gaussian```
- [x] implement deserialization for ```LogGamma```
- [x] implement deserialization for ```MultivariateGamma(correlation)```
- [x] implement deserialization for ```MultivariateGamma(covariance)```
- [x] implement deserialization for ```Amoroso```
- [x] remove entries for everything else
- [x] add unit tests that compare serialized(deserialized) output with the input
- [x] add unit tests to check value of generated LogLikelihoodBlocks for correctness
- [x] move all remaining constraints to decay/process-specific YAML files

Let me know what you think. You can access the YAML dump via ```eos-list-constraints -y```.


fredRos:
> My plan is to
>    implement serialization for all existing constraint types,
you mean to/from YAML? Sounds good
>    move all existing constraints to decay/process-specific YAML files
A good idea. That would make it much more organized

eos/utils/kinematic.cc
```diff
@@ -34,6 +42,8 @@ namespace eos
         std::map<std::string, unsigned> variables_map;
 
         std::vector<std::string> variables_names;
+
+        std::vector<KinematicVariable> variables;
```
fredRos: Do we really need this extra variable? What about `variables_data`?

eos/utils/kinematic.cc
```diff
@@ -54,7 +64,7 @@ namespace eos
                 _imp->variables_map[v->first] = index;
```
fredRos: Can you turn the C++03 loop into a range-based for loop, please?

eos/constraint.cc
```diff
@@ -162,6 +175,38 @@ namespace eos
 
             return os;
         }
+
+        virtual void serialize(YAML::Emitter & out) const
```
fredRos: If this is, what should we do with the arcane `as_string` methods in `LogLikelihood`? Do you want to update them, too? Or just keep it as it is because it's working?

fredRos:
One more thing: the constraint include the source of the numbers. But you should explain (say at the top of the output) how to use this as a key into `references.yaml` or `references.json` (which one actually?)

fredRos:
Sorry, I clicked on approve but I actually want some minor changes

eos/utils/kinematic.cc
dvandyk: We do have wrappers for forward iterators of type ```T_```, and indirect iterators that convert ```T_ *``` and ```std::shared_ptr<T_>``` iterators to iterators for ```T_```. We do not have iterators that
accept a lambda to turn 'nothing' into ```KinematicVariable```. This is also not costing much, since
it simply means one ```std::shared_ptr<>``` more per kinematic variable. If you have a better idea, I'm all ears!

eos/constraint.cc
dvandyk: ```as_string``` is used in the existing data dumps to HDF. So better keep it for now.

eos/utils/kinematic.cc
dvandyk: In line 57? Sure can do!

dvandyk: Done.

dvandyk:
> One more thing: the constraint include the source of the numbers. But you should explain (say at the top of the output of eos-list-constraints -y) how to use this as a key
For the time being, this is purely internal and meant to reduce the workload of converting existing constraints into a YAML-based format.

> into references.yaml or references.json (which one actually?)
There is only one, isn't there? ```references.yaml```. There is no JSON used anywhere in EOS anymore,
since it does not permit comments.

eos/constraint.cc
fredRos: Ok

eos/utils/kinematic.cc
fredRos: Alright. We can still improve this letter if there is a better idea. I don't have one now

dvandyk:
@fredRos I think I have completed the serialization part. Can you please check that I did not forget anything or messes something up?

fredRos:
Hmm, there are no tests for the YAML output. I guess simple tests would be to serialize and deserialize. When do you want to implement the deserialization?

Another point I just noticed: What about dumping some of the Amoroso constraints? I think we don't need them and they cause maintenance problem now. I would just keep the general form and one convenience method. I see that there is lots of code duplication for Amorosi in `constraint.cc`, too.

dvandyk:
> - [ ] add unit tests to check value of generated LogLikelihoodBlocks for correctness

@fredRos Any chance that I can convince you to have a go at that? I think it would suffice
to generate YAML inputs for (some of) the existing test cases in ```eos/statistics/log-likelihood_TEST.cc```,  and check for the outcome of the likelihood. Luckily by now we can create pseudoobservables for any ```Parameter``` through ```Observable::make```.

dvandyk:
@fredRos No need anymore, I did it already. Tests based on the values of the LLH created by the Amoroso, Gaussian, LogGamma, and the two MultivariateGaussian constrainted entries have been added.

PR #73, opened 2018-04-15
Begin Jupyter integration

dvandyk:
@fredRos Have a look [here](https://github.com/dvandyk/eos/blob/6c8d2baccfa3584e201ddb6bfe4e6795e6182be2/examples/jupyter-uncertainty.ipynb) for a display of the notebook.

### DIFF
--- a/python/eos/__init__.py
+++ b/python/eos/__init__.py
@@ -51,3 +51,15 @@ from _eos import *
 
 from .data import *
 from .plot import *
+
+try:
+    if __IPYTHON__:
+        ip = get_ipython()
+        html_formatter = ip.display_formatter.formatters['text/html']
+
+        from .ipython import __format_Parameter, __format_KinematicVariable, __format_Observable
+        html_formatter.for_type(Parameter, __format_Parameter)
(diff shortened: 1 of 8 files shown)


dvandyk:
@fredRos Gentle ping :smile: Please review and rebase+merge, if you have the time.

python/eos/__init__.py
```diff
@@ -51,3 +51,15 @@ def FromWCxf(wc):
 
 from .data import *
 from .plot import *
+
+try:
+    if __IPYTHON__:
```
fredRos: Use similar `try ... except` to avoid `/usr/lib/python3/dist-packages/matplotlib/__init__.py:1352: UserWarning:  This call to matplotlib.use() has no effect
because the backend has already been chosen;
matplotlib.use() must be called *before* pylab, matplotlib.pyplot,
or matplotlib.backends is imported for the first time.`

that gets displayed in notebook

dvandyk: That works nicely. I have disabled the selection of the PGF backend for any iPython based interpreter.

PR #88, opened 2018-05-11
[eos] Implement B->{D^*,rho} form factor parametrizations following [BSZ2015]

akokulu7:


eos/form-factors/mesonic-impl.hh
```diff
@@ -590,7 +610,7 @@ namespace eos
 
             static std::string _par_name(const std::string & ff_name)
             {
-                return std::string("B->K^*::alpha^") + ff_name + std::string("@BSZ2015");
+                return std::string(Process_::label) + std::string("::alpha^") + ff_name + std::string("@BSZ2015");
```
dvandyk: @akokulu7 Would you be so kind to split this change into a separate earlier commit? This is a true bug fix, even though it did not affect anything before now.

eos/form-factors/mesonic-impl.hh
```diff
@@ -53,6 +54,25 @@ namespace eos
...
+        static constexpr const char * label = "B->D^*";
+        static constexpr double mB = 5.279;
+        static constexpr double mV = 2.0103;
+        static constexpr double mBc = 6.2751;
+        static constexpr double mR2_0m = (mBc + 0.000) * (mBc + 0.000);
+        static constexpr double mR2_1m = (mBc + 0.056) * (mBc + 0.056);
+        static constexpr double mR2_1p = (mBc + 0.492) * (mBc + 0.492);
+    };
+
+    struct BToRho {
```
dvandyk: Can you move the ```BToRho``` class below ```BToKstar```, such that we keep the classes orderd alphabetically? It helps finding things later on.

eos/form-factors/mesonic.cc
```diff
@@ -34,6 +35,14 @@ namespace eos
...
     /* B_{u,d} -> K^* */
 
+    constexpr double BToDstar::mR2_0m;
+    constexpr double BToDstar::mR2_1m;
+    constexpr double BToDstar::mR2_1p;
+
+    constexpr double BToRho::mR2_0m;
+    constexpr double BToRho::mR2_1m;
+    constexpr double BToRho::mR2_1p;
+
```
dvandyk: See my comment above on the ordering.

eos/form-factors/mesonic.cc
```diff
@@ -138,6 +147,8 @@ namespace eos
             { KeyType("B->K^*",     "KMPW2010"), &KMPW2010FormFactors<PToV>::make           },
             { KeyType("B->K^*",     "BFW2010"),  &BFW2010FormFactors<BToKstar, PToV>::make  },
             { KeyType("B->K^*",     "BSZ2015"),  &BSZ2015FormFactors<BToKstar, PToV>::make  },
+            { KeyType("B->D^*",     "BSZ2015"),  &BSZ2015FormFactors<BToDstar, PToV>::make  },
+            { KeyType("B->rho",     "BSZ2015"),  &BSZ2015FormFactors<BToRho, PToV>::make    },
```
dvandyk: And here again: Please insert ```"B->rho"``` in alphabetical order.

eos/form-factors/mesonic_TEST.cc
```diff
@@ -759,3 +760,161 @@ class BToKstarBSZ2015FormFactorsTest :
...
+        TEST_CHECK_NEARLY_EQUAL(0.307107, ff->t_2(6.1), eps);
+
+        TEST_CHECK_NEARLY_EQUAL(0.179298, ff->t_3(0.1), eps);
+        TEST_CHECK_NEARLY_EQUAL(0.196067, ff->t_3(2.1), eps);
+        TEST_CHECK_NEARLY_EQUAL(0.215543, ff->t_3(4.1), eps);
+        TEST_CHECK_NEARLY_EQUAL(0.238325, ff->t_3(6.1), eps);
+    }
+} b_to_rho_bsz2015_form_factors_test;
+
+class BToDstarBSZ2015FormFactorsTest :
```
dvandyk: Ah, in this case, please move up the B->D^* test case above the others.

eos/form-factors/mesonic_TEST.cc
```diff
@@ -759,3 +760,161 @@ class BToKstarBSZ2015FormFactorsTest :
...
+    virtual void run() const
+    {
+        static const double eps = 5.1e-3;
+
+        Parameters p = Parameters::Defaults();
+        std::shared_ptr<FormFactors<PToV>> ff = FormFactorFactory<PToV>::create("B->D^*@BSZ2015", p);
+        TEST_CHECK(ff.get() != nullptr);
+
+        /*
+         * for the B->D^* case no SSE parameters exist yet, so for the moment we set them to zero and once the GKvD2018 calculation is fitted one can then simply update this test case accordingly.
```
dvandyk: @akokulu7 that's a good idea, thanks! We can then update the test case later on!

eos/observable.cc
```diff
@@ -184,6 +185,86 @@ namespace eos
...
+            make_observable("B->D^*::A_2(s)/A_1(s)", "B->D^*",
+                            &FormFactors<PToV>::a_2, &FormFactors<PToV>::a_1),
+
+            make_observable("B->D^*::A_12(s)/A_1(s)", "B->D^*",
+                            &FormFactors<PToV>::a_12, &FormFactors<PToV>::a_1),
+
+            make_observable("B->D^*::T_23(s)/T_2(s)", "B->D^*",
+                            &FormFactors<PToV>::t_23, &FormFactors<PToV>::t_2),
+
+            // B -> rho Form Factors
```
dvandyk: Again, please move B->rho below B->K^*, such that alphabetical order is maintained.

eos/parameters/form-factors.yaml
```diff
@@ -673,6 +673,12 @@
     min:     -0.105
     max:     +0.165
     comments: ""
+# a value for B->K^*::alpha^A12_0 was missing previously, set it to zero default value for the moment
```
dvandyk: @akokulu7 You will find that the form factor class does not use alpha^A12_0 at all. The reason is that the A12 form factor at q2 = 0 can be expressed through the A0 form factor; similarly also for T23. See [here](https://github.com/eos/eos/blob/v0.1.1/eos/form-factors/mesonic-impl.hh#L613) for the initialization of the parameters, and [here](https://github.com/eos/eos/blob/v0.1.1/eos/form-factors/mesonic-impl.hh#L660-L668) for the use in the A12 form factor.

Please remove this piece.

eos/parameters/form-factors.yaml
```diff
@@ -713,6 +719,12 @@
     min:     +1.095
     max:     +1.695
     comments: ""
+# a value for B->K^*::alpha^T2_0 was missing previously, set it to zero default value for the moment
+"B->K^*::alpha^T2_0@BSZ2015" :
```
dvandyk: As above for A12_0: please remove this piece.

eos/parameters/form-factors.yaml
```diff
@@ -820,3 +832,219 @@
     min:     0.75
     max:     1.25
     comments: "GeV^2"
+
+# No valid form-factor parameters for B->D* exist yet, so we leave everything at zero for the moment.
```
dvandyk: exist -> are available

akokulu7:
Changes are implemented accordingly. @dvandyk , could you please merge now?

dvandyk:
@akokulu7 I think something went awry in your first commit message, it is garbled a bit.
Also, try to be concise with it. Mentioning ```std::string``` is not helpful to understand what the bug was (i.e., implicitly always loading B->K^* form factor parameters in BSZ2015).

akokulu7:
@dvandyk , ah sorry, right, I didn't notice that before. What would you suggest then to write as a commit title for this specific bug fix?

dvandyk:
How about '[form-factors] Fix bug in parameter specifications for [BSZ2015] form factors' ?

akokulu7:
thanks @dvandyk , the last changes are implemented and the PR should now be ready for "merge" to the base.

PR #96, opened 2018-05-17
propagate uncertainty to use MCMC samples

### LINKED ISSUE #82: eos-propagate-uncertainty should be able to use MCMC samples

fredRos:
To fix #82

### DIFF
--- a/src/clients/eos-propagate-uncertainty.cc
+++ b/src/clients/eos-propagate-uncertainty.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2011, 2012 Frederik Beaujean
+ * Copyright (c) 2011, 2012, 2018 Frederik Beaujean
  *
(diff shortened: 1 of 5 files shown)


src/clients/eos-propagate-uncertainty.cc
```diff
@@ -82,14 +83,21 @@ class CommandLine :
...
         CommandLine() :
             config(PriorSampler::Config::Default()),
             parameters(Parameters::Defaults()),
             pmc_sample_min(0),
             pmc_sample_max(0),
-            pmc_sample_directory("/data")
+            pmc_sample_directory("/data"),
+            mcmc_sample_min(0),
+            mcmc_sample_max(0),
+            mcmc_prefer_prerun(false)
```
dvandyk: Shouldn't this default to ```true```?

fredRos: I don't think so. The prerun by definition is expected to show bias. So if you have main run and prerun, the default should be to take samples from the main run.

PR #102, opened 2018-05-25
[infra] Fix problems with non-standard Minuit2 installations

EduRom7:


dvandyk:
@greenwald FYI, first commit by E. Romero.

PR #111, opened 2018-07-06
Add B->D^* form factor contraints from the lattice

akokulu7:


eos/constraints/B-to-V-form-factors.yaml
```diff
@@ -1,5 +1,20 @@
 # vim: set fdm=marker :
 
+#########################
+# B -> D^* form factors #
+#########################
+# {{{
+B->D^*::A_1[s_max]@HDW-2017:
```
dvandyk: This is from the paper by the HPQCD collaboration, correct? I think it's a better idea
to name it HPQCD-2017.

PR #112, opened 2018-07-18
Specify root environment for installing dependencies

EduRom7:
Added `sudo` in the dependencies installation step.


dvandyk:
@EduRom7 We do not use merge commits in EOS. Please try

```
   git checkout -b manual origin/master
   git cherry-pick ea2c6d4    # this pick your new commit and applies it on top of origin/master
```

Also, the commit message does not follow the usual rules. Since only the ```manual/```
subdirectory is affected, please prefix the message with '[manual] '. You can do this
using ```git commit --amend```. Thanks!

PR #67, opened 2018-03-24
Fix building in environment with non-standard yaml-cpp location

dvandyk:
@daritter Please check if this works for you, and approve.

PR #116, opened 2018-09-18
[infra] Fix various problems with non-standard locations

daritter:
I tried to compile the latest master with all dependencies in non-standard locations (and custom compiled boost). This lead to a variety of problems

* `boost_python_suffix` was not determined correctly
* abi flag was missing from python library (I only have `libpython3.6m`)
* GSL includes were not found as GSL was in non-standard location
* yaml-cpp was not found but the commit from pr #67 worked fine after rebasing
* missing `-lhdf5` leading to undefined reference

This pr fixes all of these issues for me but since I'm not an expert on the internal structure there might be better ways to do this.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -100,12 +100,40 @@ AC_CHECK_LIB([boost_system], main,
 	)
 dnl }}}
 dnl {{{ yaml-cpp
+PKG_CHECK_MODULES([YAMLCPP],
+	[yaml-cpp],
+	[
+		YAMLCPP_CXXFLAGS="$YAMLCPP_CFLAGS"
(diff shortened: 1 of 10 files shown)


daritter:
@dvandyk I have no idea what's wrong with the darwin build ... it seems to fail before every reaching eos compilation

dvandyk:
Thanks Martin, this all looks good. I have fixed the CI error on Darwin by now. Can you rebase this branch on top of master and force push? I would like to see this build on Darwin before merging. Thanks!

daritter:
Hi Danny, my pleasure

PR #117, opened 2018-09-28
Apply the BCL2008 parametrization for B->D,K transitions to include f_T

akokulu7:


eos/form-factors/mesonic-impl.hh
```diff
@@ -776,7 +776,7 @@ namespace eos
         static constexpr const double m2_Br0p = 6.420 * 6.420; // B_c scalar
         static constexpr const double tau_p = (m_B + m_P) * (m_B + m_P);
         static constexpr const double tau_m = (m_B - m_P) * (m_B - m_P);
-        static constexpr const bool uses_tensor_form_factors = false;
+        static constexpr const bool uses_tensor_form_factors = true;
```
dvandyk: Ok, this is the correct way to add f_T to the B->D form factor class.

eos/form-factors/mesonic-impl.hh
```diff
@@ -797,6 +797,7 @@ namespace eos
              */
             UsedParameter _f_plus_0, _b_plus_1, _b_plus_2;
             UsedParameter            _b_zero_1, _b_zero_2;
+            UsedParameter _f_t_0,    _b_t_1,    _b_t_2;
```
dvandyk: This is wrong. You are modifying the class that is used when ```uses_tensor_form_factors``` is set to ```false```.
You do not need to modify any code, beyond the change in line 779.

eos/form-factors/mesonic-impl.hh
```diff
@@ -840,12 +844,21 @@ namespace eos
                 return _f_plus_0 / (1.0 - s / Process_::m2_Br0p) * (1.0 + _b_zero_1 * zbar + _b_zero_2 * z2bar);
             }
 
-            virtual double f_t(const double &) const
```
dvandyk: See above. Please revert this part.

eos/form-factors/mesonic-impl.hh
```diff
@@ -816,7 +817,10 @@ namespace eos
                 _b_plus_1(p[std::string(Process_::label) + "::b_+^1@BCL2008"],  *this),
                 _b_plus_2(p[std::string(Process_::label) + "::b_+^2@BCL2008"],  *this),
                 _b_zero_1(p[std::string(Process_::label) + "::b_0^1@BCL2008"],  *this),
-                _b_zero_2(p[std::string(Process_::label) + "::b_0^2@BCL2008"],  *this)
```
dvandyk: See above, please revert this part.

eos/form-factors/mesonic-impl.hh
dvandyk: Please revert this part.

eos/parameters/form-factors.yaml
```diff
@@ -468,6 +473,58 @@
...
+    max:      0.0
+    comments: ""
+"B->D::b_T^2@BCL2008" :
+    central:  0.0
+    min:      0.0
+    max:      0.0
+    comments: ""
+
+# form factor parameters for B->K for [BCL2008] parametrization
+"B->K::f_+(0)@BCL2008" :
```
dvandyk: Sorry, is this needed? Isn't there already a BCL2008 parameter set for B->K transitions?

eos/form-factors/mesonic-impl.hh
akokulu7: i see, okay reverted.

eos/form-factors/mesonic-impl.hh
akokulu7: reverted.

eos/form-factors/mesonic-impl.hh
akokulu7: reverted.

eos/parameters/form-factors.yaml
akokulu7: @dvandyk , it seems like it is needed. in line 394 it says "# form factor parameters for B->pi, B->K in [BCL2008] parametrization, ..."  however there was no BCL2008 parameter set for B->K transitions before (only for B->pi was included before), that is why I added them.  In addition, there is the following problem with eos when evaluating the f_T form factor for B->K case:

"eos-evaluate --kinematics s 4 --observable "B->K::f_T(s);form-factors=BCL2008"
Caught exception: 'No form factor found for process 'B->K' and tag 'BCL2008'!'"

while, it does evaluates f_T for pi and D. Why is it so, could you please try this?

eos/form-factors/mesonic-impl.hh
akokulu7: okay, please see the new commit ef8613f.

akokulu7:
@dvandyk please see 
5b68b4a and 5e2babb, the issue should be resolved by those.

dvandyk:
@akokulu7 Can you rebase on top of the current master branch and force push here? I would like to see the tests cases to pass, in particular on Darwin.

akokulu7:
@dvandyk when i try to rebase (git rebase gh/master) it says "Current branch master is up to date". my  present branch is already 2 commits ahead of the eos:master, so I don't get what do you mean?

dvandyk:
Sorry my fault. I did not see your new push with the CI-related commit. Something went awry with the TravisCI tests. I re-started those test jobs. Let's see what happens. If the PR is not updated with the test results I can always forcefully merge this PR.

akokulu7:
@dvandyk perfect! thanks.

PR #118, opened 2018-10-01
Generic NP contribution to B->Dlnu including tensor operator effects

akokulu7:


dvandyk:
@cBobeth Christoph, any chance you could review this one?

cBobeth:
- have checked expression of normalized differential decay with with eq.(8) in 1309.0301 
-> seems ok
- there are hard-coded integration boundaries for the integrated R(D), might replace them by using actual values of B- and D-meson masses and lepton masses, unless these are given by experiment. Still in this case there should be a reference where these integration boundaries come from.

akokulu7:
@cBobeth thanks for the review! regarding the integration boundaries for R(D), they are actually determined using the actual values of the masses. For example the integration over the full PS region [m_{l}^2, (m_B-m_D)^2] translates to [0.011164, 11.62] bin when using the actual masses (for muon as the default), i think once this "choice" is specified it should be clear then, no? or do you have any other suggestion?

cBobeth:
- But the B- and D-meson masses change slightly, depending on option for spectator quark q = u or d?! Also, when the user changes the lepton masses to some different value, again this is not taken care of. I would not put fixed values there, but just use m_B, m_D and m_l variables to calculate the boundaries?!

- Of course you could also introduce a second function for a (partially) integrated R(D) where the user can give arbitrary integration boundaries (s_min_mu, s_max_mu, s_min_tau, s_max_tau), not necessarily the full phase space in q^2

dvandyk:
@akokulu7 If it's not to much of a hassle, I would prefer to have a three-variable observable, i.e., depending on q^2_mu,min, q^2_tau,min, q^2_max. A separate commit would be a good idea for adding this observable.

akokulu7:
@dvandyk good! I have already done this and will soon do a separate commit for it.

akokulu7:
@cBobeth and @dvandyk  can you please see the latest commit, what do you think?

cBobeth:
- It looks fine to me. If the user really needs a separate s_max for mu and tau, he can easily add it.

- Maybe still a separate function call without any arguments for the full phase space would be welcome by users, which could call your current function just calculating s_min_mu, s_min_tau and s_max from the actual values m_l, m_B, m_D used in the instantiation of the user ("this->")

akokulu7:
@cBobeth okay. but s_max is independent of the leptons so why should we have two s_max?

cBobeth:
Two separate s_max might be needed if experiments measure the branching fractions with different binning for mu and tau channel.

akokulu7:
okay. Beside, via this three argument construction i think the user can easily now adjust in what part of the PS region they want to compute R_D, so in my opinion another function would not be necessary. @dvandyk if you are fine with this could pls you merge this PR, cheers

dvandyk:
As long as the physics description is correct, we can merge. Adding further (sub)observables is always possible.

PR #122, opened 2018-10-04
Generic NP contribution to B->D^*lnu including tensor operators effects

akokulu7:
@dvandyk and @cBobeth can you please review this PR, cheers.

### DIFF
--- a/eos/b-decays/b-to-d-l-nu.cc
+++ b/eos/b-decays/b-to-d-l-nu.cc
@@ -110,7 +110,12 @@ namespace eos
             const complex<double> cs2 = wc.csl();
             const complex<double> cT  = wc.ct();
             // normalized(|V_cb|=1) differential decay width including NP (in SM cvl=1, and all other couplings are zero)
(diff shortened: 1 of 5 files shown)


cBobeth:
- I have checked with 1309.0301 eq.(9) and could not find any issues
- Please do not write everything on a single line here https://github.com/eos/eos/blob/f388569afd3595d9742e164b959ca9021bf73ad5/eos/b-decays/b-to-dstar-l-nu.cc#L122 some line-breaks would be helpful when comparing, maybe as in eq.(9).
- One point, which concerns also B->D l v_l', is that in 1309.0301 they have given the result for fixed letpon flavor l = tau, but the flavor of the neutrino is still l' = e, mu, tau arbitrary. Therefore Wilson coefficient C_V1^l' always comes in combination with delta_{l' tau}. Of course experimentally they measure the sum over the three neutrino flavors l'. What you have programmed is just the branching ratio for l' = tau. Maybe we should write this somewhere explicitely. Otherwise of course we should introduce Wilson coefficients that carry a neutrino-flavor index etc. and then give branching ratios that include the summation over the neturino-flavor index.

Issue #132, opened 2018-11-24
Enable eos-plot to plot observable as functions of a parameter


dvandyk:
Presently, ```Plotter.plot_observable``` checks for the presence of the key ```kinematic```, using
its value as the name of a kinematic variable. When discussing with Stephan, I had proposed to
use two mutually exclusive keys: ```kinematic``` and ```parameter```. If both are set, issue an error.
If neither are set, issue an error. If ```kinematic``` is set, create ```var``` as a ```KinematicVariable```
via the ```Kinematics``` object. If ```parameter``` is set, obtain ```var``` as a ```Parameter``` via the ```Parameters``` object. The interfaces are identical for kinematic variables and parameters.

@gubernari and @SKuerten are already working on this.

PR #133, opened 2018-11-26
Fix issue #132

SKuerten:


dvandyk:
@skuerten Failed with
> KeyError: 'neither kinematic nor parameter found; do not know how to map x to a variable'

There seems to be a logical error in the checks.

PR #141, opened 2019-04-04
[manual] corrected definition of dipole operators for b->s  +  some m…

cBobeth:
Please have a look


dvandyk:
Hm, this will not apply since it contains a merge commit. Can you clean that up? (You merged rather than rebasing, and we require a linear commit history!)

dvandyk:
The change looks fine, thanks @cBobeth!

PR #142, opened 2019-04-19
Full angular distribution for B->D^*lnu

akokulu7:
@cBobeth could you please review this PR? cheers.

### DIFF
--- a/eos/b-decays/b-to-dstar-l-nu.cc
+++ b/eos/b-decays/b-to-dstar-l-nu.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2018 Ahmet Kokulu
+ * Copyright (c) 2018, 2019 Ahmet Kokulu
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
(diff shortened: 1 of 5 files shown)


eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -19,18 +19,183 @@
...
+            inline double v6()      const   { return _vv[5]; }
+            inline double v7()      const   { return _vv[6]; }
+            inline double v8()      const   { return _vv[7]; }
+            inline double v9()      const   { return _vv[8]; }
+            inline double v10()     const   { return _vv[9]; }
+            inline double v11()     const   { return _vv[10];}
+            inline double v12()     const   { return _vv[11];}
+
+            // longitudinal polarization amplitude
+            inline double normalized_amplitude_polarization_L() const
```
cBobeth: Maybe do not call them "normalized_amplitude_polarization_L" - they do not seem real observables according to eq.(9) in 1405.3719. Please introduce instead 
    Gamma_L = 3/4 A_L
    Gamma_R = 3/4 A_T
the decay width for longitudinal and transversal polarized D^*

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -19,18 +19,183 @@
...
+            }
+
+            // transverse polarization amplitude
+            inline double normalized_amplitude_polarization_T() const
+            {
+                return  2.0 * (vv1T() - vv2T() / 3.0);
+            }
+
+            // redefined decay width
+            inline double normalized_decay_width() const
```
cBobeth: Do not call it "normalized_decay_width", but just decay width and define it as

  Gamma = Gamma_L + Gamma_T

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -93,146 +282,221 @@ namespace eos
...
-    BToDstarLeptonNeutrino::differential_branching_ratio(const double & s) const
-    {
-        return _imp->differential_branching_ratio(s);
-    }
+    /* q^2-differential observables */
 
+    // |Vcb|=1
     double
-    BToDstarLeptonNeutrino::integrated_branching_ratio(const double & s_min, const double & s_max) const
+    BToDstarLeptonNeutrino::normalized_differential_branching_ratio(const double & s) const
```
cBobeth: What means "normalized" here? Maybe drop this?

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -19,18 +19,183 @@
...
+            inline double vv2T()    const  { return _vv[4]; }
+            inline double vv3T()    const  { return _vv[5]; }
+            inline double vv4T()    const  { return _vv[6]; }
+            inline double vv5T()    const  { return _vv[7]; }
+            inline double vv10T()   const  { return _vv[8]; }
+            inline double vv20T()   const  { return _vv[9]; }
+            inline double vv30T()   const  { return _vv[10];}
+            inline double vv40T()   const  { return _vv[11];}
+
+            inline double v1()      const   { return _vv[0]; }
```
cBobeth: Could not find why you need these definitions, can they be removed?

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -77,11 +241,36 @@ namespace eos
 
             u.uses(*form_factors);
             u.uses(*model);
+
+        }
+
+        // normalization according to DSD2014
```
cBobeth: maybe add also equation number when possible:
according to DSD2014 eq.(7)

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -77,11 +241,36 @@ namespace eos
...
+        double norm(const double & s) const
+        {
+            // charged lepton velocity in the dilepton rest frame
+            double v    = (1.0 - m_l * m_l / s);
+            double lam  = lambda(m_B * m_B, m_Dstar * m_Dstar, s);
+            double p    = std::sqrt(lam) / (2.0 * m_B);
+            // normalized prefactor (|Vcb|^2=1)
+            return power_of<2>(g_fermi()) * p * s * power_of<2>(v) / (3.0 * 64.0 * power_of<3>(M_PI) * m_B * m_B);
```
cBobeth: OK

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -93,146 +282,221 @@ namespace eos
...
+            // charged lepton velocity in the dilepton rest frame
+            double v        = (1.0 - m_l * m_l / s);
+            double m_l_hat  = std::sqrt(1.0 - v);
+            double lam      = lambda(m_B * m_B, m_Dstar * m_Dstar, s);
+
+            double sqrts = std::sqrt(s);
+            double NF = norm(s);
+
+            // transversity amplitudes A's. cf. [DSD2014]
+            result.a_0          = (1.0 - gA) * (m_B + m_Dstar) / (2.0 * m_Dstar * sqrts) * ( (m_B * m_B - m_Dstar * m_Dstar - s) * aff1 - lam * aff2 / power_of<2>(m_B + m_Dstar) );
```
cBobeth: Ok

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -40,35 +205,34 @@ namespace eos
...
+            model(Model::make(o.get("model", "SM"), p, o)),
+            parameters(p),
+            opt_q(o, "q", { "u", "d" }, "d"),
+            hbar(p["hbar"], u),
+            tau_B(p["life_time::B_" + opt_q.value()], u),
+            g_fermi(p["G_Fermi"], u),
+            opt_l(o, "l", {"e", "mu", "tau"}, "mu"),
+            m_l(p["mass::" + opt_l.value()], u),
+            m_B(p["mass::B_" + opt_q.value()], u),
+            m_Dstar(p["mass::D^*_" + opt_q.value()], u),
```
cBobeth: I think this is wrong here, because if
 
  barB^0 -> D^+ l^- \bar{v}_l
  B^- -> D^0 l^- \bar{v}_l

You can not use the same opt_q value for B and D !?

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -19,18 +19,183 @@
...
+            inline double a_t_2() const
+            {
+                return  vv30T() / (normalized_amplitude_polarization_L() + normalized_amplitude_polarization_T());
+            }
+
+            inline double a_t_3() const
+            {
+                return  vv40T() / (normalized_amplitude_polarization_L() + normalized_amplitude_polarization_T());
+            }
+        };
```
cBobeth: Observables ok, but please add the 4-fold angular distribution in eq.(6) of DSD2014

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -19,18 +19,183 @@
...
+            double v;
+            double NF;
+        };
+
+        // angular observables V's. cf. from [DSD2014], redifined V's in order to include NF
+        struct AngularObservables
+        {
+            std::array<double, 12> _vv;
+
+            AngularObservables(const Amplitudes & a)
```
cBobeth: OK

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -93,146 +282,221 @@ namespace eos
...
-        return integrate<GSL::QAGS>(f, s_min, s_max);
+        return _imp->differential_angular_observables(s).a_fb_leptonic();
     }
 
     double
-    BToDstarLeptonNeutrino::differential_r_d(const double & s) const
+    BToDstarLeptonNeutrino::differential_ratio_tau_mu(const double &s) const
     {
-        double br_muons;
+        double br_tau;
```
cBobeth: Is it sufficient here to save only m_l? Should'nt you also save the opt_l, because
this will decide whether you are using the "b->c mu nu_mu" or the "b->c tau nu_tau"
new physics Wilson coefficients?! Maybe in your test cases you should also use
different NP values for mu and tau Wilson coefficients and not the same.

eos/b-decays/b-to-dstar-l-nu.cc
dvandyk: Yes! Please use the format
´´´
cf. [REFERENCE], eq. (XX), p. YY
´´´
or
´´´
cf. [REFERENCE], eqs. (XX), p. YY
´´´

eos/b-decays/b-to-dstar-l-nu.cc
dvandyk: Christoph and I just discussed this. Using the same option is indeed correct, and Christoph's comment was in error.

eos/b-decays/b-to-dstar-l-nu.cc
dvandyk: I think we can merge w/o the 4-fold distribution and add it later on!?

eos/b-decays/b-to-dstar-l-nu.cc
dvandyk: Does SaveOption work with SwitchOption yet?

eos/signal-pdf.cc
```diff
@@ -125,6 +126,11 @@ namespace eos
                     Options{ },
                     KinematicRange{ "s", 0.0, 11.62, BToDLeptonNeutrino::kinematics_description_s }),
 
+            make_signal_pdf("B->D^*munu::dBR",
```
cBobeth: How you make sure that the lepton flavour is really "muons"? 
Is this the default setting?

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: @cBobeth @dvandyk thanks for the review! yes, v1, ...v12 are not needed in this case and I will remove. I will try to do the corrections/modifications soon and push for a merge.

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: okay

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: okay, as i thought.

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: i have no idea on this. @cBobeth If this is needed could you explicitly type me down the line that is to be added below that line? According to numerics that i get from eos, it seems to me that saving only m_l is sufficient and it does distinguish between the Wilson coefficients's flavors too.

eos/signal-pdf.cc
akokulu7: the default lepton in eos is muon.

cBobeth: Maybe we can put it there explicitly not relying on the setting of default values,
maybe they get changed one day, although in this case very unlikely

eos/b-decays/b-to-dstar-l-nu.cc
cBobeth: When the Wilson coefficients are requested by the command:

const WilsonCoefficients<BToC> wc = model->wilson_coefficients_b_to_c(opt_l.value(), false);

then there is explicitly "opt_l" required to specify the lepton-type,
so I wonder how this can be correct if you only save the value
of "m_l" and not the one of "opt_l"?

cBobeth: Please see integrated R_Dstar, where it is correct and add the lines

Save<std::string> save_opt_l(_imp->opt_l._value, "mu");

Save<std::string> save_opt_l(_imp->opt_l._value, "tau");

as you have done there

akokulu7: i think, I have already another line saving opt_l too  it is just below the line saving m_l;

Save<std::string> save_opt_l(_imp->opt_l._value, "tau");

this is done for the integrated ratio level, but not at the differential ratio level.  So do you simply mean to include this also in the differential level too?

akokulu7: okay I included the missing "Save<std::string> save_opt_l(_imp->opt_l._value, "tau");" and "Save<std::string> save_opt_l(_imp->opt_l._value, "mu");" worked. thanks.

eos/b-decays/b-to-dstar-l-nu_TEST.cc
```diff
@@ -186,31 +132,48 @@ class BToDstarLeptonNeutrinoTest :
...
                 p3["b->cmunumu::Im{cVR}"]         = -2.0;
                 p3["b->cmunumu::Re{cSL}"]         = +3.0;
                 p3["b->cmunumu::Im{cSL}"]         = -3.0;
                 p3["b->cmunumu::Re{cSR}"]         = +4.0;
                 p3["b->cmunumu::Im{cSR}"]         = -4.0;
-                // switch on the tensor op. contr. here
                 p3["b->cmunumu::Re{cT}"]          = +5.0;
                 p3["b->cmunumu::Im{cT}"]          = -5.0;
+                // tau mode
+                p3["b->ctaunutau::Re{cVL}"]       = +1.0;
```
cBobeth: As last thing: Could you put here in tau-NP Wilson coefficients different
numerical values then you are using above for mu-NP Wilson coefficients,
because in the way it is now, you might would not been able to catch the
mistake with the missing "opt_l" that you had in the differential R_Dstar

akokulu7: Yes, i see what you mean. I will do this too, I will let you know when I push my newer changes for a last look before merging to the base branch.

eos/signal-pdf.cc
akokulu7: so I keep it as it is, or did you mean to include something else here?

cBobeth: Yes, could you please put there explicitly that the lepton flavor is muon, I am not
sure how it works, maybe you have to put something in

  Options{{"l", "mu"}}

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: done!

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: @cBobeth and @dvandyk  I have now included also the 4-fold angular distribution, cheers

eos/signal-pdf.cc
akokulu7: done! and Options{{"l", "mu"}} seems to work.

eos/b-decays/b-to-dstar-l-nu_TEST.cc
akokulu7: done!

akokulu7:
@cBobeth could you pls have a quick look at this version of the PR before Danny merges it. cheers.

eos/b-decays/b-to-dstar-l-nu.cc
```diff
@@ -93,146 +269,272 @@ namespace eos
...
+
+    //* cf. [DSD2014], eq. (6), p. 5
+    double
+    BToDstarLeptonNeutrino::four_differential_decay_width(const double & s, const double & c_theta_l, const double & c_theta_d, const double & phi) const
+    {
+        // compute d^4 Gamma, cf. [DSD2014], p. 5, eq. (6)
+        // Trigonometric identities: Cosine squared of the angles
+        double c_theta_d_2 = c_theta_d * c_theta_d;
+        double c_theta_l_2 = c_theta_l * c_theta_l;
+        double c_phi = cos(phi);
```
cBobeth: Can you use here 

  "cos" -> "std::cos" 
  "sin" -> "std::sin"
  "sqrt" -> "std::sqrt"

unless you have written somewhere

  using std::cos, std::sin, std::sqrt;

eos/b-decays/b-to-dstar-l-nu.hh
```diff
@@ -39,25 +39,41 @@ namespace eos
 
             // Differential Observables
             double differential_branching_ratio(const double & s) const;
+            double differential_a_fb_leptonic(const double & s) const;
+            double differential_ratio_tau_mu(const double & s) const;
 
             // Differential Observables - normalized(|Vcb|=1)
             double normalized_differential_branching_ratio(const double & s) const;
 
+            // Four Differential Observables - normalized(|Vcb|=1)
```
cBobeth: Please call it

  normalized_four_differential_decay_width

to remind the user that |Vcb|=1 !

eos/b-decays/b-to-dstar-l-nu.cc
akokulu7: done!

eos/b-decays/b-to-dstar-l-nu.hh
akokulu7: done! and i think the PR should now be ready for merge. cheers

PR #148, opened 2019-05-20
Two-fold distribution for B->D l nu

akokulu7:
@cBobeth could you please review this pull request, cheers.

### DIFF
--- a/eos/b-decays/b-to-d-l-nu.cc
+++ b/eos/b-decays/b-to-d-l-nu.cc
@@ -3,7 +3,7 @@
 /*
  * Copyright (c) 2015, 2016, 2017 Danny van Dyk
  * Copyright (c) 2015 Marzia Bordone
- * Copyright (c) 2018 Ahmet Kokulu
+ * Copyright (c) 2018, 2019 Ahmet Kokulu
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -80,7 +80,58 @@ namespace eos
             u.uses(*model);
         }
 
-        // normalized to V_cb = 1
+        //* normalized(|Vcb|=1) two-fold-distribution, cf. [DSD2014], eq. (12), p. 6
+        double normalized_two_differential_decay_width(const double & s, const double & c_theta_l) const
+        {
+            //  d^2 Gamma, cf. [DSD2014], p. 6, eq. (13)
+            // Trigonometric identities
+            double c_theta_l_2 = c_theta_l * c_theta_l;
+            double s_theta_l_2 = 1.0 - c_theta_l_2;
+            double s_theta_l = std::sqrt(s_theta_l_2);
+            double c_2_theta_l = 2.0 * c_theta_l_2 - 1.0;
+            double s_2_theta_l = 2.0 * s_theta_l * c_theta_l;
+
+            double fp = form_factors->f_p(s);
+            double f0 = form_factors->f_0(s);
+            double fT = form_factors->f_t(s);
+            // running quark masses
+            double mbatmu = model->m_b_msbar(mu);
+            double mcatmu = model->m_c_msbar(mu);
+
+            double lam = lambda(m_B * m_B, m_D * m_D, s);
+            double p = sqrt(lam) / (2.0 * m_B);
+            // make sure we return NaN if s < m_l^2, v = lepton velocity in the dilepton rest frame
+            double v = (1.0 - m_l * m_l / s);
+            double v2 = v * v;
+            double ml_hat = sqrt(1.0 - v);
+            double nD = v2 * m_B * s * power_of<2>(g_fermi()) / (256.0 * power_of<3>(M_PI * m_B));
+            // NP contributions in EFT including tensor operator (cf. [DSD2014]). (in SM cvl=1, and all other couplings are zero)
+            const WilsonCoefficients<BToC> wc = model->wilson_coefficients_b_to_c(opt_l.value(), false);
+            const complex<double> vl = wc.cvl() - 1.0;
+            const complex<double> vr = wc.cvr();
+            const complex<double> sl = wc.csl();
+            const complex<double> sr = wc.csr();
+            const complex<double> gV = vr + vl;
+            const complex<double> gA = vr - vl;
+            const complex<double> gS = sr + sl;
+            const complex<double> gP = sr - sl;
+            const complex<double> tl = wc.ct();
+            // helicity amplitudes, cf. [DSD2014] eqs. 13-14
+            const complex<double> hh0 = 2.0 * m_B * p * fp * (1.0 + gV) / sqrt(s) ;
+            const complex<double> hht = (1.0 + gV) * (m_B * m_B - m_D * m_D) * f0 / sqrt(s);
+            const complex<double> hhS = - gS * (m_B * m_B - m_D * m_D) * f0 / (mbatmu - mcatmu);
+            const complex<double> hhT = - 2.0 * m_B * p * fT * tl / (m_B + m_D);
+            const complex<double> hhtS = hht - hhS / ml_hat;
+
+            double result = 2.0 * nD * p * (
+                                            std::norm(hh0) * s_theta_l_2 + (1.0 - v) * power_of<2>(std::abs(hh0) * c_theta_l - std::abs(hhtS))
+                                            + 8.0 * ( ((2.0 - v) + v * c_2_theta_l) * std::norm(hhT) - ml_hat * std::real(hhT * (std::conj(hh0) - std::conj(hhtS) * c_theta_l)))
+                                            );
+
+            return result;
+        }
+
+        // normalized to V_cb = 1, obtained using cf. [DSD2014], eq. (12), agrees with Sakaki'13 et al cf. [STTW2013]
         double normalized_differential_decay_width(const double & s) const
         {
             double fp = form_factors->f_p(s);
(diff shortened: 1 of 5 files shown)


eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -80,7 +80,58 @@ namespace eos
             u.uses(*model);
         }
 
-        // normalized to V_cb = 1
+        //* normalized(|Vcb|=1) two-fold-distribution, cf. [DSD2014], eq. (13), p. 6
+        double normalized_two_differential_decay_width(const double & s, const double & c_theta_l) const
+        {
+            //  d^2 Gamma, cf. [DSD2014], p. 6, eq. (13)
+            // Trigonometric identities
+            double c_theta_l_2 = c_theta_l * c_theta_l;
```
cBobeth: From here on to line 124 is code used multiple times in these three functions 

normalized_two_differential_decay_width
normalized_differential_decay_width
numerator_differential_a_fb_leptonic

Please create some struct that calculates these quantities for "s" and
stores them, so you can call it and use them whenever you calculate
an observable - much like in B->D* l v

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -80,7 +80,58 @@ namespace eos
...
+            // helicity amplitudes, cf. [DSD2014] eqs. 13-14
+            const complex<double> hh0 = 2.0 * m_B * p * fp * (1.0 + gV) / sqrt(s) ;
+            const complex<double> hht = (1.0 + gV) * (m_B * m_B - m_D * m_D) * f0 / sqrt(s);
+            const complex<double> hhS = - gS * (m_B * m_B - m_D * m_D) * f0 / (mbatmu - mcatmu);
+            const complex<double> hhT = - 2.0 * m_B * p * fT * tl / (m_B + m_D);
+            const complex<double> hhtS = hht - hhS / ml_hat;
+
+            double result = 2.0 * nD * p * (
+                                            std::norm(hh0) * s_theta_l_2 + (1.0 - v) * power_of<2>(std::abs(hh0) * c_theta_l - std::abs(hhtS))
+                                            + 8.0 * ( ((2.0 - v) + v * c_2_theta_l) * std::norm(hhT) - std::sqrt(1.0 - v) * std::real(hhT * (std::conj(hh0) - std::conj(hhtS) * c_theta_l)))
```
cBobeth: Can in principle replace

  std::sqrt(1.0 - v)   ->  ml_hat

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -80,7 +80,58 @@ namespace eos
             u.uses(*model);
         }
 
-        // normalized to V_cb = 1
+        //* normalized(|Vcb|=1) two-fold-distribution, cf. [DSD2014], eq. (13), p. 6
```
cBobeth: Maybe replace

  eq. (13)   ->   eq.(12) 

???

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -178,6 +275,42 @@ namespace eos
         return integrate<GSL::QAGS>(f, s_min, s_max);
     }
 
+    double
+    BToDLeptonNeutrino::differential_a_fb_leptonic(const double & s) const
+    {
```
cBobeth: Cant you just simply write for the function body:

  return _imp->numerator_differential_a_fb_leptonic(s)/_imp->normalized_differential_decay_width(s);

???

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -210,26 +343,30 @@ namespace eos
...
+    The decay B->D l nu, where l=e,mu,tau is a lepton.";
 
     const std::string
     BToDLeptonNeutrino::kinematics_description_s = "\
-The invariant mass of the l-nubar pair in GeV^2.";
+    The invariant mass of the l-nubar pair in GeV^2.";
+
+    const std::string
+    BToDLeptonNeutrino::kinematics_description_c_theta_l = "\
+    The cosine of the charged lepton's helicity angle theta_l in the l-nubar rest frame.";
```
cBobeth: Is it clear that "helicity angle" is meant w.r.t. the direction of the B-meson?
Maybe you can give more precise definition?

eos/b-decays/b-to-d-l-nu_TEST.cc
```diff
@@ -140,30 +94,41 @@ class BToDLeptonNeutrinoTest :
...
-                const double eps = 1e-5;
+                const double eps = 1e-3;
 
-                // test for different s-bins - the default lepton is muon
-                TEST_CHECK_RELATIVE_ERROR(d.normalized_integrated_branching_ratio(4.0, 8.0), 1074.2803668903907, eps);
-                TEST_CHECK_RELATIVE_ERROR(d.normalized_integrated_branching_ratio(8.0, 11.62), 706.6625032909665, eps);
-                // the full phase-space region for muon
-                TEST_CHECK_RELATIVE_ERROR(d.normalized_integrated_branching_ratio(0.011164, 11.62), 2461.2412433585673, eps);
+                // the default lepton is muon
+                TEST_CHECK_RELATIVE_ERROR(d.normalized_integrated_branching_ratio(0.011164, 11.62), 2581.58, eps);
```
cBobeth: I dont understand how the normalized integrated BR can be so large?

   |V_cb|^2 ~ (0.042)^2 ~ 0.00176

then the normal BR

  BR = 0.00176 * 2581.58 ~ 4.55 

becomes larger than 1. How can this be? Did I missunderstand something?

eos/b-decays/b-to-d-l-nu.cc
cBobeth: I think it should be the direction of the D, not the B

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -89,48 +140,87 @@ namespace eos
...
+            const complex<double> gP = sr - sl;
+            const complex<double> tl = wc.ct();
+            // helicity amplitudes, cf. [DSD2014] eqs. 13-14
+            const complex<double> hh0 = 2.0 * m_B * p * fp * (1.0 + gV) / sqrt(s) ;
+            const complex<double> hht = (1.0 + gV) * (m_B * m_B - m_D * m_D) * f0 / sqrt(s);
+            const complex<double> hhS = - gS * (m_B * m_B - m_D * m_D) * f0 / (mbatmu - mcatmu);
+            const complex<double> hhT = - 2.0 * m_B * p * fT * tl / (m_B + m_D);
+            const complex<double> hhtS = hht - hhS / ml_hat;
+
+            return 4.0 * nD * p * ( std::abs(hh0) * std::abs(hhtS) * (1.0 - v) - 4.0 * ml_hat * std::real(hhT * std::conj(hhtS)) );
```
cBobeth: Should'nt there be a minus sign in front of

  std::abs(hh0) * std::abs(hhtS) * (1.0 - v)

???

It comes from the term in (eq.12)

  (|H0| cosTh - |Hts| )^2 = |H0|^2 cosTh^2 - 2 |H0| |Hts| cosTh + |Hts|^2

and in the AFB only the one ~ cosTh enters

cBobeth: Maybe we could agree how you define the A_FB:

 A_FB = int_0^1 d^2Gamma - int_-1^0 d^2Gamma

is what I am used to. Then you are missing an overall minus sign, i.e.:

  4.0 * nD * p * ...   ->  - 4.0 * nD * p * ...

If you define 

  A_FB = int_-1^0 d^2Gamma - int_0^1 d^2Gamma

then it is ok, but for me this is an unusual definition.

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -89,48 +140,87 @@ namespace eos
...
+            const complex<double> tl = wc.ct();
+            // helicity amplitudes, cf. [DSD2014] eqs. 13-14
+            const complex<double> hh0 = 2.0 * m_B * p * fp * (1.0 + gV) / sqrt(s) ;
+            const complex<double> hht = (1.0 + gV) * (m_B * m_B - m_D * m_D) * f0 / sqrt(s);
+            const complex<double> hhS = - gS * (m_B * m_B - m_D * m_D) * f0 / (mbatmu - mcatmu);
+            const complex<double> hhT = - 2.0 * m_B * p * fT * tl / (m_B + m_D);
+            const complex<double> hhtS = hht - hhS / ml_hat;
+
+            // normalized(|V_cb|=1) differential decay width
+            return 4.0 / 3.0 * nD * p * ( std::norm(hh0) * (3.0 - v) + 3.0 * std::norm(hhtS) * (1.0 - v) + 16.0 * std::norm(hhT) * (3.0 - 2.0 * v) - 24.0 * ml_hat * std::real(hhT * std::conj(hh0)) );
```
cBobeth: OK

eos/observable.cc
```diff
@@ -611,17 +611,33 @@ namespace eos
...
 
+            make_observable("B->Dlnu::normBR",
+                            &BToDLeptonNeutrino::normalized_integrated_branching_ratio,
+                            std::make_tuple("s_min", "s_max")),
+
             make_observable("B->Dlnu::R_D(s)",
                             &BToDLeptonNeutrino::differential_r_d,
                             std::make_tuple("s")),
 
             make_observable("B->Dlnu::R_D",
```
cBobeth: Maybe you can use the new way of creating an observable that is actually
a ratio of two observables. Danny has pushed somewhere an example,
I think it is for R_Dstar(s).

eos/b-decays/b-to-d-l-nu.cc
dvandyk: This should be fixed in a separate commit basically everywhere.

eos/b-decays/b-to-d-l-nu_TEST.cc
dvandyk: @cBobeth This is a NP test case, which has the Wilson coefficients set to unphysically large values.

eos/observable.cc
dvandyk: We have a separate issue for this, which I assigned to @gubernari already :-)

eos/b-decays/b-to-d-l-nu.cc
akokulu7: @cBobeth @dvandyk thanks. but can you have a look at p.3 of https://arxiv.org/abs/1405.3719 and Fig.1,  where they define the angles (for B->D*lnu). there it says ".. theta_l is the polar angle between the charged lepton and the direction 'opposite' to D* meson in the l-nu rest frame..." . So why should it be different definition when considering the B->D case, or? i will have a look at other comments later, cheers

dvandyk: @akokulu7 Christoph's point was, the the present definition of theta_l in the description is incomplete. And as I wrote: this is not a necessary change for this PR. @cBobeth can e.g. go over the descriptions of all angular distributions to fix it.

eos/observable.cc
cBobeth: Ok

eos/b-decays/b-to-d-l-nu_TEST.cc
cBobeth: Indeed, if NP were so large, then the lifetime would be modified for sure in some way, such that BR will be smaller than 1 again, if one takes this into account.

eos/b-decays/b-to-d-l-nu.cc
cBobeth: As Danny says, the question was more or less, whether you would know what is meant by "helicity angle" - I would not understand it. So we have to check the literature for B->D,D* l v,
because I am just used to the definitions in B->K,K* ll. I will add a separate issue for this.

dvandyk:
@akokulu7 can you please revise this branch on top of master? There is a limit there fixing the test failure on MacOSX.

akokulu7:
@dvandyk , @cBobeth the PR is revised and can be merged.

eos/observable.cc
```diff
@@ -2229,7 +2245,7 @@ namespace eos
                     std::make_tuple("s_min", "s_max")),
 
 	    make_observable("Lambda_b->Lambdall::M_28@LowRecoil",
-                    &LambdaBToLambdaDilepton<LowRecoil>::integrated_m28,
+                    &LambdaBToLambdaDilepton<LowRecoil>::integrated_m8,
```
cBobeth: Are you sure about this change? Seems like an accidental deleting ?!

eos/b-decays/b-to-d-l-nu.cc
```diff
@@ -89,48 +140,87 @@ namespace eos
...
+            const complex<double> gP = sr - sl;
+            const complex<double> tl = wc.ct();
+            // helicity amplitudes, cf. [DSD2014] eqs. 13-14
+            const complex<double> hh0 = 2.0 * m_B * p * fp * (1.0 + gV) / sqrt(s) ;
+            const complex<double> hht = (1.0 + gV) * (m_B * m_B - m_D * m_D) * f0 / sqrt(s);
+            const complex<double> hhS = - gS * (m_B * m_B - m_D * m_D) * f0 / (mbatmu - mcatmu);
+            const complex<double> hhT = - 2.0 * m_B * p * fT * tl / (m_B + m_D);
+            const complex<double> hhtS = hht - hhS / ml_hat;
+
+            return - 4.0 * nD * p * ( std::abs(hh0) * std::abs(hhtS) * (1.0 - v) - 4.0 * ml_hat * std::real(hhT * std::conj(hhtS)) );
```
cBobeth: OK

eos/b-decays/b-to-d-l-nu.cc
cBobeth: You did not remove the code duplication?!

akokulu7: i tried something for this but it didn't work.., so i prefer to keep it this way, unless you want to do this before merging.

eos/observable.cc
akokulu7: ah thanks for pointing this out to me. i don't remember touching this at all but it looks accidental as you said. i corrected this now. cheers

eos/b-decays/b-to-d-l-nu.cc
cBobeth: Should'nt be similar to B->D^* lv where you have the struct of the amplitudes?

akokulu7:
@cBobeth perhaps it is something simple but i can't figure it out at the moment.., so yes pls merge as it is, thanks.
btw: i see that all travis tests passes, but it still shows the status as "pending check"

dvandyk:
Hm, seems to have been stuck. I have restarted one of the tests, and it now flags as 'All checks have passed'.

kshumm:
DID YOU RESPECTED PEOPLE complete YOUR PROJECT WITH B_>Dlvbar.i am stuck in the angles for lbar and vbar .

Issue #153, opened 2019-05-23
Adapt B->P and B->V LCSRs to different B-mesons


dvandyk:
- [x] add a spectator quark option 'q' to ```BLCDAs``` in ```eos/form-factors/b-lcdas.cc```;
 - [x] split the declaration of structs within the ```lcsrs``` namespace in ```eos/form-factors/analytic-b-to-v-lcsr.hh``` into a forward declaration (e.g. ```struct BToRho;``` in the same file) and a full declaration (```struct BToRho { ... };``` in the .cc file);
 - [x] split the file ```eos/form-factors/analytic-b-to-v-lcsr.cc``` into one file per process, to avoid overly long recompilation times;
 - [x] split the declaration of structs within the ```lcsrs``` namespace in ```eos/form-factors/analytic-b-to-p-lcsr.hh``` into a forward declaration (e.g. ```struct BToPi;``` in the same file) and a full declaration (```struct BToPi { ... };``` in the .cc file);
 - [x] split the file ```eos/form-factors/analytic-b-to-p-lcsr.cc``` into one file per process, to avoid overly long recompilation times;
 - [x] ass mass Phi and decay constants for Phi and D_s
 - [x] add B_s -> phi, B_s -> D^(*) form factors.

dvandyk:
@gubernari is keeping track of this in gubernari/eos@issue153

dvandyk:
Fixed as of a3c6605. Thanks @gubernari!

PR #156, opened 2019-06-04
Fix issue #153

dvandyk:
@gubernari says this does not pass all tests yet. Let's have a look....


dvandyk:
```
libtool: compile:  clang++-4.0 -DHAVE_CONFIG_H -I. -I../../../../eos/form-factors -I../.. -I../../../.. -std=c++14 -Wall -Wextra -pedantic -O2 -g -march=x86-64 -std=c++14 -std=c++14 -MT libeosformfactors_la-mesonic.lo -MD -MP -MF .deps/libeosformfactors_la-mesonic.Tpo -c ../../../../eos/form-factors/mesonic.cc  -fPIC -DPIC -o .libs/libeosformfactors_la-mesonic.o
../../../../eos/form-factors/mesonic.cc:24:10: fatal error: 
      'eos/form-factors/analytic-b-to-kstar.hh' file not found
#include <eos/form-factors/analytic-b-to-kstar.hh>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
@gubernari seems that you did not remove this ```#include``` statement

dvandyk:
Although it is flagged as 'All checks have failed', the last job has only 'failed' due to exceeding the time limit. Before being killed, it had started building a debian package, which is the last step. I will split the 16.06 job into two shorter ones (compile + test, and compile + build ```.deb``` file).

PR #160, opened 2019-06-07
Fix some bugs when installing in virgin Ubuntu installations

dvandyk:
1. The binary packages were missing a runtime dependency on ```libyaml-cpp0.5v5``` on both xenial and bionic.
2. Debian/Ubuntu do not use 'site-packages' for the installation of distribution-provided Python modules and extensions. Instead, they use 'dist-packages'. Changed the ```$(pythondir)``` and ```$(pyexecdir)``` variables on Debian/Ubuntu only.


dvandyk:
Thanks to @daritter for private comments on the Python directories in the Belle 2 Externals. This seems to work, so I'm going to merge.

Issue #167, opened 2019-06-20
Add script to output MCMC modes on a per-chain basis


SKuerten:
I am using the plotScript.py in the private scripts repository. This script works with python 2.7 a script working with python 3.6 would be useful.

PR #168, opened 2019-06-20
Fix issue #167: Add eos-print-modes script

dvandyk:
@SKuerten Does this work for you?

Fixes: #167


SKuerten:
There were some issues:
* the number of elements in the list were wrong, now it correctly counts them.
* commas in between the results of one mode were not conform to other scripts.
* I added output for the global mode

dvandyk:
@SKuerten did you push it to gh-dvandyk/issue167 ?

PR #170, opened 2019-06-23
Fix issue #166: Organize parameters in sections and groups

dvandyk:
I kept all parameter names as they are, but we should seriously move to using ```QualifiedName``` also for parameter names.


eos/parameters/sm-and-eft.yaml
```diff
@@ -0,0 +1,866 @@
...
+          central: 0.0
+          min:     0.0
+          max:     0.0
+      'b->ctaunutau::Im{cT}' :
+          central: 0.0
+          min:     0.0
+          max:     0.0
+  # }}}
+
+  # b -> c l^- nubar WET Parameters
```
cBobeth: Should be "b -> u" instead of "b -> c"

eos/parameters/sm-and-eft.yaml
```diff
@@ -0,0 +1,866 @@
...
+  # {{{
+  - title: '$b\to c\ell^-\bar\nu$ WET Parameters'
```
cBobeth: b->u

eos/parameters/sm-and-eft.yaml
```diff
@@ -0,0 +1,866 @@
...
+    description: >
+        The list of parameters describing the $b\to c\ell^-\bar\nu$ Wilson coefficients.
```
cBobeth: b->u

eos/parameters/form-factors.yaml
```diff
@@ -1,2548 +1,2558 @@
...
-    central: 0.0
-    min:     0.0
-    max:     0.0
-"B->pipi::c^Ftime_1_2@FvDV2018" :
-    central: 0.0
-    min:     0.0
-    max:     0.0
+  # Parameters in LCSR Calculations with Pion LCDAs
+  # {{{
+  - title: 'Parameters in LCSR Calculations with Pion LCDAs'
```
cBobeth: Pions and also K^* = light meson LCDA's ?

dvandyk:
@cBobeth thanks!

PR #178, opened 2019-07-11
Added updated Lb -> L mu mu BF number using only TeVatron normalisation

tblake00:
Added an updated value for Lambda_b->Lambdall::BR@LowRecoil to Lb-to-L-ll.yaml using only TeVatron production information. Updates the LHCb 2015 branching fraction number.


dvandyk:
@tblake00 Thanks Tom! Can I convince you to amend the commit to read "[constraints] Adapt ..."? Reasoning is that things are very much interdependent, and it helps to see from the commit message which part of EOS has been changed. One of these days I should write a "how to contribute" document...

dvandyk:
Passed before renaming, no need to run the CI again here. Thanks Tom!

PR #183, opened 2019-08-06
[utils] Make scale dependence explicit in wilson_coefficiens_b_to_s

gubernari:


dvandyk:
Problem with the MacOSX VM on travis prohibits Nico merging this; I'll do it.

PR #189, opened 2019-09-04
[parameters] Added Lb lifetime from HFLAV update for PDG2019

tblake00:


dvandyk:
https://travis-ci.org/eos/eos/jobs/580795549#L1873

> /src/eos/rare-b-decays/lambda-b-to-lambda-dilepton_TEST.cc:55:One relative error of 'd.differential_branching_ratio(16.0)' = '1.107310466481266e-07' and '+1.0923e-7' = '1.0923e-07' is greater than 0.0001. The results are 0.01355578849 and 0.01374207313125147

That's the test case you contributed 2 years ago.
This is quite in line with my plot that we are debugging right now.

dvandyk:
Replacing this PR with PR #192, which also fixes a bug in Lambda_b -> Lambda ll.

PR #202, opened 2019-10-01
[b-decays] Removed code duplication

### LINKED ISSUE #154: Code duplication in B->D l nu
Remove code duplication (3 times) when calculating in observables the amplitudes. Create dedicated struct for the calculation of the B->D l nu amplitudes and other intermediate quantities like overall prefactor, kinematics etc., which is then called from functions that predict observables. This is analogous to already done in B->D* l nu for the amplitudes

cBobeth:
Github: resolve #154

### DIFF
--- a/eos/b-decays/b-to-psd-l-nu.cc
+++ b/eos/b-decays/b-to-psd-l-nu.cc
@@ -31,6 +31,22 @@
 
 namespace eos
 {
+    namespace b_to_psd_l_nu
+    {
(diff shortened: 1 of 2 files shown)


dvandyk:
Bah. Typo in my commit message. Let's see if the Travis jobs finish now. Will push it manually after fixing the messages :-)

dvandyk:
Pushed manually.

PR #208, opened 2019-10-28
[eos] Rewriting SignalPDF and companion classes

SKuerten:
Rewriting SignalPDF to disentangle the pdf and its normalization.


dvandyk:
Thanks Stephan!

PR #213, opened 2019-12-09
Changes for v0.3.1

dvandyk:
@gubernari Compilation of your changes and mine, plus a new commit that exports the remaining q2-differential angular observables in B->D^*lnu.

Can you please review and "rebase and merge" if you approve?


dvandyk:
Tests failed since I had not update the BRs tests for B_q->D_q^*lnu. Also, a comment by Nico made me notice that the angular observables for B_s->D_s^*lnu were not exported at all. Both are fixed now with a force push.

dvandyk:
One of my python changes broke running the examples for the manual. We should have more of those!
Fixed and force pushed. Let's see if everything works now...

Issue #196, opened 2019-09-20
Prepare EOS for B->pi l nu analysis


dvandyk:
For Dom's project we need to
 - [ ] add all BaBar and Belle results for the B -> pi l nu q2-distributions as constraints
 - [ ] add all Lattice results for B ->pi form factors as constraints
 - [ ] add all B->pi form factors with pion LCDA; make Borel-param rescaling an optional feature (->@dvandyk)

For convenience we need to
 - [ ] show observables without a description in Jupyter if ```showall``` keyword is set to ```True```

PR #235, opened 2020-06-27
Changes related to issue #196

dvandyk:
Includes some major additions to the python-based CLI (eos-analysis).


dvandyk:
@cBobeth see anything that would preclude me from merging?

PR #246, opened 2020-10-27
BGL parametrisation of the B->D and B->D^* form factors

dvandyk:
Contributed by @cBobeth and @gubernari.

@cBobeth I have kept parts of the changes to B->Vlnu, which were still necessary for finite results at the endpoint even with your switch from A_2 -> A_12.


eos/statistics/log-likelihood.cc
```diff
@@ -1330,7 +1330,7 @@ namespace eos
                 {
                     double llh = (*b)->evaluate();
                     if (! std::isfinite(llh))
-                        return llh;
+                        return -std::numeric_limits<double>::infinity();
```
cBobeth, line 1333: Is the minus sign in front of std::... ok?

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -1096,6 +1169,103 @@ groups:
           min:     +4.667079193615467
           max:     +4.667079193615467
           comments: ''
+      # form factor parameters for B->D for [BGL1997]
```
cBobeth, line 1172: Correct in comment: ... B->D^* ...

cBobeth:
For me the canges seem fine.

eos/statistics/log-likelihood.cc
dvandyk: Yes.

MartinJung1978:
Hi, remember that one of the BGL parameters can be eliminated because of form-factor relations at q2=0. Also, the fit cannot give the overall normalization, since we only consider shape information.

MartinJung1978:
I had a look at the tests: they seem ok, i.e. I get similar values. I haven't tried to reproduce yours exactly, so probably some inputs are slightly different at the moment.

dvandyk:
> Hi, remember that one of the BGL parameters can be eliminated because of form-factor relations at q2=0. Also, the fit cannot give the overall normalization, since we only consider shape information.

@MartinJung1978 I agree that this should be implemented. However, the current implementation is not wrong in that regard. Hence, let's merge this first, and improve later.

eos/parameters/hadronic-matrix-elements.yaml
dvandyk: Fixed.

Issue #243, opened 2020-09-30
[C++14] Replace all uses of typedef with using


dvandyk:
From a discussion with @SKuerten: Instead of code that looks like
```C++
typedef QualifiedName KeyType;
```
we should use the more readable and modern code
```C++
using KeyType = QualifiedName;
```
@mreboud Want to try your hand on this one?

mreboud:
Yes!

mreboud:
@dvandyk Should I make this replacement everywhere? What about external codes such as e.g. integrate-curbature... ?

dvandyk:
@mreboud I would avoid external code, of which integrate-cubature is presently the only one AFAICS.

PR #247, opened 2020-10-28
[C++14] Replace all uses of typedef with using

mreboud:
Solves issue #243 by replacing all uses of typedef with using.
Typedefs are kept in eos/utils/integrate-cubature.cc/hh as this is external code.

### DIFF
--- a/eos/constraint.cc
+++ b/eos/constraint.cc
@@ -59,7 +59,7 @@ namespace eos
     template <>
     struct WrappedForwardIteratorTraits<ConstraintEntry::ObservableNameIteratorTag>
     {
-        typedef std::vector<QualifiedName>::const_iterator UnderlyingIterator;
+        using UnderlyingIterator = std::vector<QualifiedName>::const_iterator;
     };
     template class WrappedForwardIterator<ConstraintEntry::ObservableNameIteratorTag, const QualifiedName>;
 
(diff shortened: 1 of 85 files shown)


src/clients/eos-sample-pmc
```diff
@@ -0,0 +1,210 @@
+#! /bin/bash
```
dvandyk, line 1: This file seems to have been added erroneously.

dvandyk (changes requested):
Merci @mreboud!

This looks good, except for the one erroneously added file in the src/clients commit.
Nevertheless, I'd ask you to change the commit messages to the following format:

  [DIR] Message in imperative

where DIR is a metavariable for the most affected directory/library. So for your PR it would look like:

  [form-factors] Replace typedefs with using declarations
  [statistcs] Replace typedefs with using declarations
  [utils] Replace typedefs with using declarations
  [clients] Replace typedefs with using declarations
  [eos] Replace typedefs with using declarations

You can simply reword using 'rebase --interactive'.

Cheers!

src/clients/eos-sample-pmc
mreboud: Yes my mistake, thanks!

mreboud:
> Merci @mreboud!
> 
> This looks good, except for the one erroneously added file in the src/clients commit.
> Nevertheless, I'd ask you to change the commit messages to the following format:
> 
> [DIR] Message in imperative
> 
> where DIR is a metavariable for the most affected directory/library. So for your PR it would look like:
> 
> [form-factors] Replace typedefs with using declarations
> [statistcs] Replace typedefs with using declarations
> [utils] Replace typedefs with using declarations
> [clients] Replace typedefs with using declarations
> [eos] Replace typedefs with using declarations
> 
> You can simply reword using 'rebase --interactive'.
> 
> Cheers!

Ok I do this!

PR #253, opened 2020-11-25
Update installation.rst

pluegh:
Minor fixes: Apple's current OS name and clearer, more consistent section title


dvandyk:
Handled in #255

Issue #171, opened 2019-06-24
Various improvements to the parameter descriptions


dvandyk:
Depends on closing #165:

 - [x] Add LaTeX representation for all form factor parameters
 - [x] Add LaTeX representation for masses
 - [x] Add LaTeX representation for CKM parameters

 - [x] Retire parameter ```lambda_B_p```, and replace with 1 / ```B::1/lambda_B_p``` (mind the test cases!)
 - [x] Rename all ```B->K::a_{1,2,4}@{1,2.2}GeV``` to ```K::...``` (mind the test cases!)
 - [x] Retire parameters ```B->K^*::a_{perp,par}``` and replace with existing ```K^*::..``` (mind the test cases!)

 - [x] Replace ```mu``` in eos/rare-b-decays/ with a new parameter ```b->s::mu_b``` (to be placed in sm-and-eft.yaml, b->s  EFT group)
 - [x] Rename ```G_Fermi``` to ```mu->enunubar::G_Fermi```
 - [x] Rename ```hbar``` to ```QM::hbar```

dvandyk:
Any updates on this issue?

gubernari:
No sorry, I was busy with the other issues and I forgot about this. I will look into it after we close issue191

dvandyk:
Moving to 0.4

SKuerten:
- what was the consensus for ```G_Fermi```? Something like ```QM::G_Fermi``` or ```decay-constant::G_Fermi``` (from Standard Model & EFT Parameters)?
- The 4th case which was discussed in the meeting was ```mu``` to ```b->s::mu_b```, correct?

dvandyk:
@SKuerten it was ``WET::G_Fermi``. And for the scale, ``b->s::mu_b`` is correct.

SKuerten:
How do we want 
 - [ ] Retire parameter ```lambda_B_p```, and replace with 1 / ```B::1/lambda_B_p``` (mind the test cases!)
exactly?

There is already a parameter called ```B::1/lambda_B_p``` with a different value for ```lambda_B_p```  "cf. [BHvD2010] central value: 0.485 GeV" vs. "cf. [GKvD2018] central value: 0.460 GeV".
When I just change ```lambda_B_p``` to 1 / ```B::1/lambda_B_p``` all the test-cases will fail. How should I solve this? I could introduce ```B::1/lambda_B_p@BHvD2010``` ... I am not sure if this is wanted.

gubernari:
[...]

PR #256, opened 2020-12-03
[parameters] Update masses and add latex names for masses

mreboud:
Related to #171

### DIFF
--- a/eos/parameters/hadron-properties.yaml
+++ b/eos/parameters/hadron-properties.yaml
@@ -135,12 +135,6 @@ groups:
           central: +3.686097
           min:     +3.686072
           max:     +3.686122
-      # B meson masses
-      'mass::B^*_d' :
-          central: +5.32465
-          min:     +5.32440
-          max:     +5.32490
-          comments: 'cf. [PDG2012], p. 84'
   # }}}
 
   # Baryon Masses
--- a/eos/parameters/hadron-properties.yaml
+++ b/eos/parameters/hadron-properties.yaml
@@ -9,78 +9,93 @@ groups:
     parameters:
       # unflavoured meson masses
       'mass::pi^0' :
-          central: +0.1349766
-          min:     +0.1349760
-          max:     +0.1349772
-          comments: 'cf. [PDG-Live], May 2014'
+          central: +0.1349768
+          min:     +0.1349763
+          max:     +0.1349773
+          latex:   '$m_{\pi^0}$'
+          comments: "cf. [PDG-Live], December 2020"
       'mass::pi^+' :
-          central: +0.13957018
-          min:     +0.13956983
-          max:     +0.13957053
-          comments: ''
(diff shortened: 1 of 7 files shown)


eos/parameters/hadron-properties.yaml
```diff
@@ -5,177 +5,233 @@ groups:
   # Pseudoscalar Meson Masses
   # {{{
   - title: 'Pseudoscalar Meson Masses'
-    description: ''
+    description: 'Last update: December 2020'
```
dvandyk, line 8: I would not do that. The web page display for which version of EOS this was done.

eos/parameters/hadron-properties.yaml
```diff
@@ -5,177 +5,233 @@ groups:
...
-          comments: 'cf. [PDG2019]'
+          latex:   '$m_{B_s^*}$'
+          comments: 'cf. [PDG-Live], December 2020'
   # }}}
 
   # Vector Meson Masses
   # {{{
   - title: 'Vector Meson Masses'
-    description: ''
+    description: 'Last update: December 2020'
```
dvandyk, line 104: See above.

eos/parameters/hadron-properties.yaml
```diff
@@ -5,177 +5,233 @@ groups:
...
+          max:     +5.32491
+          latex:   '$m_{B^*}$'
+          comments: 'cf. [PDG-Live], December 2020'
   # }}}
 
   # Baryon Masses
   # {{{
   - title: 'Baryon Masses'
-    description: ''
+    description: 'Last update: December 2020'
```
dvandyk, line 200: And above.

dvandyk (changes requested):
Thanks Meril for updating these. However, this has caused a few test case failures. Those test case should be update to use the original value for the masses they depended upon.

Also, can you prefix the commit message with '[paramters] ' please?

eos/parameters/hadron-properties.yaml
mreboud: I agree, I will remove this

eos/parameters/hadron-properties.yaml
mreboud: I agree, I will remove this

eos/parameters/hadron-properties.yaml
mreboud: I agree, I will remove this

mreboud:
> Thanks Meril for updating these. However, this has caused a few test case failures. Those test case should be update to use the original value for the masses they depended upon.

Yes, the main purpose of the commit was to add charmonium resonances ; I thought I could update other parameters in the same commit but it is indeed more complex that this. Somehow the checks passed locally... I am investigating

> Also, can you prefix the commit message with '[paramters] ' please?

Sure I keep forgetting this.

PR #258, opened 2020-12-03
[python] Fix MatplotlibDeprecationWarning

pluegh:
Use single string for pgf.preamble instead of list

### DIFF
--- a/python/eos/plot/config.py
+++ b/python/eos/plot/config.py
@@ -41,10 +41,8 @@ matplotlib.rcParams['ytick.direction'] = 'out'
 
 matplotlib.rcParams['text.usetex'] = True
 
-matplotlib.rcParams['pgf.preamble'] = [
-    r'\usepackage[hidelinks]{hyperref}',
-    r'\usepackage{amsmath}',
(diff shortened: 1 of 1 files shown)


python/eos/plot/config.py
```diff
@@ -41,10 +41,6 @@
 
 matplotlib.rcParams['text.usetex'] = True
 
-matplotlib.rcParams['pgf.preamble'] = [
-    r'\usepackage[hidelinks]{hyperref}',
-    r'\usepackage{amsmath}',
-    r'\usepackage{xcolor}'
-]
+matplotlib.rcParams['pgf.preamble'] = r'\usepackage[hidelinks]{hyperref}\usepackage{amsmath}\usepackage{xcolor}'
```
dvandyk, line 44: Thanks for the initiative Philip! Can you modify this a little bit? I would prefer to have a multiline python string here, i.e.,

```
matplotlib.rcParams['pgf.preamble'] = r'''
\usepackage{...}
'''
```

with one package per line. This makes it easier to read diffs!

PR #260, opened 2020-12-06
Fix example notebooks

pluegh:


dvandyk:
@pluegh Can you please rebase this on top of the current master branch? I cannot merge this one as is with a clean history. Also, I think you squash the three commits into one.

One further element: Can you please make sure that the notebook is cleanly run once before saving. It seems that the outputs of some cells are not in order here.

Issue #257, opened 2020-12-03
Access a sub-sample of a constraints set at run-time


mreboud:
It would be useful to have the possibility to specify a subset of a kinematic range of a constraint at run-time.
This could be done by adapting the structures, based on ConstraintEntryBase in [/eos/constraint.cc](https://github.com/eos/eos/blob/master/eos/constraint.cc), that have a kinematics option.

dvandyk:
As a first attempt, I would specify a range of elements to use, e.g. by appending a ```begin``` and ```end``` option to the constraint's name. Example:

```
B^0->D^+e^-nu::BRs@Belle-2015A;begin=0,end=10   # this would be equivalent to all elements           
B^0->D^+e^-nu::BRs@Belle-2015A;begin=0,end=6   # this would be equivalent to the first 6 elements
```

The relevant code is here: https://github.com/eos/eos/blob/58563acc2efdda3eef1261e17f192eac252fc2de/eos/constraint.cc#L1135-L1170

I didn't see a built-in way to copy only a subvector or a submatrix in the GSL. While views exist, I don't see a memcpy or anything like that.

mreboud:
Indeed I don't see how we can memcpy a subvector/submatrix. Unless you have a better solution, I will get a view and use it to initialize the new structure.

dvandyk:
I think we can safely skip the view aspect, since we know who created the vectors and matrices (i.e., us) and they are stored consecutively in memory.

mreboud:
What is the aim of the "response" option in the MultivariateGaussianCovarianceConstraintEntry structure? I see that it is a `n_measurements * n_predictions` matrix and wonder how it should be treated when specifying a sub-sample of the measurements?

dvandyk:
Some experimental analyses do not provide unfolded results, but require instead forward folding of the theory predictions to the detector level. I would suggest for the time being to only apply the range reduction to the index of corresponding to the ```n_measurements``` dimension.

dvandyk:
@ESEberhard @mreboud Can we get a clean PR that fixes the outstanding problems? I think Mr Eberhard has two commits ready for that.

mreboud:
I do this asap (today)

dvandyk:
I think this can be closed now.

@mreboud Please reopen if I'm wrong...

mreboud:
[...]

PR #261, opened 2020-12-07
Add an option to the MultivariateGaussian class to run on a restricted number of measurments

mreboud:
First attempt to solve #257

### DIFF
--- a/eos/constraint.cc
+++ b/eos/constraint.cc
@@ -27,6 +27,7 @@
 #include <eos/utils/private_implementation_pattern-impl.hh>
 #include <eos/utils/qualified-name.hh>
 #include <eos/utils/stringify.hh>
+#include <eos/utils/destringify.hh>
 #include <eos/utils/wrapped_forward_iterator-impl.hh>
 
 #include <boost/filesystem/operations.hpp>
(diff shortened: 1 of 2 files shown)


eos/constraint.cc
```diff
@@ -1145,19 +1146,30 @@ namespace eos
                     throw InternalError("make_multivariate_gaussian_covariance_constraint<measurements=" + stringify(dim_meas) + ",predictions=" + stringify(dim_pred) + ">: " + name.str() + ": '" + this->observables[i].str() + "' is not a valid observable name");
             }
 
-            // create GSL vector for the mean
-            gsl_vector * means = gsl_vector_alloc(dim_meas);
-            gsl_vector_memcpy(means, this->means);
+	    //If specified, these options allow to specify a subset of the measurements
```
dvandyk, line 1149: Logic looks good, but what happened with the indentations?

eos/constraint.cc
```diff
@@ -1145,19 +1146,30 @@ namespace eos
...
+	    unsigned begin = destringify<unsigned>(options.get("begin", "0"));
+	    unsigned end = destringify<unsigned>(options.get("end", stringify(dim_meas)));
+	    unsigned sub_dim = end - begin;
+
+	    if (! sub_dim>0)
```
dvandyk, line 1154: I would make it clearer and write ```0 > sub_dim```, unless you have a reason for this I do not see.

eos/constraint.cc
```diff
@@ -1145,19 +1146,30 @@ namespace eos
                     throw InternalError("make_multivariate_gaussian_covariance_constraint<measurements=" + stringify(dim_meas) + ",predictions=" + stringify(dim_pred) + ">: " + name.str() + ": '" + this->observables[i].str() + "' is not a valid observable name");
             }
 
-            // create GSL vector for the mean
-            gsl_vector * means = gsl_vector_alloc(dim_meas);
-            gsl_vector_memcpy(means, this->means);
+	    //If specified, these options allow to specify a subset of the measurements
+	    unsigned begin = destringify<unsigned>(options.get("begin", "0"));
+	    unsigned end = destringify<unsigned>(options.get("end", stringify(dim_meas)));
+	    unsigned sub_dim = end - begin;
```
dvandyk, line 1152: I was wondering if it would be helpful (for potential later adaptations) to name this ```subdim_meas```?

eos/constraint_TEST.cc
```diff
@@ -965,6 +965,51 @@ class ConstraintDeserializationTest :
...
+                    "dof: 3"
+                );
+
+                YAML::Node node = YAML::Load(input);
+
+                std::shared_ptr<ConstraintEntry> entry(ConstraintEntry::FromYAML("Test::MultivariateGaussian(Covariance)", node));
+                TEST_CHECK(nullptr != entry.get());
+
+                Constraint c = entry->make("Test::MultivariateGaussian(Covariance)", Options{{"begin","0"}, {"end","2"}});
+		std::vector<LogLikelihoodBlockPtr> blocks(c.begin_blocks(), c.end_blocks());
```
dvandyk, line 1000: Again, what happened with the indentations? Did you use tabs?

dvandyk (changes requested):
Just some cosmetic changes, and then it's good to go! Thanks Meril!!

eos/constraint.cc
mreboud: Absolutely, the name sub_dim was referring to "subset dimension", but yours is better

eos/constraint_TEST.cc
mreboud: hmm emacs is indenting automatically and displays the result correctly. I don't know why it appears weird on github, I will fix this.

eos/constraint.cc
mreboud: I have no reason :)

dvandyk (approved):

eos/constraint.cc
mreboud: Hmm all well thought this was stupid. I defined sub_dim as unsigned, so:
 1 - The if statement is always true
 2 - If `end - begin < 0`, sub_dim is not well defined and will certainly lead to an error...

dvandyk: Yes, good point. I didn't see that. I would suggest to use signed integeers, and
do the error checking more explicitly:

```
if (end > dim_meas)
    throw OptionValueError(... + "cannot use a value of 'end' pointing beyond the number of measurements.");

if (begin < 0)
   throw OptionValueError(... + "cannot use a negative value for 'begin'");

if (begin >= end)
   throw OptionValueError(... + "cannot use a value for 'begin' equal to or larger than 'end'");
```

No need to revert the PR, just push a correction in a separate commit on top of this branch and we handle it here.

mreboud: Ok I do this, thank you

Issue #157, opened 2019-06-05
Add non-local matrix elements


dvandyk:
- [ ] add a base class template ```NonLocalCorrelator<Process_, Transition_>``` and specialize for ```Transition_ = PToV``` and ```Transition_ = PToP```.

PR #268, opened 2021-02-03
Implement parametrizations of non-local form-factor for P->P and P->V transitions

mreboud:
First step for issue #157.
Implements B->K(*)charmonium decays and related observables.


dvandyk:
Thanks Meril!

PR #269, opened 2021-02-09
Changes for B->D^(*)lnu code

dvandyk:


cBobeth (commented):
I think we forgot to increase number of integration points in b-to-vec-l-nu.cc in line 370?! It should be from 256 -> 4096.

eos/b-decays/b-to-vec-l-nu.cc
```diff
@@ -408,87 +408,87 @@ namespace eos
...
+        return 3.0 / 4.0 * o.vv4T();
     }
 
     double
-    BToVectorLeptonNeutrino::differential_J4_normalized(const double & q2) const
+    BToVectorLeptonNeutrino::differential_J4(const double & q2) const
     {
         auto   o = _imp->differential_angular_observables(q2);
-        return 3.0 / 4.0 * -o.vv10T() / o.normalized_decay_width();
+        return 3.0 / 4.0 * -o.vv10T();
```
cBobeth, line 480: We need to check whether there should be these minus signs in

J_4, 6s, 6c, 7, 9

dvandyk: Check, or did you come to a consensus with @MartinJung1978 that this is not conforming to the Belle definitions?

cBobeth: We checked that this is not conforming with the Belle 2018 definition. You might cherry-pick from my git-repo (branch BtoDstarlnu) latest commit for correction of minus signs and adaption of test cases.

cBobeth: I try to finish another commit with a test case for B->D* lv angular observables including new physics contributions that could be also added to this pull request.

cBobeth: Code changes look fine for me.

PR #271, opened 2021-02-13
Stop using Minuit2

dvandyk:
Not being able to download the Minuit2 source code on a new laptop focuses the mind brilliantly...


mreboud (approved):
We still need to update the documentation, right?

dvandyk:
@mreboud good point. Wanna try your hand?

mreboud:
Yes, I just need to remove all instance of Minuit2, right?

dvandyk:
I think so!

dvandyk:
@mreboud will merge it right now, we just need to keep in mind to fix the documentation before the next release.

PR #272, opened 2021-02-13
Include python CLI scripts in the Python wheels

dvandyk:


mreboud (commented):
@dvandyk Why are you updating manylinux only?

mreboud (approved):
Ok I was a bit confused but everything is fine for me!

PR #276, opened 2021-02-25
Remove Minuit2 from the documentation

mreboud:
Additional fix for #271


dvandyk (approved):
Hm, we need to remove the dependencies on Minuit2 also from the .deb packages and from the homebrew formula.

mreboud:
> Hm, we need to remove the dependencies on Minuit2 also from the .deb packages and from the homebrew formula.

Ah I missed that sorry!

Issue #275, opened 2021-02-24
A possible mistake in the B->Plnu NP observables


DomLeljak:
In `eos/b-decays/b-to-psd-l-nu.cc` line 304 (in the definition of `numerator_differential_a_fb_leptonic`) says:

- `std::abs(amp.h_0) * std::abs(amp.h_tS) * (1.0 - amp.v)`.

A comment states this is derived from `[DDS:2014A]`, eq. (12), and this term should come from the linear cos(theta) term in the first line of eq. (12). However, according to `[DDS:2014A]`, eq. (12), _if the entire (H_0 cos(theta) - H_tS) term is under absolute value_, line 304 should say something like:

- `std::real(amp.h_0 * std::conj(amp.h_tS)) * (1.0 - amp.v)`.

The two expressions are the same if the Wilson coefficients are real, but I think line 304 is wrong currently if they are considered complex. The final expression can/should be cross checked with `[STTW:2013A]`, eq. (53a). @dvandyk please check...

dvandyk:
@DomLeljak thanks for this, I concur that this seems incorrect. ``git blame`` names me for this line, since I merged the B->Dlnu and B->pilnu codes. However, this bug was already in the B->Dlnu code before that. I think I can trace it back to @akokulu7's commit 7bb8aab8.

@cBobeth: @DomLeljak's solution seems correct. Do you concur?

dvandyk:
Dom, please go ahead with the fix!

akokulu7:
dear all (@dvandyk, @cBobeth , @DomLeljak ), 
concerning this issue I have received a mention email yesterday from github/eos and today I have found some time to crosscheck this issue (#275) for you, based on my past notes/programs. I have found my related mathematica program and now numerically crosschecked your point related to A_FB^leptonic  in   DDS:2014A versus STTW:2013A. (of course with complex WCs).
I confirm that the term in eq. (12) from [DDS:2014A], (H0 * cos(theta) - HtS)^2  is wrongly typed in this paper and should be understood as |H0 * cos(theta) - HtS|^2    instead of  (|H0| * cos(theta) - |HtS|)^2, as Dom pointed out. It is true that the discrepancy would only be noticeable when there are complex couplings.
So I concur this fix too! Thanks @DomLeljak  for pointing this out.
cheers, ahmet

PR #277, opened 2021-02-25
All the changes from Issue196

DomLeljak:
@dvandyk The fixup for #275 is in e6d13c4 and c829b01.

There's a lot of stuff, probably a lot of conflicts, let me know if you want to do it together.


DomLeljak:
I just realized, probably should have compared to dvandyk/eos at issue196 first...

cBobeth (commented):
Please correct also line 276 in b-to-psd-l-nu.cc

DomLeljak:
Missed line 276 while patching... Thanks.

dvandyk:
We cleaned this up. @cBobeth we fixed line 276 in the process. Thanks for the heads up!

cBobeth:
I did not yet check whether there are some other issues. There is the 1/m_lep^2 singularity in the amplitude H_tS, which we could make explicit, as in B->D* lv, but this can also be a separate commit.

dvandyk:
@cBobeth this is what we used for our recent paper, so I'd like to have it in the master repository if it passes the tests, which it did.

cBobeth:
@dvandyk: Yes, it can be a separate commit. Since it is just rewriting equivalently, all test cases should not be affected.

PR #279, opened 2021-02-25
[python] Fix possible log(0)

pluegh:
Encountered when np.sum( ) returns infinity.


dvandyk:
@pluegh Is this correct? If any single weight is ``NaN``, then the sum will also be, and therefore all normalized weights. I think the order is correct. In the present version, the only way a log(0) can appear is due to **all** weights being zero, right?

PR #288, opened 2021-03-11
[python] Fix automatic formatting of axes ticks

pluegh:
The changes are tested for all cases using the notebook provided in `examples/inference.ipynb`. The user is warned about potential issues, covering all cases.


pluegh:
A small update of the example notebooks might also be in order. Nothing broke, but the default usage is slightly easier now. I'll update the PR.

python/eos/plot/plotter.py
```diff
@@ -105,18 +105,21 @@ def setup_plot(self):
...
                 self.xscale = float(myx['scale'])
                 self.xticks = matplotlib.ticker.FuncFormatter(lambda x, pos, xscale=self.xscale: self.xformat.format(x=x / xscale))
-            else:
-                self.xticks = matplotlib.ticker.StrMethodFormatter(self.xformat)
+                self.ax.xaxis.set_major_formatter(self.xticks)
 
-            self.ax.xaxis.set_major_formatter(self.xticks)
+            else:
+                if 'format' in myx:
+                    eos.warn("Argument plot:x:format is only used when plot:x:scale is used")
```
dvandyk, line 122: This change should also be reflected in the documentation (``doc/api/python.rst``)

dvandyk (changes requested):
Except for the missing change to the documentation this looks very good!

python/eos/plot/plotter.py
pluegh: Done!

dvandyk (approved):

PR #293, opened 2021-03-23
[python] Add new plot type: 'constraint2D'

dvandyk:
@mreboud Could you please check my logic to compute the ellipses?

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -1,5 +1,5 @@
 # Copyright (c) 2018 Frederik Beaujean
-# Copyright (c) 2017, 2018 Danny van Dyk
+# Copyright (c) 2017, 2018, 2021 Danny van Dyk
 #
 # This file is part of the EOS project. EOS is free software;
 # you can redistribute it and/or modify it under the terms of the GNU General
@@ -48,6 +48,7 @@ class Plotter:
         self.plot_types = {
             'band':                  Plotter.Band,
             'constraint':            Plotter.Constraint,
+            'constraint2D':         Plotter.Constraint2D,
             'constraint-overview':   Plotter.ConstraintOverview,
             'contours2D':            Plotter.Contours2D,
             'expression':            Plotter.Expression,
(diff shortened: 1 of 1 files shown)


python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+                self.xobservable = xitem['observable']
+                self.xkinematics = xitem['kinematics'] if 'kinematics' in xitem else {}
+                self.xoptions    = xitem['options']    if 'options'    in xitem else {}
+
+                if 'y' not in item:
+                    raise KeyError('no y-axis specification for 2D constraint')
+
+                yitem = item['y']
+                if 'observable' not in yitem:
+                    raise KeyError('no observable in 2D constraint\'s x-axis specification')
```
pluegh, line 721: `x-axis` → `y-axis`

python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+
+                    means  = np.array(constraint['means'])
+                    xmean  = means[xidx]
+                    ymean  = means[yidx]
+
+                    cov    = np.array(constraint['covariance'])
+                    xsigma = np.sqrt(covariance[xidx, xidx])
+                    ysigma = np.sqrt(covariance[yidx, yidx])
+                    rho    = cov[xidx, yidx] / (xsigma * ysigma)
+                else:
```
pluegh, line 749: `if constraint['type'] == 'MultivariateGaussian'` is easier to understand

python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+                sigmas = [1.0, 2.0, 3.0]
+                alphas = np.linspace(0.0, self.alpha, len(sigmas) + 1)[1:]
+                for sigma, alpha in zip(sorted(sigmas, reverse=True), alphas):
+                    xwidth = 2.0 * xsigma * np.sqrt(1.0 + rho) * sigma
+                    ywidth = 2.0 * ysigma * np.sqrt(1.0 - rho) * sigma
+
+                    ellipse = matplotlib.patches.Ellipse((xmean, ymean), width=xwidth, height=ywidth, angle=45,
+                                          alpha=alpha, color=self.color, linewidth=1, fill=True)
+                    self.plotter.ax.add_patch(ellipse)
+            else: # univariate gaussian
```
pluegh, line 786: again, `if constraint['type'] == 'Gaussian'` would be easier to comprehend

python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+                    sigma_hi = np.sqrt(constraint['sigma-stat']['hi']**2 + constraint['sigma-sys']['hi']**2)
+                    sigma_lo = np.sqrt(constraint['sigma-stat']['lo']**2 + constraint['sigma-sys']['lo']**2)
+
+                    xlo, xhi = (mean - sigma_lo, mean + sigma_hi)
+                    ylo, yhi = self.plotter.yrange
+
+                    rect = matplotlib.patches.Rectangle((xlo, ylo), width=xhi - xlo, height=yhi - ylo,
+                                                       alpha=self.alpha, color=self.color)
+                    self.plotter.ax.add_patch(rect)
+                else: # 'y' in self.item
```
pluegh, line 816: `if 'y' in self.item`

pluegh (changes requested):
Looking good except for a single typo. And I would avoid `else` altogether when all cases that are tested against are known.

python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+                    rho    = np.array(constraint['correlations'])[xidx, yidx]
+
+                sigmas = [1.0, 2.0, 3.0]
+                alphas = np.linspace(0.0, self.alpha, len(sigmas) + 1)[1:]
+                for sigma, alpha in zip(sorted(sigmas, reverse=True), alphas):
+                    xwidth = 2.0 * xsigma * np.sqrt(1.0 + rho) * sigma
+                    ywidth = 2.0 * ysigma * np.sqrt(1.0 - rho) * sigma
+
+                    ellipse = matplotlib.patches.Ellipse((xmean, ymean), width=xwidth, height=ywidth, angle=45,
+                                          alpha=alpha, color=self.color, linewidth=1, fill=True)
```
mreboud, line 784: I don't understand your ellipse, why is the angle set at 45? Shouldn't this angle depend on the correlation?

mreboud (changes requested):

python/eos/plot/plotter.py
```diff
@@ -671,6 +672,181 @@ def plot(self):
...
+
+                    xidx = observables.index(xobservable)
+                    yidx = observables.index(yobservable)
+
+                    means  = np.array(constraint['means'])
+                    xmean  = means[xidx]
+                    ymean  = means[yidx]
+
+                    sigmas = np.sqrt(
+                          (np.array(constraint['sigma-stat-hi'])**2 + np.array(constraint['sigma-stat-lo'])**2) / 4.0
```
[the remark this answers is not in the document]
dvandyk: Discussed this with Meril. We will use the square of sum of absolute values for the time being.

python/eos/plot/plotter.py
dvandyk: I will investigate. Might have implemented the matplotlib example incorrectly.

dvandyk: I made a mistake. The various transformations do not commute. Fixed.

PR #295, opened 2021-03-26
[python] Add an option to Analysis to fix parameters to a set of values

mreboud:


mreboud:
Ah sorry, I realize that the description is not updated

python/eos/analysis.py
```diff
@@ -81,11 +83,18 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={})
         eos.debug('manual_constraints:')
         for cn, ce in manual_constraints.items():
             eos.debug(' - {name}'.format(name=cn))
+        eos.debug('fixed_parameters:')
+        for cn, ce in fixed_parameters.items():
```
dvandyk, line 87: Should that not me ``pn, pv`` as in parameter name, parameter value?

dvandyk (approved):
Mostly OK, wouldn't mind merging it. However, see one comment on weird naming of iterator elements.

@mreboud: merge if you can't be bothered :smile:

python/eos/analysis.py
mreboud: It definitely should :)

PR #296, opened 2021-03-27
[python] Add method to get installation directory

mreboud:


python/eos/__init__.py
```diff
@@ -58,6 +58,9 @@ def warn(msg, *args, **kwargs):
 def installation_time():
     return _time.ctime(_os.path.getmtime(eos.__file__))
 
+def installation_dir():
+    return eos.__file__[:-11]
```
dvandyk, line 62: Nope, let's do this properly!

```
return _os.path.dirname(eos.__file__)
```

dvandyk (changes requested):
Let's please use the tools Python gave us! :smile:

PR #299, opened 2021-04-01
Add a 'basics' example notebook

dvandyk:


examples/basics.ipynb
```diff
@@ -0,0 +1,462 @@
...
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [
+    {
+     "output_type": "execute_result",
+     "data": {
+      "text/plain": [
+       "<_eos.Parameter at 0x7f61ba9fff38>"
+      ],
+      "text/html": "\n    <table>\n        <tr>\n            <th>$$m_\\mu$$</th>\n            <td>(eos.Parameter)</td>\n        </tr>\n        <tr>\n            <th>current value</th>\n            <td><tt>0.105658</tt></td>\n        </tr>\n        <tr>\n            <th>default value</th>\n            <td><tt>0.105658</tt></td>\n        </tr>\n    </table>"
```
mreboud, line 76: Why is the muon mass latex rep rendered with two `\` ?

mreboud (commented):
Excellent, I like it :)

examples/basics.ipynb
dvandyk: Ah, that's because it needs to be escaped. Particular problem when converting from and to Python.
Python knows ``r'\mu'`` and uses only a single backslash. But ``'\\mu'`` needs two, to escape the backslash. Otherwise it would use the control character ``\m``, which is the carriage return.

At no point do we write ``\\`` in the code. In the YAML, using ``'...'`` ensures that. In Python, using ``r'...'`` "raw strings" ensures it. And in C++, using ``R"(...)"`` raw strings ensures that.

PR #300, opened 2021-04-06
[constraints] Add RD^(*) measurments average from HFLAV

MuslemRahimi:


eos/constraints/B-to-D-l-nu.yaml
```diff
@@ -132,4 +132,21 @@ B^0->D^+l^-nu::KinematicalDistribution[w]@Belle-2015A:
       0.008462011094650328, 0.01124041430207108, 0.013615071527758647, 0.015895479300835378,
       0.02293292241393046]
     dof: 9
+
```
dvandyk, line 135: The metadata (which source, what year) is missing. Have a look how this is done in lines 1-7.

eos/constraints/B-to-D-l-nu.yaml
```diff
@@ -132,4 +132,21 @@ B^0->D^+l^-nu::KinematicalDistribution[w]@Belle-2015A:
...
+B->D^(*)lnu::R_D^(*)@HFLAV-2019:
+    type: MultivariateGaussian(Covariance)
+    dim: 2
+    observables:
+        [ B->Dlnu::R_D, B->D^*lnu::R_D^*]
+    options:
+        [ { l: mu, q: d }, { l: tau, q: d }]
+    kinematics:
+    - {q2_tau_max: 11.62, q2_tau_min: 3.15, q2_mu_min: 0.01, q2_mu_max: 11.62}
```
dvandyk, line 144: Hm, I'm surprised that the ``eos/constraints_TEST`` passes. But nothing for you to worry about.

dvandyk (changes requested):
Except for the metadata, tis looks good. Please fix that, and force push the fixed commit to the same remote branch to update this PR.

eos/constraints/B-to-D-l-nu.yaml
mreboud: Why shouldn't it pass?

dvandyk: For some reason I had in mind that ``constraint_TEST`` deserializes and then serializes the constraints, to only check if the input string is identical to the output string. But we only do that for a small set of test cases. The full set of constraints is *only* created, but not serialized.
All is well :+1:

PR #301, opened 2021-04-08
Update basics.ipynb

romyoc:
I just made some tiny adjustments to the text. Perhaps in the observable section we can include something about how you can alter an observables value depending on what you set your parameter to. There is already something like this in input 22 but maybe it would be good to emphasise this more. Otherwise everything seems very straightforward.

### DIFF
--- a/examples/basics.ipynb
+++ b/examples/basics.ipynb
@@ -29,9 +29,9 @@
   },
   {
    "source": [
-    "The import statement should go through without any error messages. If there are errors, they might related to the following issues:\n",
+    "The import statement should go through without any error messages. If there are errors, they might be related to the following issues:\n",
     " - missing Python packages (try to run ``!pip3 install -U matplotlib numpy scipy``);\n",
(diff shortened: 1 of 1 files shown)


examples/basics.ipynb
```diff
@@ -41,7 +41,7 @@
     "## (Almost) Everything can be a Parameter\n",
     "\n",
     "In EOS we make a lot of use of the ``eos.Parameter`` class. It provides access to one real-valued and scalar parameter, among a diverse and large set of parameters. You cannot create an object of type ``eos.Parameter`` yourself. Instead, you can create a new\n",
-    "set of Parameters, and extract your favourite parameter to get to know it."
+    "set of parameters, from which your favourite can be extracted and looked at."
```
dvandyk, line 44: your favourite **parameter**?

dvandyk (changes requested):
The changes are fine, please review my one suggestion.
Please use ```git commit --amend``` to change the first line of the commit messag.
It should read ```[examples] Update 'basics' example notebook```.

dvandyk:
Replaced by #322. Closing.

PR #302, opened 2021-04-08
WIP: Several implementations for the study of b->sll decays

mreboud:
This large PR apply the following changes:

- Trade `Low-` and `LargeRecoil` templates for `GP2004` and `BFS2004` tags for B->Kll and B->K*ll
- Use `nonlocal-formfactor` framework to implement Bs->phi psi
- Update and adds measurement for these decays
- Implement B->K(*)ll and B_s->phi ll à la GvDV2020


eos/rare-b-decays/nonlocal-formfactors-p-to-p.cc
```diff
@@ -269,6 +277,13 @@ namespace eos
                     return 0.0;
                 }
 
+                virtual complex<double> normalized_plus(const double & q2) const
+                {
+                    const double F_plus = form_factors->f_p(q2);
+
+                    return H_plus(q2)/F_plus;
```
dvandyk, line 284: Please use some whitespaces around the operator ``/`` for better visibility.

eos/rare-b-decays/nonlocal-formfactors-p-to-p.cc
```diff
@@ -61,6 +61,11 @@ namespace eos
                     return 0.0;
                 }
 
+                virtual complex<double> normalized_plus(const double & q2) const
```
dvandyk, line 64: ``normalized_plus`` sounds like H normalized to unity, rather than the ratio with the local form factor. Can we use ``ratio_plus`` instead?

eos/rare-b-decays/nonlocal-formfactors-p-to-p.cc
```diff
@@ -610,6 +635,14 @@ namespace eos
         return abs(this->_imp->nff->Hhat_plus(q2));
     }
 
+    template <typename Process_>
+    double
+    NonlocalFormFactorObservable<Process_, nff::PToP>::abs_normalized_plus(const double & q2) const
```
dvandyk, line 641: Why the absolute value?

eos/rare-b-decays/nonlocal-formfactors-p-to-p.cc
```diff
@@ -610,6 +635,14 @@ namespace eos
...
+    {
+        return abs(this->_imp->nff->normalized_plus(q2));
```
dvandyk, line 643: ``abs`` could be ``::abs`` (i.e., the C version) rather than ``std::abs``. Can you make sure to add ``using std::abs`` somewhere at the top of the ``.cc`` file, but within the ``eos`` namespace?

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -99,6 +99,20 @@ namespace eos
                 {
                     return 0.0;
                 }
+                virtual complex<double> normalized_perp(const double & q2) const
```
dvandyk, line 89: Ditto: ``normalized`` vs ``ratio``.

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -473,6 +487,32 @@ namespace eos
...
                 }
 
+                virtual complex<double> normalized_perp(const double & q2) const
+                {
+                    const double m_V2  = pow(m_V, 2);
+                    const double m_B2  = pow(m_B, 2);
+                    const double lambda = pow(m_B2, 2) + pow(m_V2, 2) + pow(q2, 2) - 2*m_B2*m_V2 - 2*m_B2*q2 - 2*q2*m_V2;
+                    const double F_perp = pow(2 * lambda, 0.5) / (m_B+m_V) / m_B * form_factors->v(q2);
+
+                    return H_perp(q2)/F_perp;
```
dvandyk, line 484: Ditto, whitespaces.

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -473,6 +487,32 @@ namespace eos
...
+                }
+
+                virtual complex<double> normalized_para(const double & q2) const
+                {
+                    const double F_para = pow(2, 0.5) * (m_B+m_V) / m_B * form_factors->a_1(q2);
+
+                    return H_para(q2)/F_para;
```
dvandyk, line 491: Whitespaces

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -473,6 +487,32 @@ namespace eos
...
+                }
+
+                virtual complex<double> normalized_long(const double & q2) const
+                {
+                    const double m_V2  = pow(m_V, 2);
+                    const double m_B2  = pow(m_B, 2);
+                    const double lambda = pow(m_B2, 2) + pow(m_V2, 2) + pow(q2, 2) - 2*m_B2*m_V2 - 2*m_B2*q2 - 2*q2*m_V2;
+                    const double F_long = ( (m_B2-m_V2-q2)*pow(m_B+m_V, 2)*form_factors->a_1(q2) - lambda*form_factors->a_2(q2) ) / (2*m_V*m_B2*(m_B+m_V));
+
+                    return H_long(q2)/F_long;
```
dvandyk, line 501: Whitespaces.

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -473,6 +487,32 @@ namespace eos
                     return H_residue_psi2s(phiParam, alpha_0, alpha_1, alpha_2);
                 }
 
+                virtual complex<double> normalized_perp(const double & q2) const
+                {
+                    const double m_V2  = pow(m_V, 2);
+                    const double m_B2  = pow(m_B, 2);
+                    const double lambda = pow(m_B2, 2) + pow(m_V2, 2) + pow(q2, 2) - 2*m_B2*m_V2 - 2*m_B2*q2 - 2*q2*m_V2;
```
dvandyk, line 481: Why not use ``eos::lambda``, which is defined in ``eos/utils/kinematic.hh``? :smile:

eos/rare-b-decays/nonlocal-formfactors-p-to-v.cc
```diff
@@ -473,6 +487,32 @@ namespace eos
...
+
+                    return H_para(q2)/F_para;
+                }
+
+                virtual complex<double> normalized_long(const double & q2) const
+                {
+                    const double m_V2  = pow(m_V, 2);
+                    const double m_B2  = pow(m_B, 2);
+                    const double lambda = pow(m_B2, 2) + pow(m_V2, 2) + pow(q2, 2) - 2*m_B2*m_V2 - 2*m_B2*q2 - 2*q2*m_V2;
+                    const double F_long = ( (m_B2-m_V2-q2)*pow(m_B+m_V, 2)*form_factors->a_1(q2) - lambda*form_factors->a_2(q2) ) / (2*m_V*m_B2*(m_B+m_V));
```
dvandyk, line 499: Whitespaces!

eos/rare-b-decays/nonlocal-formfactors.hh
```diff
@@ -87,6 +87,13 @@ namespace eos
 
             ///@}
 
+            ///@name Evaluate the ratio between non-local and local formfactors.
+            ///@{
+
+            virtual complex<double> normalized_plus(const double & /*q2*/) const = 0;
```
dvandyk, line 93: It's OK to have a named argument here. Warnings only appear in the definition if ``q2`` is unused.

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -3366,6 +3366,78 @@ groups:
           central: +0.0
           min:     +0.0
           max:     +0.0
+      'B_s->phiccbar::Re{alpha_0^perp}@GvDV2020' :
+          central: +0.0
+          min:     +0.0
+          max:     +0.0
```
dvandyk, line 3372: Can you add a simple latex representation for these? Skip the GvDV2020, all we need is the B_s->phi part, the polarization, and the order.

eos/rare-b-decays/Makefile.am
```diff
@@ -7,6 +7,7 @@ lib_LTLIBRARIES = libeosrarebdecays.la
 libeosrarebdecays_la_SOURCES = \
 	b-to-k-charmonium.cc b-to-k-charmonium.hh \
 	b-to-kstar-charmonium.cc b-to-kstar-charmonium.hh \
+        bs-to-phi-charmonium.cc bs-to-phi-charmonium.hh \
```
dvandyk, line 10: This is going to cause trouble. There are spaces instead of tabs.

dvandyk (changes requested):
Most relevant: to use ``eos::lambda`` :smile:
Have a look at the other comments.

eos/parameters/hadronic-matrix-elements.yaml
mreboud: Yes, I was waiting for issue 171 to be closed before, to avoid painful rebasing.

eos/rare-b-decays/nonlocal-formfactors-p-to-p.cc
mreboud: It needs to be a double no? I also added imaginary and real parts

eos/rare-b-decays/Makefile.am
mreboud: My mistake, I don't know what caused this, sublime business I think. Anyway, corrected

eos/rare-b-decays/observables.cc
```diff
@@ -334,7 +335,11 @@ namespace eos
 
                 /// Branching ratio of B -> K^* psi
                 make_observable("B->K^*charmonium::branching_ratio",
-                        &BToKstarCharmonium::branching_ratio)
+                        &BToKstarCharmonium::branching_ratio),
+
+                // Branching ratio of B -> phi psi
+                make_observable("B_s->phicharmonium::branching_ratio",
```
dvandyk, line 341: Argh, I overlooked this in a previous commit: Please use the following notation for the observables, to make it more manageable:
```
B->K^*psi::BR
B_s->phi::psi::BR
```
etc.

mreboud: Why do you have :: between phi and psi and not between K^* and psi ?

dvandyk: The idea is to use the name of the process as a namespace / prefix.

``B->K^*psi::BR`` is the observable ``BR`` in the ``B->K^*psi`` namespace.

mreboud: Then I would write `B_s->phipsi::BR`, or am I missing something?

dvandyk: Yes. Ah, I mistyped the original comment.  How did that happen?  ```B_s->phipsi::BR``` is what I'm after.

mreboud: Excellent, I will do the change, thanks!

dvandyk (approved):

PR #314, opened 2021-05-21
Fixes for Romy

dvandyk:
- R_K has two implementation; remove the specific one and keep the general one
 - B->Kll observables use GSL::QNG, which fails to reach the needed precision in the q2 range 1.1 to 6.0 GeV^2; switch to GSL::QAGS


mreboud:
I was indeed also removing this implementation of R_K (idem for R_K*) in #302

PR #318, opened 2021-05-25
B_q <-> Bbar_q mixing

dvandyk:
Available only for q=s at the moment, but readily extensible.

TODOs:
 - [x] test case for SM component
 - [x] test case for WilsonScan component
 - [x] test case for DeltaM_s

### DIFF
--- a/eos/meson-mixing/bq-mixing.cc
+++ b/eos/meson-mixing/bq-mixing.cc
@@ -1,0 +1,121 @@
+/* vim: set sw=4 sts=4 et foldmethod=syntax : */
+
+/*
+ * Copyright (c) 2021 Danny van Dyk
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include <eos/meson-mixing/bq-mixing.hh>
+#include <eos/utils/destringify.hh>
+#include <eos/utils/model.hh>
+#include <eos/utils/power_of.hh>
+#include <eos/utils/private_implementation_pattern-impl.hh>
+#include <eos/utils/options-impl.hh>
+
+namespace eos
+{
+    using std::norm;
+
+    template <>
+    struct Implementation<BMixing>
+    {
+        std::shared_ptr<Model> model;
+
+        UsedParameter hbar;
+
+        UsedParameter g_fermi;
+
+        SwitchOption opt_q;
+
+        UsedParameter m_B;
+
+        UsedParameter f_B;
+
+        UsedParameter tau_B;
+
+        UsedParameter R_1;
+        UsedParameter R_2;
+        UsedParameter R_3;
+        UsedParameter R_4;
+        UsedParameter R_5;
+
+        Implementation(const Parameters & p, const Options & o, ParameterUser & u) :
+            model(Model::make(o.get("model", "SM"), p, o)),
+            hbar(p["QM::hbar"], u),
+            g_fermi(p["WET::G_Fermi"], u),
+            opt_q(o, "q", {"s"}),
+            m_B(p["mass::B_" + opt_q.value()], u),
+            f_B(p["decay-constant::B_" + opt_q.value()], u),
+            tau_B(p["life_time::B_" + opt_q.value()], u),
+            R_1(p["B_s<->Bbar_s::R^1"], u),
+            R_2(p["B_s<->Bbar_s::R^2"], u),
+            R_3(p["B_s<->Bbar_s::R^3"], u),
+            R_4(p["B_s<->Bbar_s::R^4"], u),
+            R_5(p["B_s<->Bbar_s::R^5"], u)
+        {
+            u.uses(*model);
+        }
+
+        complex<double> M_12() const
+        {
+            const double mu = 4.2;
+            const auto wc = model->wilson_coefficients_sbsb(mu);
+
+            // cf. [DDHLMSW:2019A]
+            const std::array<complex<double>, 8> contributions{{
+                wc.c1()  * R_1(),
+                wc.c2()  * R_2(),
+                wc.c3()  * R_3(),
+                wc.c4()  * R_4(),
+                wc.c5()  * R_5(),
+                wc.c1p() * R_1(), // primed operators share the hadronic matrix elements of their unprimed partners
+                wc.c2p() * R_2(),
+                wc.c3p() * R_3(),
+            }};
+
+            complex<double> result = 0.0;
+            for (auto & c : contributions)
+            {
+                result += c;
+            }
+
+            // cf. [BBL:1995A], eq. (XVIII.17), p. 153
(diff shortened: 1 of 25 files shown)


eos/meson-mixing/bq-mixing.cc
```diff
@@ -0,0 +1,121 @@
...
+            }};
+
+            complex<double> result = 0.0;
+            for (auto & c : contributions)
+            {
+                result += c;
+            }
+
+            // cf. [BBL:1995A], eq. (XVIII.17), p. 153
+            return 4.0 * g_fermi / std::sqrt(2.0) * power_of<2>(model->ckm_tb() * std::conj(model->ckm_ts()))
```
mreboud, line 95: Shouldn't we have a g_fermi^2 somewhere? I don't understand your prefactor.

dvandyk: The Hamiltonian for F=2 is normalized to G_F^2 * M_W^2, or g'^4 / M_W^2, whatever you prefer. However, one needs to separate the operator from the coefficient and a normalization. The common normalization of the WET in terms of chiral fields is 4 G_F / sqrt(2), which I use here, and which matches the normalization of the Bern basis, except for CKM factors.

The remainder, i.e., G_F * M_W^2, is hidden in the SM value of c1.

mreboud: ahh sorry, it makes sense now

eos/meson-mixing/bq-mixing.cc
```diff
@@ -0,0 +1,121 @@
...
+            R_3(p["B_s<->Bbar_s::R^3"], u),
+            R_4(p["B_s<->Bbar_s::R^4"], u),
+            R_5(p["B_s<->Bbar_s::R^5"], u)
+        {
+            u.uses(*model);
+        }
+
+        complex<double> M_12() const
+        {
+            const double mu = 4.2;
```
cBobeth, line 73: Hard-coded scale - maybe allow to set this via some option to some other value? Of course this would also require to do the RG evolution the R_1,...,5 to that new value of mu. At least this is something to keep in mind for the future. Currently the input for R_1,....,5 should be provided at mu = 4.2 GeV.

eos/meson-mixing/bq-mixing.hh
```diff
@@ -0,0 +1,49 @@
...
+{
+    class BMixing :
+        public ParameterUser,
+        public PrivateImplementationPattern<BMixing>
+    {
+        public:
+            BMixing(const Parameters & parameters, const Options & options);
+            ~BMixing();
+
+            // Observables
```
cBobeth, line 39: I wonder whether also M_12 should be an "observable". Then it can be combined in the future with observables in other decays, i.e. for example all the time-dependent CP asymmetries in all kind of B-> M_1 M_2 decays, which allow for mixing-induced CP violation etc.

eos/utils/wilson_scan_model.hh
```diff
@@ -114,6 +114,36 @@ namespace eos
             virtual WilsonCoefficients<BToS> wilson_coefficients_b_to_s(const double & mu, const std::string & lepton_flavour, const bool & cp_conjugate) const;
     };
 
+    template <>
+    class WilsonScanComponent<components::DeltaB2> :
+        public virtual ModelComponent<components::DeltaB2>
+    {
+        protected:
+            /* b->s Wilson coefficients */
```
cBobeth, line 122: Maybe  as comment something like 

/* sbar b sbar b Wilson coefficients */

?

eos/utils/standard-model.cc
```diff
@@ -695,6 +696,134 @@ namespace implementation
...
+        else
+        {
+            Log::instance()->message("sm_component<deltabs2>.wc", ll_error)
+                << "mu_t > m_t_pole!";
+
+            alpha_s_m_t_pole = QCD::alpha_s(_m_t_pole__deltabs2, _alpha_s_Z__deltabs2, _m_Z__deltabs2, beta5);
+        }
+
+        // calculate m_t at the matching scale in the MSbar scheme
+        const double m_t_msbar_m_t_pole = QCD::m_q_msbar(_m_t_pole__deltabs2, alpha_s_m_t_pole, 5.0);
```
cBobeth, line 757: Use variable nf (in line 723) here instead of 5.0 ?!

eos/utils/standard-model.cc
```diff
@@ -695,6 +696,134 @@ namespace implementation
...
+
+        // calculate dependent inputs
+        const double log_t = std::log(_mu_0__deltabs2 / _m_W__deltabs2);
+        const double xt = power_of<2>(m_t_mu_0 / _m_W__deltabs2);
+        const double xt2 = xt  * xt,
+                     xt3 = xt2 * xt,
+                     xt4 = xt2 * xt2;
+        const double lnxt = std::log(xt), ln2xt = lnxt * lnxt;
+        // GSL and [BBL:1995A] convention for the dilogarithm
+        //  - gsl_sf_dilog(1.0 - x) = L_2(1.0 - x)
```
cBobeth, line 768: Are you sure about the minus sign in front of gsl_sf_dilog(1.0 - x)?

eos/utils/standard-model.cc
```diff
@@ -695,6 +696,134 @@ namespace implementation
...
+
+        /*
+        * Initial scale Wilson coefficients from the top sector, cf. [BBL:1995A], Eqs. (XIII.1) to (XIII.5), pp. 118.
+        */
+
+        // anomalous mass dimension, eq. (XII.7), in the five-flavour scheme
+        const double Nc      = 3.0;
+        const double gamma_0 = 6.0 * (Nc - 1.0) / Nc;
+        const double gamma_1 = (-21.0 + 57.0 / Nc - 19.0 / 3.0 * Nc + 4.0 / 3.0 * nf) * (Nc - 1.0) /(2.0 * Nc);
+        const double d5      = gamma_0  / (2.0 * QCD::beta_function_nf_5[0]);
```
cBobeth, line 779: Why not use def. of beta5[0] in line 725 instead of QCD::beta_function_nf_5[0]? Same in lines 780, 781...

eos/utils/standard-model.hh
```diff
@@ -123,6 +123,37 @@ namespace eos
...
+            UsedParameter _m_W__deltabs2;
+            UsedParameter _m_Z__deltabs2;
+
+            /* Matching scales */
+            UsedParameter _mu_0__deltabs2;
+
+        public:
+            SMComponent(const Parameters &, ParameterUser &);
+
+            /* b->s Wilson coefficients */
```
cBobeth, line 153: Change comment?

eos/meson-mixing/bq-mixing.cc
dvandyk: I agree that this might needs changing at some point, but I would deem it good enough for now.

eos/meson-mixing/bq-mixing.hh
dvandyk: Meril and I think that M_12 is not an observable. To access the phase or any other aspect of M_12, we can simply use the BMixing class in any other class.

cBobeth: Even when it is hidden by the "Implementationpattern"?

eos/utils/wilson_scan_model.hh
dvandyk: Indeed!

eos/utils/standard-model.cc
dvandyk: Done

eos/utils/standard-model.cc
dvandyk: This was not meant as a minus sign, but element of a list of convention differences. In the end, none remained. I will remove the dash.

dvandyk: Done

eos/utils/standard-model.hh
dvandyk: Done

eos/utils/standard-model.cc
dvandyk: Overlooked, thanks! Done.

PR #320, opened 2021-05-26
Check for non-ascii characters before deserializing a constraint

dvandyk:


gubernari (approved):

eos/constraint.cc
```diff
@@ -1606,6 +1611,9 @@ namespace eos
     ConstraintEntry *
     ConstraintEntry::FromYAML(const QualifiedName & name, const std::string & s)
     {
+        if (s.cend() != std::find_if(s.cbegin(), s.cend(), [&] (unsigned char c) -> bool { return (c > 0x7f); }))
```
pluegh, line 1614: Comment on magic constant `0x7f`?

pluegh (commented):
Comment might be sensible

eos/constraint.cc
dvandyk: Done.

PR #324, opened 2021-06-17
[python] Correct issue related to yaml.dump and numpy types

mreboud:
It doesn't really look like pretty code... Any suggestion?

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -135,7 +135,7 @@ class Analysis:
         # add manual constraints to the likelihood
         for constraint_name, constraint_data in manual_constraints.items():
             import yaml
-            yaml_string = yaml.dump(constraint_data)
+            yaml_string = yaml.dump(self._sanitize_manual_input(constraint_data))
             constraint_entry = eos.ConstraintEntry.deserialize(constraint_name, yaml_string)
             constraint = constraint_entry.make(constraint_name, self.global_options)
             self.log_likelihood.add(constraint)
@@ -164,6 +164,24 @@ class Analysis:
     def _par_to_x(self, par):
         """Internal function that rescales the parameters par to [-1, 1]"""
         return np.array([(2 * v - b[0] - b[1]) / (b[1] - b[0]) for v, b in zip(par, self.bounds)])
+
+
+    def _sanitize_manual_input(self, data):
+        """Helper function that converts all entries of a manual_constraint from numpy types to basic python types"""
+        if np.issubdtype(type(data), int):
+            return int(data)
+        if np.issubdtype(type(data), float):
+            return float(data)
+        if np.issubdtype(type(data), str):
+            return str(data)
+
+        if type(data) is dict:
+            return { k: self._sanitize_manual_input(v) for k, v in data.items() }
+        if np.issubdtype(type(data), list):  # data of type dict also matches but they are covered before
+            return list(map(self._sanitize_manual_input, data))
+
+        # all valid cases are covered above
+        raise ValueError("Unexpected entry type {} in manual_constraint".format(type(data)))
 
 
     def clone(self):


python/eos/analysis.py
```diff
@@ -155,6 +155,18 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={},
...
+        """Helper function that cast to float all entries of a manual_constraint"""
+        if type(data) is dict:
+            return { k: self.__sanitize_manual_input(v) for k, v in data.items() }
+        if not np.isscalar(data):
+            return list(map(self.__sanitize_manual_input, data))
+        if np.issubdtype(type(data), int):
+            return int(data)
+        if np.isreal(data):
+            return float(data)
+        return data
```
pluegh, line 168: Might be safer to throw an exception here and not return per default at all? We expect the input numbers to be either int or real, right?

python/eos/analysis.py
```diff
@@ -155,6 +155,18 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={},
         for n in varied_parameter_names - used_parameter_names:
             eos.warn('likelihood does not depend on parameter \'{}\'; remove from prior or check options!'.format(n))
 
+    def __sanitize_manual_input(self, data):
```
pluegh, line 158: Why the double underscore here? Will there be derived classes with their own sanitisation method?

python/eos/analysis.py
```diff
@@ -155,6 +155,18 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={},
...
+        """Helper function that cast to float all entries of a manual_constraint"""
+        if type(data) is dict:
```
pluegh, line 160: Maybe separate with two brief comments the recursion and casting sections.

python/eos/analysis.py
mreboud: No, they can be strings (e.g. observable names)

python/eos/analysis.py
mreboud: Isn't that the syntax to write an internal function in a python class?

pluegh: A single underscore politely indicates that a function is not intended for the user api and double underscores are processed by the interpreter to be class-specific to avoid accidents. Everything would work, but I'd say a single underscore is the most conventional.

python/eos/analysis.py
pluegh: In that case, I'd either go with a comment, or better
```
if isinstance(dat, str):
    return data

# all legal cases are covered above
raise ValueError("...")
```

python/eos/analysis.py
dvandyk: The comment is also incomplete. You do more than cast to ``float``. You also convert iterables to lists.

python/eos/analysis.py
dvandyk: I completely agree.

python/eos/analysis.py
mreboud: That is true...

python/eos/analysis.py
```diff
@@ -155,6 +155,18 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={},
...
+            return { k: self.__sanitize_manual_input(v) for k, v in data.items() }
+        if not np.isscalar(data):
+            return list(map(self.__sanitize_manual_input, data))
```
dvandyk, line 163: Hm. This is not correct. A ``dict`` also yields ``np.isscalar(dict()) == False``. I think we should explicitly handle only the classes we know. Or we use the ``collections`` module.

``` Python
from collections.abc import Collection, Mapping
...
if isinstance(data, Mapping):  # first check for mapping, i.e., dict-like
    return { self.__sanitize_manual_input(k): self.__sanitize_manual_input(v) for k, v in data.items() }
elif isinstance(data, Collection): # non-mapping collections, i.e., list-like
    return list(map(self.__sanitize_manual_input, data))
...
```

@pluegh can you check that? I've had trouble using ``collections.abc.Sequence`` to detect ``numpy.ndarray``, hence the ugliness above.

dvandyk (changes requested):

python/eos/analysis.py
pluegh: I will cover it in unit tests, coming as soon as possible.

mreboud: It is correct since `dict` has been tested just before. But I agree that this is not elegant...

dvandyk (approved):

PR #327, opened 2021-06-20
Fix issue #313

### LINKED ISSUE #313: Varying CKM matrix elements in WilsonScan
Currently, WilsonScan supports adding the Wolfenstein parameters to the likelihood, but not V_{CKM} (e.g. V_{ub} when considering $B\to \pi$). 

Is it a good idea to allow for varying simultaneously V_{CKM} elements and all Wilson coefficients?

@dvandyk suggested to open an issue to hear @mreboud 's and @cBobeth 's opinion.

dvandyk:
This commit fixes #313 by

 - [x] replacing the ``SMComponent<components::CKM>`` with ``CKMScanComponent`` in both ``WilsonScanModel`` and ``ConstrainedWilsonScanModel``
 - [x] setting the required CKM matrix elements (abs and arg) in ``b-decays`` to either the values based on the current default Wolfeinstein params or based on the Wolfenstein params provided in the test case
 - [x] repeating the above for ``rare-b-decays``. However, one test case is stubborn, should use default Wolfenstein values but still fails: ``b-to-kstar-gamma-bfs2004_TEST.cc``, the Bobeth compatibility test

The order of commits is such that the CKM matrix element are fixed first.

### DIFF
--- a/eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
+++ b/eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
@@ -63,10 +63,24 @@ class BToKstarGammaTest :
                 p["b->s::Re{c7'}"] = -0.00659; // m_s(m_b) / m_b(m_b) * Abs{c7} = 85 / 4200 * Abs{c7}
                 p["b->s::c8"] = -0.181;
                 // PDG 2010 CKM parameters
-                p["CKM::A"] = 0.812;
-                p["CKM::lambda"] = 0.22543;
-                p["CKM::rhobar"] = 0.144;
-                p["CKM::etabar"] = 0.342;
+                p["CKM::A"]         =  0.812;
+                p["CKM::lambda"]    =  0.22543;
+                p["CKM::rhobar"]    =  0.144;
+                p["CKM::etabar"]    =  0.342;
+                p["CKM::abs(V_ub)"] =  0.003540950873054711;
+                p["CKM::arg(V_ub)"] = -1.1728563751359748;
+                p["CKM::abs(V_cb)"] =  0.04126451344307112;
+                p["CKM::arg(V_cb)"] =  0.0;
+                p["CKM::abs(V_tb)"] =  0.9991419776905534;
+                p["CKM::arg(V_tb)"] =  0.0;
+                p["CKM::abs(V_td)"] =  0.008576901910577167;
+                p["CKM::arg(V_td)"] = -0.37951557931964897;
+                p["CKM::abs(V_us)"] =  0.22542858674178629;
+                p["CKM::arg(V_us)"] =  0.0;
+                p["CKM::abs(V_cs)"] =  0.9734167680132911;
+                p["CKM::arg(V_cs)"] = -3.119448393424795e-05;
+                p["CKM::abs(V_ts)"] =  0.04051834255894421;
+                p["CKM::arg(V_ts)"] = -3.123445879630718;
                 p["decay-constant::B_d"] = 0.200;
                 p["mass::b(MSbar)"] = 4.2;
                 p["mass::c"] = 1.27;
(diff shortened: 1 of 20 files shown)


eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
```diff
@@ -156,6 +184,14 @@ class BToKstarGammaBobethCompatibilityTest :
             };
 
             Parameters p = Parameters::Defaults();
+            //p["CKM::abs(V_ub)"] =  0.003631275231633653;
```
cBobeth, line 187: I had a look on this. If you use default CKM, then the test seems to fail with the first observable, which is BR. So this is what one would expect.

If you uncomment the values of the CKM elements, then the test for BR runs through, but the test for the next observable fails. This observable is S. 

Can you confirm this?

mreboud: I concur... I am investigating

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
```diff
@@ -63,10 +63,24 @@ class BToKstarGammaTest :
...
+                p["CKM::lambda"]    =  0.22543;
+                p["CKM::rhobar"]    =  0.144;
+                p["CKM::etabar"]    =  0.342;
+                p["CKM::abs(V_ub)"] =  0.003540950873054711;
+                p["CKM::arg(V_ub)"] = -1.1728563751359748;
+                p["CKM::abs(V_cb)"] =  0.04126451344307112;
+                p["CKM::arg(V_cb)"] =  0.0;
+                p["CKM::abs(V_tb)"] =  0.9991419776905534;
+                p["CKM::arg(V_tb)"] =  0.0;
+                p["CKM::abs(V_td)"] =  0.008576901910577167;
```
mreboud, line 76: @dvandyk how did you compute those values? Taking the above A, lambda, etabar and rhobar, I find (small) inconsistencies.

dvandyk: I used the EOS python interface, set A, lambda, rhobar, and etabar, and evaluated. Basically as follows:
```
import eos, cmath
p = eos.Parameters()
p['CKM::A'].set(...)
... # ditto for the rest
m = eos.Model.make('SM', p, eos.Options())
display(abs(m.ckm_ub()), cmath.phase(m.ckm_ub()))
... # ditto for the remaining CKM matrix elements
```

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
cBobeth, line 187: The test works for me, if I use these values

            p["CKM::abs(V_ub)"] =  0.003631275231633653;
            p["CKM::arg(V_ub)"] = -1.210765774253535;
            p["CKM::abs(V_cb)"] =  0.041996951915501936;
            p["CKM::arg(V_cb)"] =  0.0;
            p["CKM::abs(V_tb)"] =  0.9991111398988599;
            p["CKM::arg(V_tb)"] =  0.0;
            p["CKM::abs(V_us)"] =  0.22534851424454577;
            p["CKM::arg(V_us)"] =  0.0;
            p["CKM::abs(V_cs)"] =  0.9734061890640963;
            p["CKM::arg(V_cs)"] = -0.0000330419933699906;
            p["CKM::abs(V_ts)"] =  0.04121211253368258;
            p["CKM::arg(V_ts)"] = -3.1230250245535283;
            p["CKM::abs(V_td)"] =  0.008859566045351227;
            p["CKM::arg(V_td)"] = -0.38266;

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
mreboud: Ok, the difference that I have are probably negligible (there are mostly on the arguments).

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
mreboud: I agree, all CKM components need to be fixed

eos/meson-mixing/bq-mixing.cc
```diff
@@ -111,7 +111,7 @@ namespace eos
     BMixing::delta_m() const
     {
         // cf. [BBL:1995A], eq. (XVIII.16), p. 153
-        return 2.0 * std::abs(_imp->M_12()) / _imp->hbar();
+        return 2.0 * std::abs(_imp->M_12()) / _imp->hbar() * 1.0e-12; // return value in ps^-1
```
mreboud, line 114: This causes an error now, should we just change the test file?

PR #328, opened 2021-06-20
[doc] Improve several aspects

pluegh:
- Streamline the installation instructions
  - most end-users will use the pip installation: now visible at first sight
  - structure sections via specific use cases (OS -> installation method) instead of technical solution (Installation method -> platform specific)
- Change descriptions to action instead of intent (e.g. use `pip3` instead of PyPi); easier and more concrete
- Fix typos using a language tool

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -2,21 +2,32 @@
 Installation
 ############
 
-If you intend to use EOS without any modifications, then you should follow the steps outlined in `Installing with a Package Manager`_
-or `Installing with the Python Package Index (PyPI)`_.
-If you intend to modify EOS for your personal purpose, then you must follow the steps outlined in `Installing from Source`_.
+We distribute EOS as packaged software for end-users and also as source code for developers.
 
 
-*********************************
-Installing with a Package Manager
-*********************************
+*****
+Linux
+*****
+
+Installation with `pip3`
+========================
+
+This is the recommended installation method using the 'package installer for Python' `pip3`:
+
+::
+
+  pip3 install --user eoshep
+
+You can now use the EOS Python module via ``import eos``. In addition, some command line scripts have become available.
+To update EOS, run ``pip3 install --user --upgrade eoshep``.
 
 
-Ubuntu Linux
-============
+Installation with `apt` (Ubuntu, Debian and Siblings)
+=====================================================
 
-Pre-built binary packages of the most recent release of EOS for Ubuntu Linux LTS release 18.04, are available through the Packagecloud web service.
-They can be installed through the following steps:
+EOS is available through the Packagecloud web service.
+The installation requires several steps:
+
 
 1. Create the file ``/etc/apt/sources.list.d/eos.list`` by running the following command in a shell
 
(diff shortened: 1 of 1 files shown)


doc/installation.rst
```diff
@@ -262,6 +256,7 @@ The recommended configuration is achieved by running the following command
       --enable-pmc
 
 If the ``configure`` script finds any problems with your system, it will complain loudly.
+The flag ``with-boost-python-suffix`` might be necessary, depending on the installation of Python and BOOST.
```
pluegh, line 259: At least this is the case on macOS, where you install `boost-python3` via brew.

pluegh: Specifically, check for the available library version with `ls /usr/local/lib/libboost_python*`, resulting in e.g. `/usr/local/lib/libboost_python39.a`. Then use `--with-boost-python-suffix=39`.

doc/installation.rst
```diff
@@ -2,21 +2,31 @@
...
-Installing with a Package Manager
-*********************************
+*****
+Linux
+*****
+
+Installation with `pip3`
+========================
+
+This is the recommended installation method using the package installer `pip3`:
```
dvandyk, line 15: installer -> manager? That's why they are usually called!

mreboud (commented):
Looks good to me, can we add a paragraph on conda, spyder and all these python utilities that make installing EOS quite complex?

doc/installation.rst
```diff
@@ -2,21 +2,31 @@
...
 
+::
+
+  pip3 install --user eoshep
 
-Ubuntu Linux
-============
+You can now use EOS via the Python module or as system commands.
```
dvandyk, line 21: You can now use the EOS Python module via ``import eos``. In addition, some command line scripts have become available (link to a new section on command-line scripts that we still need to write)

doc/installation.rst
```diff
@@ -2,21 +2,31 @@
...
+
+
+Installation with `apt` (Ubuntu)
```
dvandyk, line 24: Ubuntu/Debian and similar?

doc/installation.rst
```diff
@@ -38,75 +48,59 @@ You might need to install the ``curl`` program first, if not already installed.
 
   apt-get install curl
 
-3. Install the pre-built binary packages and its dependencies by running the following commands
+3. Install the pre-built binary packages and their dependencies by running the following commands
```
dvandyk, line 51: "package" (singular) and "its" dependencies, right? Or do we still need Minuit2?

doc/installation.rst
```diff
@@ -38,75 +48,59 @@ You might need to install the ``curl`` program first, if not already installed.
...
   apt-get install eos
 
-4. On older version of Ubuntu, you might need to update your installation of the Matplotlib Python package.
-   We recommend to install the most up-to-date Python package with from the Python Package Index, using
+4. On older versions of Ubuntu, you might need to update your installation of the Matplotlib Python package.
+   We recommend installing the most up-to-date Python package from the Python Package Index, using
 
 ::
 
   pip3 install --user h5py matplotlib numpy pypmc pyyaml scipy
```
dvandyk, line 63: Should we use ``--update`` here? Does it fail, if the packages are not yet installed?

doc/installation.rst
```diff
@@ -38,75 +48,59 @@ You might need to install the ``curl`` program first, if not already installed.
...
 
   brew tap eos/eos
   brew install --HEAD eos
 
-We recommend to install the most up-to-date Python package with from the Python Package Index, using
+We recommend installing the most up-to-date Python package from the Python Package Index, using
 
 ::
 
   pip3 install --user h5py matplotlib numpy pypmc pyyaml scipy
```
dvandyk, line 85: See above on ``--update``

doc/installation.rst
```diff
@@ -38,75 +48,59 @@ You might need to install the ``curl`` program first, if not already installed.
...
-Installing from Source
-**********************
+*****************************
+Build and Install from Source
+*****************************
 
 The following instructions explain how to install EOS from source on a Linux-based operating system,
-such as Debian or Ubuntu, or MacOS X.
-Other flavors of Linux or UNIX-like operating systems will likely work as well.
+such as Debian, Ubuntu, or MacOS X.
```
dvandyk, line 94: macOS is not a Linux-based operating system.

doc/installation.rst
```diff
@@ -167,7 +161,7 @@ FFTW3
 If you have administrator access to the computers on which you use EOS,
 we recommend you install the above packages via your system's software management system.
 
-On a Debian/Ubuntu based operating system you can install the prerequisite software with the ``APT`` package management system,
+On a Debian/Ubuntu-based operating system you can install the prerequisite software with the ``APT`` package management system,
```
dvandyk, line 164: You used ``apt`` above, while here we use ``APT``. Lowercase is probably better, since it matches the actual command.

doc/installation.rst
dvandyk: I agree with the second comment.

dvandyk (changes requested):

doc/installation.rst
mreboud: In principle Minuit2 is not needed anymore!

doc/installation.rst
mreboud: no, if the package is already installed, it either updates it or sends a message like "latest version already installed".

doc/installation.rst
pluegh: It's a specific (not generic) name, but we can set it in quotation marks.

doc/installation.rst
pluegh: I like 'siblings' for derivatives of Debian.

doc/installation.rst
pluegh: For what we describe, the first installation of eos, using no flags does the job and has no "side effects".

I propose to add the explicit comment instead that EOS can be updated with the same command except for the additional flag `--upgrade` or `-U` for short.

pluegh: I agree!

pluegh: Except that the flag is ``--upgrade``

doc/installation.rst
pluegh: "The following instructions explain how to install EOS from source on a Linux-based operating system, such as Debian or Ubuntu, and macOS."

doc/installation.rst
```diff
@@ -252,7 +247,7 @@ Next, you must configure the EOS build using the ``configure`` script.
 To use the EOS Python interface you must pass ``--enable-python`` to the call ``configure``.
 The default is \cli{--disable-python}.
```
dvandyk, line 248: While you are at it, can you fix the LaTeX macro used here? It should be ``--disable-python``

doc/installation.rst
```diff
@@ -296,7 +297,7 @@ For ``BASH``, which is the default Debian/Ubuntu shell, add the following lines
   export PYTHONPATH+=":$PREFIX/lib/python3.6/site-packages"
 
 Note that in the above ``python3.6`` piece must be replaced by the appropriate Python version with which EOS was built.
```
dvandyk, line 299: Note that in the above **the** ``python3.6`` piece ...

dvandyk (changes requested):
Two little changes.

PR #337, opened 2021-06-25
[python] Sanitize the weights in the computations of the perplexity

### LINKED ISSUE #336: Nan in the computation of perpexlity
Due to numerical approximations, components with very small weighs yield errors in PMC sampling: `RuntimeWarning: divide by zero encountered in log` due to the line
https://github.com/eos/eos/blob/edce93909e4632fb646876063002ec54a5de17e0/python/eos/analysis.py#L407
When the normalized weight is zero, the contribution to the entropy should simply be zero. This is supposed to be accounted for by the previous lines, where non-positive weights are set to a small value `eps`:https://github.com/eos/eos/blob/edce93909e4632fb646876063002ec54a5de17e0/python/eos/analysis.py#L404-L405
I think that this step should be performed **after** the normalization and not **before**.
Idem here: https://github.com/eos/eos/blob/edce93909e4632fb646876063002ec54a5de17e0/python/eos/analysis.py#L432

@dvandyk Do you agree?

mreboud:
I tested this locally, should we think about a test case for this?
Closes #336.

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -184,6 +184,24 @@ class Analysis:
         raise ValueError("Unexpected entry type {} in manual_constraint".format(type(data)))
 
 
+    @staticmethod
+    def _perplexity(weights):
+        """Helper function that computes the perplexity of an array of weights"""
+        # sum positive finite weights only
+        weights_sum = np.sum(weights,
+            where = np.logical_or(weights > 0, np.isfinite(weights)))
+        if weights_sum <= 0:
+            return 0.
+        else:
+            normalized_weights = weights / weights_sum
+            # mask negative and nan weights
+            normalized_weights = np.ma.MaskedArray(normalized_weights, copy=False,
+                mask=(np.logical_or(normalized_weights <= 0, np.isnan(normalized_weights))))
+            entropy = - np.sum(normalized_weights * np.log(normalized_weights.filled(1.0)))
+            perplexity = np.exp(entropy) / len(normalized_weights)
+            return perplexity
+
+
     def clone(self):
         """Returns an independent instance of eos.Analysis."""
         return eos.Analysis(**self.init_args)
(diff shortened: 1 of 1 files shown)


python/eos/analysis.py
```diff
@@ -400,19 +400,29 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
             generating_components.append(origins)
             samples = sampler.samples[:]
             last_weights = np.copy(sampler.weights[-1][:, 0])
-            for i, w in enumerate(last_weights):
-                if w <= 0 or np.isnan(w):
-                    last_weights[i] = eps
-            normalized_last_weights = last_weights / np.sum(last_weights)
+            normalized_last_weights = last_weights / np.sum(
```
dvandyk, line 403: This fails if none of the weights is positive, which is a pathological corner case. Should we test for it and skip the computation of the perplexity in that case?

python/eos/analysis.py
```diff
@@ -425,10 +435,15 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
-        normalized_weights = adjusted_weights / np.sum(adjusted_weights)
+        normalized_weights = adjusted_weights / np.sum(
+            adjusted_weights,
+            where = np.logical_or(adjusted_weights > 0, np.isfinite(adjusted_weights))
+        ) # sum positive finite weights only
+        normalized_weights = np.where(
+                np.logical_or(normalized_weights <= 0, np.isnan(normalized_weights)),
+                eps,
+                normalized_weights
+        ) # replace negative and nan weights by eps
```
dvandyk, line 446: Seeing the same code thrice, should we make it a separaste function? Also, is there any benefit to using [pypmc's function ```pypmc.tools.perp```](https://pypmc.github.io/api.html#pypmc.tools.convergence.perp)?

A quick look at that source code seems better for the zero weights, but has not a workable solution for NaN and negative weights.

dvandyk (commented):
Could be merged as is, might be improved.

dvandyk:
Also, there is a typo in the commit message: ```perpexlity``` -> ```perplexity```

python/eos/analysis.py
mreboud: Indeed the pypmc's `perp` doesn't treat nan weights, that are even propagated due to the normalization. We could write our own `perp` function.

python/eos/analysis.py
mreboud: Yes, I set perplexity to zero in this case

dvandyk:
Please go ahead with merging if the tests pass.

PR #340, opened 2021-07-06
Small python changes

mreboud:
This PR:
  - Fixes a small bug in the labels of plotter.py
  - Scales the sampler proposal back to the initial parameter space in pmc-sample
  - Allows to fix a perplexity threshold to stop PMC adaptations
  - Allows to return final step samples only in pmc-sample

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -366,7 +366,7 @@ class Analysis:
             return(parameter_samples, weights, np.array(observable_samples))
 
 
-    def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.random.mtrand):
+    def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.random.mtrand, return_final_only=True):
         """
         Return samples of the parameters and log(weights)
 
@@ -377,6 +377,7 @@ class Analysis:
         :param steps: Number of adaptation steps.
         :param final_N: Number of samples that shall be drawn after all adaptation steps.
         :param rng: Optional random number generator (must be compatible with the requirements of pypmc.sampler.importance_sampler.ImportancSampler)
+        :param return_final_only: If set to True, only returns the samples and weights of the final sampling step, after all adaptations have finished.
 
         :return: A tuple of the parameters as array of length N = pre_N * steps + final_N, the (linear) weights as array of length N, and the
             final proposal function as pypmc.density.mixture.MixtureDensity.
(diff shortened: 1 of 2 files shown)


python/eos/analysis.py
```diff
@@ -434,6 +434,14 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
         samples = np.apply_along_axis(self._x_to_par, 1, sampler.samples[:])  # Rescale the parameters back to their original bounds
         weights = sampler.weights[:][:, 0]
+        # rescale proposal components back to their physical bounds
+        for component in sampler.proposal.components:
+            rescaled_mu = self._x_to_par(component.mu)
+            rescaled_sigma = np.array([[
+                component.sigma[i, j] * (bj[1] - bj[0]) * (bi[1] - bi[0]) / 4 for j, bj in enumerate(self.bounds)
+                ] for i, bi in enumerate(self.bounds)])
+            component.update(rescaled_mu, rescaled_sigma)
+
```
dvandyk, line 450: Good catch!

python/eos/analysis.py
```diff
@@ -377,6 +377,7 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
         :param steps: Number of adaptation steps.
         :param final_N: Number of samples that shall be drawn after all adaptation steps.
         :param rng: Optional random number generator (must be compatible with the requirements of pypmc.sampler.importance_sampler.ImportancSampler)
+        :param return_final_only: If set to True, only returns the samples and weights of the final adaptation step.
```
dvandyk, line 380: The final step is a pure evaluation step without adaptations, right? So this is slightly off.

How about:
``If set to True, only returns the samples and weights of the final sampling step, after all adaptations have finished.`` ?

dvandyk (approved):

python/eos/analysis.py
```diff
@@ -428,12 +432,31 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
+            rescaled_sigma = np.array([[
+                component.sigma[i, j] * (bj[1] - bj[0]) * (bi[1] - bi[0]) / 4 for j, bj in enumerate(self.bounds)
+                ] for i, bi in enumerate(self.bounds)])
+            component.update(rescaled_mu, rescaled_sigma)
+
+        # rescale the samples back to their physical bounds
+        if return_final_only:
+            # only returns the final_N final samples
+            samples = np.apply_along_axis(self._x_to_par, 1, sampler.samples[-final_N:])
+            weights = sampler.weights[:][:, 0]
```
mreboud, line 455: This was supposed to be
```python
sampler.weights[-final_N:][:, 0]
```

PR #341, opened 2021-07-07
Python tests

mreboud:
This PR adds the possibility to write tests for Python implementations.

### DIFF
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -63,7 +63,8 @@ _eos_la_LDFLAGS = -module -avoid-version -export-symbols-regex PyInit__eos @BOOS
 _eos_la_LIBADD = $(top_builddir)/eos/libeos.la $(top_builddir)/eos/utils/libeosutils.la -lboost_python$(BOOST_PYTHON_SUFFIX)
 
 TESTS = \
-	eos_TEST.py
+	eos_TEST.py \
+	eos/analysis_TEST.py
 
 EXTRA_DIST += $(TESTS)
 
--- a/python/Makefile.am
+++ b/python/Makefile.am
@@ -21,7 +21,7 @@ EXTRA_DIST = \
 AM_TESTS_ENVIRONMENT = \
 			 export EOS_TESTS_CONSTRAINTS="$(top_srcdir)/eos/constraints"; \
 			 export EOS_TESTS_PARAMETERS="$(top_srcdir)/eos/parameters"; \
-			 export PYTHONPATH="$(top_builddir)/python/.libs/:$(top_builddir)/python";
+			 export PYTHONPATH="$(top_builddir)/python/.libs/:$(top_srcdir)/python";
 
 TEST_EXTENSIONS = .py
 PY_LOG_COMPILER = $(PYTHON)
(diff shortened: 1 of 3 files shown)


python/Makefile.am
```diff
@@ -63,7 +63,8 @@ _eos_la_LDFLAGS = -module -avoid-version -export-symbols-regex PyInit__eos @BOOS
 _eos_la_LIBADD = $(top_builddir)/eos/libeos.la $(top_builddir)/eos/utils/libeosutils.la -lboost_python$(BOOST_PYTHON_SUFFIX)
 
 TESTS = \
-	eos_TEST.py
+   	eos_TEST.py \
```
mreboud, line 66: Ah I forgot to correct the 4 spaces here...

python/eos/analysis_TEST.py
```diff
@@ -0,0 +1,80 @@
...
+        for case in input_output_cases:
+            # check value and type
+            self.assertEqual(
+                eos.Analysis._sanitize_manual_input(case['dirty']),
+                case['clean']
+            )
+            self.assertEqual(
+                type(eos.Analysis._sanitize_manual_input(case['dirty'])),
+                type(case['clean'])
+            )
```
pluegh, line 76: I believe the assertEquals are not performing a deep comparison of the dicts. At least for the type check, we only verify that a dict is going in and out.

We should in any case add the inverse check to also test our test method, i.e. check that the cleaned output is not equal to the dirty input, as long as we only check cases where we expect sanitisation.

pluegh (changes requested):

python/eos/analysis_TEST.py
```diff
@@ -0,0 +1,80 @@
...
+            'likelihood': [
+                'B->D::f_++f_0@HPQCD2015A',
+                'B->D::f_++f_0@FNALMILC2015B',
+                'B^0->D^+e^-nu::BRs@Belle-2015A',
+                'B^0->D^+mu^-nu::BRs@Belle-2015A'
+            ]
+        }
+
+        # Test analysis definition
+        analysis = eos.Analysis(**analysis_args)
```
pluegh, line 28: This call fails in our CI pipeline. Ideas?

python/eos/analysis_TEST.py
mreboud: For the types I agree, that is why I am also testing `str` and `float`. For the second point I also agree, I will add this test

python/eos/analysis_TEST.py
mreboud: I do know ^^
This is weird... Are we sure `make install` is ran before we test?

pluegh: @dvandyk should be able to settle this quickly :)

I think we don't install, but set the paths to the build dirs. Not sure if these are correct or if it's something else.

python/eos/analysis_TEST.py
mreboud: Pffiou this was not trivial since you cannot unittest both the type and the value in one shot... I finally solved this by testing that the initial yaml strings were different while the sanitize yaml strings are the same. In the end this is even a better test since this is the aim of the helper function...

python/eos/analysis_TEST.py
mreboud: Indeed `make install` is not the culprit, it is locally not even needed... so I don't know yet

Issue #69, opened 2018-03-26
GSL{Vector,Matrix,RNG}Ptr


dvandyk:
Following the discussion of PR #66, it would be nice to use RAII for handling pointers to GSL objects.
Do we want something like here: https://github.com/dvandyk/eos/commit/b1c942737e00cac15c73791d8e7c61261be3297c ??

The only problem I see is that we would need either
 - to include ```gsl_rng.h```, ```gsl_vector.h```, ```gsl_matrix.h``` whereever we use one of them,
 - or have separate C++ headers for each.

And I think in some private feature branch we're already using complex-valued GSL matrices and vectors,
so that's another layer of GSL objects.

fredRos:
Wouldn't it suffice to `#include <gsl_interface.hh>` and it loads the required headers? It would be OK for me. I agree, the implementation for every gsl data type we use is a  bit lengthy but it provides a clean solution at the call site

dvandyk:
@mreboud Would you like to try your hand on this one?

mreboud:
I will be happy to try!

Issue #338, opened 2021-07-05
Add support for mixture densities as part of the log-likelihood


dvandyk:
This will enable to use the output of a PMC fit in case of large final perplexity (> ~90%).

Steps:
 - in ``eos/statistics/log-likelihood.hh``
     - [ ] document ``LogLikelihood::Mixture``
 - in ``eos/statistics/log-likelihood.cc``
     - [ ] fix computation of ``MixtureBlock::number_of_observations()``: this should be the number of observations of each component, not their sum!
 - in ``eos/constraint.hh``
     - [ ] add ``MixtureConstraintEntry`` class
     - [ ] (de)serialization should copy MultivariateGaussian as much as possible, but include
``` YAML
type: Mixture
dim: ... # special case for dim = 1
observables: [...] # as in MultivariateGaussian(Covariance)
weights: [....] # normalized to 1
components:
    - type: MultivariateGaussian(Covariance)
      means: [...]
      covariance: [...]
    - type: MultivariateGaussian(Covariance)
      means: [...]
      covariance: [...]
```
 - in [``eos/constraint.cc``](https://github.com/eos/eos/blob/9380e13a26cb0967350ac0c1d2e8e3bc347c01bf/eos/constraint.cc#L1637)
    - [ ] register the new ``MixtureConstraintEntry`` for deserialization

PR #342, opened 2021-07-09
Fixes issue #69 (GSL interface) and #338 (MixtureConstraintEntry)

dvandyk:


eos/constraint_TEST.cc
```diff
@@ -1134,6 +1134,139 @@ class ConstraintDeserializationTest :
...
+                TEST_CHECK_EQUAL(1, blocks.size());
+
+                Parameters p = Parameters::Defaults();
+                LogLikelihood llh(p);
+                llh.add(c);
+
+                // evaluation off mode
+                p["mass::b(MSbar)"] = 4.6;
+                p["mass::c"] = 1.3;
+                TEST_CHECK_NEARLY_EQUAL(llh(), -4.779399451, 1e-8);
```
mreboud, line 1267: This is a value you get from the code I guess? Or could you check it against two multivariate Gaussian distributions?

eos/utils/gsl-interface_TEST.cc
```diff
@@ -0,0 +1,90 @@
...
+        virtual void run() const
+        {
+            // default construction
+            {
+                GSLMatrixPtr();
+            }
+
+            // allocation with alloc
+            {
+                GSLMatrixPtr p(gsl_matrix_alloc(10, 10));
```
mreboud, line 44: If I understand correctly this is also indirectly testing the destructors, right?

mreboud (approved):

eos/constraint_TEST.cc
dvandyk: Did that!

mreboud: Excellent :)

eos/utils/gsl-interface_TEST.cc
dvandyk: Yes. The ``MatrixDeleter`` is used.

Issue #335, opened 2021-06-24
WCxf import not working


peterstangl:
I am getting an error if I try to load Wilson coefficients from a WCxf object (using the current development version of eos). This can be reproduced e.g. by the following code:
```python
import eos
from wilson import Wilson

w = Wilson({'cbtaunutau::cSL': 0.3,}, scale=4.8, eft="WET", basis="EOS")
eos.Parameters.FromWCxf(w.wc)
```
 The error message is
```python
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-2-991fa46dbddc> in <module>
      3 
      4 w = Wilson({'cbtaunutau::cSL': 0.3,}, scale=4.8, eft="WET", basis="EOS")
----> 5 eos.Parameters.FromWCxf(w.wc)

~/.local/lib/python3.8/site-packages/eos/parameter.py in FromWCxf(wc)
    109         parameters = _Parameters.Defaults()
    110         for name in wc.dict:
--> 111             qn     = eos.QualifiedName(name)
    112             prefix = qn.prefix_part()
    113             coeff  = qn.name_part()

NameError: name 'eos' is not defined
```
I guess this can be fixed by simply replacing
https://github.com/eos/eos/blob/edce93909e4632fb646876063002ec54a5de17e0/python/eos/parameter.py#L111
by
```python
            qn     = QualifiedName(name)
```

mreboud:
Thanks @peterstangl. I can reproduce your error with a different Wilson coefficient. For `'cbtaunutau::cSL'` you are also using the development version of wcxf, right?

dvandyk:
Hi @peterstangl,

this error appears, since the wcxf bases are not versioned. I had to start at one end :-)
wcxf-import with the basis recently merged will be working again with the next release.

peterstangl:
@mreboud @dvandyk yes, sorry it maybe was not the best example since I was using the development version of wcxf as well and the basis differs from the released one. But this error is independent of the Wilson coefficient and should be reproducible with any Wilson coefficients no matter which version of the basis is used. The problem is that `eos.QualifiedName` cannot be found since `eos` is not available in the namespace. But `QualifiedName` actually is available since it is imported at the beginning of the file, so changing line 111 as I wrote above should solve this problem.

mreboud:
[...]

PR #343, opened 2021-07-10
Fix eos.Parameter.FromWCxf

dvandyk:
- depends on merging wilson-eft/wilson#71
 - add test case

### DIFF
--- a/python/eos/parameter.py
+++ b/python/eos/parameter.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2017-2020 Danny van Dyk
+# Copyright (c) 2017-2021 Danny van Dyk
 #
 # This file is part of the EOS project. EOS is free software;
 # you can redistribute it and/or modify it under the terms of the GNU General
@@ -99,7 +99,23 @@ class Parameters(_Parameters):
 
 
     @staticmethod
-    def FromWCxf(wc):
+    def FromWCxf(w):
+        """
+        Imports Wilson coefficients into an eos.Parameters object from a wilson.Wilson object.
+
+        :param w: WET Wilson coefficients in the EOS basis.
+        :type w: wilson.Wilson
+        """
+
+        # extract wc data
+        wc = w.wc
+
+        if wc.basis != 'EOS':
+            raise ValueError('Wilson coefficients must be passed to this function in the EOS basis')
+
+        # Needs to be revisited, once EOS support WET-4 or WET-3 bases
+        if wc.scale != 4.2:
+            raise ValueError('Wilson coefficients must be passed to this function at the reference scale 4.2 GeV')
 
         # the following coefficients are treated as real-valued in EOS
         real_coeffs = [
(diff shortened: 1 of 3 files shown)


python/eos/parameter.py
```diff
@@ -99,24 +99,40 @@ def _repr_html_(self):
...
+
+        # extract wc data
+        wc = w.wc
+
+        if wc.basis != 'EOS':
+            raise ValueError('Wilson coefficients must be passed to this function in the EOS basis')
+
+        # Needs to be revisited, once EOS support WET-4 or WET-3 bases
+        if wc.scale != 4.2:
+            raise ValueError('Wilson coefficients must be passed to this function at the reference scale 4.2 GeV')
```
mreboud, line 119: Should we open an issue to do this automatically?

python/eos/parameter.py
```diff
@@ -99,24 +99,40 @@ def _repr_html_(self):
...
         parameters = _Parameters.Defaults()
-        for name in wc.dict:
-            qn     = eos.QualifiedName(name)
+        for name, value in wc.dict.items():
+            qn     = QualifiedName(name)
             prefix = qn.prefix_part()
             coeff  = qn.name_part()
-            value  = wc.dict[name]
-            if (not value.imag == 0) and (coeff in real_coeffs):
-                raise ValueError('WC {0} does not support non-zero imaginary part'.format(name))
```
mreboud, line 116: I remember that Peter Stangl mentioned small imaginary part due to running and matching https://github.com/eos/eos/issues/335#issuecomment-874747524

python/eos/parameter.py
```diff
@@ -99,24 +99,40 @@ def _repr_html_(self):
...
 
             # Add values provided by WCxf to EOS central (SM) values
-            if str(coeff) in real_coeffs:
+            if (prefix == 'b->s') and (coeff in real_coeffs):
+                if (abs(value.imag) < 1.0e-10):
```
mreboud, line 133: Shouldn't this be `> 1e-10`?

python/eos/parameter_TEST.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ClassMethodTests(unittest.TestCase):
+
+    def test_importing_from_wcxf(self):
+
+        from wilson import Wilson
+
+        inputs = {
+	    'cbtaunutau::cSL':  0.3 + 0.1j,
+            'ubenue::cVL':     -0.1 + 0.2j,
```
mreboud, line 14: Indentations look weird, no?

python/eos/parameter_TEST.py
```diff
@@ -0,0 +1,21 @@
...
+        }
+        w = Wilson(inputs, scale=4.2, eft="WET", basis="EOS")
+        p = eos.Parameters.FromWCxf(w)
+
```
mreboud, line 18: Can we check the values of `p`, to ensure correct import? This is more for future updates of the code...

mreboud (changes requested):

python/eos/parameter.py
dvandyk: And this is why a test case is needed!

python/eos/parameter_TEST.py
dvandyk: Yes, will fix it.

python/eos/parameter_TEST.py
dvandyk: Yes. To be kept in mind: Wilson/WCxf parametrizes the BSM contributions. We still have to add the SM contributions on top of that.

python/eos/parameter.py
dvandyk: Hmm, I kind of like that the user must be aware of this issue. It should not be hardship for the user to use match_run correctly.

python/eos/parameter.py
dvandyk: Indeed, see below.

python/eos/parameter.py
dvandyk: Added to the test case.

dvandyk:
I am now updating the CI/CD images to include wilson.

mreboud (approved):

Issue #64, opened 2018-03-22
Handle GSL error gracefully


dvandyk:
Presently EOS exits with ```abort()``` when the GSL error handler is invoked. We should find a way to change that.

fredRos:
What could you prefer? Throw an exception like this

```c++
class GSLError :
        public Exception
```

Or should it not inherit from Exception?

dvandyk:
Inheriting from ```Exception``` is perfectly fine.

fredRos:
The [gsl docs](https://www.gnu.org/software/gsl/doc/html/err.html) explain this well. We have a number of places in the code where we turn of the default error handler. Look for 
```c++
gsl_error_handler_t * default_gsl_error_handler = gsl_set_error_handler_off();
```
We could set a custom handler globally, for example with a static variable during initialization of the library. But I see a problem with multiple threads because the error handling is defined globally within GSL. For example, if some thread explicitly checks errors, it turns off the handler. A concurrently executing thread relies on the error handler throwing exceptions but no exception can be thrown. 

For our applications, it seems that the error handler is deactivated only during ctors or initialization routines so we may get away with it w/o thread race conditions.

dvandyk:
The GSL documentation recommends to switch off the default error handler or to replace it with a custom one.

Coming back to this issue, my plan is now to:
 - [ ] audit the code to make sure the return values of calls to ```gsl_*``` are inspected;
 - [ ] create a new exception class ```GSLError``` that can be thrown in such cases in which we cannot continue running;
 - [ ] disable the GSL error handler within a free-standing function with ```__attribute__((constructor))```, to be included in ```libeos.so```.

mreboud:
@dvandyk by "inspected", do you mean (1) to ensure that the pointer exists and (2) that the gsl objects are properly unallocated?

Issue #70, opened 2018-03-26
GSL deprecated gsl_linalg_cholesky_decomp


dvandyk:
Might be a good idea to
 - [ ] move to ```gsl_linalg_cholesky_decomp1```,
 - [ ] make use of the triangular shape of the cholesky decomposition output.

Related to PR #66.

PR #344, opened 2021-07-10
Fix issues #64 and #70

dvandyk:


dvandyk:
FFS. In the ``manylinux2014`` image the GSL version is so old (< 2.2) that the new function ``gsl_linalg_cholesky_decomp1`` is not available.

Suggestions?

mreboud:
`gsl_linalg_cholesky_decomp` without the 1 is there for backward compatibility so it should be linked to `gsl_linalg_cholesky_decomp1` when it exists and the old method otherwise, no?

dvandyk:
> `gsl_linalg_cholesky_decomp` without the 1 is there for backward compatibility so it should be linked to `gsl_linalg_cholesky_decomp1` when it exists and the old method otherwise, no? see https://github.com/ampl/gsl/blob/a367169b777cdaf98a7b54a91daac007f56875b6/linalg/cholesky.c#L55

See the next line! ``It will be removed in a future release.``

I think I'll add a ``#define GSL_LINALG_CHOLESKY_DECOMP`` that points to the old method if the GSL version is 2.1 or older, and to the new method otherwise.

mreboud:
> > `gsl_linalg_cholesky_decomp` without the 1 is there for backward compatibility so it should be linked to `gsl_linalg_cholesky_decomp1` when it exists and the old method otherwise, no? see https://github.com/ampl/gsl/blob/a367169b777cdaf98a7b54a91daac007f56875b6/linalg/cholesky.c#L55
> 
> See the next line! `It will be removed in a future release.`
> 
> I think I'll add a `#define GSL_LINALG_CHOLESKY_DECOMP` that points to the old method if the GSL version is 2.1 or older, and to the new method otherwise.

Yes I don't see any other solution...

PR #346, opened 2021-07-15
Unify plot args

pluegh:
Unify Plotter.Observable items 'kinematic' and 'parameter' to a single key 'variable' (which was also still implemented before as a proxy for 'kinematic'). This is a breaking change! We need to check and likely modify any documentation on this.

All changes are covered with unit tests, where test infrastructure was added where needed. Anticipated user errors are caught and reported with meaningful (expressive) error messages to improve the user experience.

### DIFF
--- a/python/eos/observable.py
+++ b/python/eos/observable.py
@@ -84,3 +84,12 @@ class Observables(_Observables):
         result += r'</table>'
 
         return(result)
+
+    @staticmethod
+    def _get_obs_entry(name):
+        obs = Observables()[name]
+        if obs is None:
+            raise ValueError("Observable with name '" + name + "' is not known")
+        else:
+            return obs
+
--- a/python/eos/observable_TEST.py
+++ b/python/eos/observable_TEST.py
@@ -1,0 +1,23 @@
+import unittest
+import eos
+
+class StaticMethodTests(unittest.TestCase):
+
+    def test_get_obs_entry_0(self):
+        "name is valid"
+
+        valid_name = 'B->Dlnu::dBR/dq2;l=e,q=d'
+        eos.Observables._get_obs_entry(valid_name)
+
+
+    def test_get_obs_entry_1(self):
+        "name is invalid"
+
+        invalid_name = 'prefix::Philipp'
+        with self.assertRaisesRegex(ValueError, "Observable with name 'prefix::Philipp' is not known"):
+            eos.Observables._get_obs_entry(invalid_name)
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=5)
+
--- a/python/eos/parameter.py
+++ b/python/eos/parameter.py
@@ -128,3 +128,11 @@ class Parameters(_Parameters):
         return parameters
 
 
+    @staticmethod
+    def _assert_valid_name(name):
+        "Raise error if 'name' is not a valid EOS paramter name"
+
+        valid_names = [p.name() for p in Parameters()]
+        if name not in valid_names:
+            raise ValueError("Parameter name '" + name + "' is not known")
+
--- a/python/eos/parameter.py
+++ b/python/eos/parameter.py
@@ -127,12 +127,3 @@ class Parameters(_Parameters):
 
         return parameters
 
-
-    @staticmethod
-    def _assert_valid_name(name):
-        "Raise error if 'name' is not a valid EOS paramter name"
-
-        valid_names = [p.name() for p in Parameters()]
-        if name not in valid_names:
-            raise ValueError("Parameter name '" + name + "' is not known")
-
--- a/python/eos/parameter_TEST.py
+++ b/python/eos/parameter_TEST.py
@@ -1,0 +1,21 @@
+import unittest
+import eos
+
+class StaticMethodTests(unittest.TestCase):
+
+    def test_assert_valid_name_0(self):
+        "Value of 'name' is valid"
+
+        eos.Parameters._assert_valid_name('mass::mu')
+
+
+    def test_assert_valid_name_1(self):
+        "Value of 'name' is invalid (for small-scale physics)"
+
+        with self.assertRaisesRegex(ValueError, "Parameter name 'mass::your_mum' is not known"):
+            eos.Parameters._assert_valid_name('mass::your_mum')
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=5)
+
--- a/python/eos/parameter_TEST.py
+++ b/python/eos/parameter_TEST.py
@@ -1,20 +1,7 @@
 import unittest
 import eos
 
-class StaticMethodTests(unittest.TestCase):
-
-    def test_assert_valid_name_0(self):
-        "Value of 'name' is valid"
-
-        eos.Parameters._assert_valid_name('mass::mu')
-
-
-    def test_assert_valid_name_1(self):
-        "Value of 'name' is invalid (for small-scale physics)"
-
-        with self.assertRaisesRegex(ValueError, "Parameter name 'mass::your_mum' is not known"):
-            eos.Parameters._assert_valid_name('mass::your_mum')
-
+# no tests yet
 
 if __name__ == '__main__':
     unittest.main(verbosity=5)
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -1,5 +1,6 @@
 # Copyright (c) 2018 Frederik Beaujean
 # Copyright (c) 2017, 2018, 2021 Danny van Dyk
+# Copyright (c) 2021 Philip Lüghausen
 #
 # This file is part of the EOS project. EOS is free software;
 # you can redistribute it and/or modify it under the terms of the GNU General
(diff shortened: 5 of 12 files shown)


python/eos/plot/plotter_TEST.py
```diff
@@ -0,0 +1,65 @@
...
+        "Handle 'variable' key not given"
+
+        self.plot_args['contents'][0].pop('variable')
+        with self.assertRaisesRegex(ValueError,
+            "Missing key for plot of observable 'B->Dlnu::dBR/dq2;l=e,q=d': 'variable'"):
+            eos.plot.Plotter(self.plot_args).plot()
+
+
+    def test_variable_key_3(self):
+        "Handle value of 'variable' not a kinematic var. nor a paramter"
```
pluegh, line 54: param*e*ter

python/eos/plot/plotter_TEST.py
```diff
@@ -0,0 +1,65 @@
...
+        self.plot_args['contents'][0]['variable'] = 'neitherKinNorPar'
+
+        with self.assertRaisesRegex(ValueError,
+            "Value of 'variable' for observable 'B->Dlnu::dBR/dq2;l=e,q=d' is neither " +
+            "a valid kinematic variable or parameter: 'neitherKinNorPar'"):
+            eos.plot.Plotter(self.plot_args).plot()
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=5)
```
pluegh, line 65: newline

python/eos/parameter_TEST.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def test_assert_valid_name_1(self):
+        "Value of 'name' is invalid (for small-scale physics)"
+
+        with self.assertRaisesRegex(ValueError, "Parameter name 'mass::your_mum' is not known"):
+            eos.Parameters.assert_valid_name('mass::your_mum')
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=5)
```
pluegh, line 20: newline

python/eos/observable_TEST.py
```diff
@@ -0,0 +1,40 @@
...
+        "name and kinamtic name are valid"
+        valid_name = 'B->Dlnu::dBR/dq2;l=e,q=d'
+        invalid_kinematic = 'qToTheTwo'
+        with self.assertRaisesRegex(ValueError,
+            "Kinematic variable 'qToTheTwo' does not match known ones: \['q2'\]"):
+            eos.Observables.assert_valid_name_and_kinematic(valid_name, invalid_kinematic)
+
+
+if __name__ == '__main__':
+    unittest.main(verbosity=5)
```
pluegh, line 40: newline

pluegh (commented):
Typos

python/eos/parameter_TEST.py
```diff
@@ -0,0 +1,20 @@
+import unittest
+import eos
+
+class StaticMethodTests(unittest.TestCase):
+
+    def test_assert_valid_name_0(self):
+        "Value of 'name' is valid"
```
pluegh, line 7: Different notation to `observable_TEST.py`. What's better?

python/eos/plot/plotter_TEST.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+    def test_variable_key_3(self):
+        "Handle value of 'variable' not a kinematic var. nor a paramter"
+
+        self.plot_args['contents'][0]['variable'] = 'neitherKinNorPar'
+
+        with self.assertRaisesRegex(ValueError,
+            "Value of 'variable' for observable 'B->Dlnu::dBR/dq2;l=e,q=d' is neither " +
+            "a valid kinematic variable or parameter: 'neitherKinNorPar'"):
```
pluegh, line 60: or → nor. Test will not pass :(

python/eos/observable_TEST.py
```diff
@@ -0,0 +1,40 @@
+import unittest
+import eos
+
+class StaticMethodTests(unittest.TestCase):
+
+    def test_assert_valid_name_0(self):
+
+        "name is valid"
```
dvandyk, line 7: Is this supposed to be a comment?

python/eos/parameter.py
```diff
@@ -128,3 +128,10 @@ def FromWCxf(wc):
         return parameters
 
 
+    @staticmethod
+    def assert_valid_name(name):
+        "Raise error if 'name' is not a valid EOS paramter name"
+
+        valid_names = [p.name() for p in Parameters()]
```
dvandyk, line 135: That seems dirty. Why don't you add a ``Parameters::has(const std::string & key)`` method and export it?

python/eos/parameter_TEST.py
dvandyk: You first have to tell me what these pseudo comments mean.

python/eos/plot/plotter.py
```diff
@@ -283,8 +283,27 @@ def __init__(self, plotter, item):
...
             eos.info('   plotting EOS observable "{}"'.format(oname))
 
+            # handle plot variable, create kinematic or parameter
+            if not 'variable' in item:
+                raise ValueError("Missing key for plot of observable '" + oname + "': 'variable'")
+            try:
+                eos.Parameters.assert_valid_name(item['variable'])
+                var = parameters.declare(item['variable'], np.nan)
+            except ValueError:
+                eos.Observables.assert_valid_name_and_kinematic(oname, item['variable'])
```
dvandyk, line 296: Why do you assert again the name of the observable? You did so in line 286 above.
All you want is the observable entry, which you can obtain in line 286 already with the proposed change for the first commit (return the observable entry for a valid name).

python/eos/plot/plotter.py
```diff
@@ -283,8 +283,27 @@ def __init__(self, plotter, item):
...
+                var = parameters.declare(item['variable'], np.nan)
+            finally:
+                raise ValueError("Value of 'variable' for observable '" + oname +
+                    "' is neither a valid kinematic variable nor parameter: '" + item['variable'] + "'")
+
+            # create kinematics
+            if 'kinematics' in item:
+                for k, v in item['kinematics'].items():
```
dvandyk, line 304: You should probably check if the variable was a kinematic variable, and avoid overriding it here.This should probably be an error.

python/eos/plot/plotter_TEST.py
```diff
@@ -0,0 +1,65 @@
...
+                        'range': [0.02, 11.63],
+                    },
+                ]
+            }
+
+
+    def test_variabe_key_0(self):
+        "Check valid case: variable is kinemtic"
+
+        eos.plot.Plotter(self.plot_args)
```
dvandyk, line 34: Does this work on the CI, where there is no X server? Would probably be a good idea to write it to ``/dev/null`` or similar.

python/eos/observable_TEST.py
pluegh: Well, a description of the test instead of a long function name. The strings show up in the (verbose) test report.

python/eos/observable.py
```diff
@@ -78,3 +78,22 @@ def _repr_html_(self):
...
+    def assert_valid_name(name):
+        obs_names = [ob[0] for ob in Observables()]
+        if name in obs_names:
+            pass
+        else:
+            raise ValueError("Observable with name '" + name + "' is not known")
+
+    @staticmethod
+    def assert_valid_name_and_kinematic(name, kinematic):
+        Observables.assert_valid_name(name)
```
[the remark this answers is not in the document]
pluegh: I would expect an assertion method to pass quietly if it is positive and raise an exception otherwise, but not to return a value.

Since a kinematic is tied to an observable and the operation is cheap and not used often, I simply assert again. It's the easiest code possible and perceptible performance does not take a dent.

python/eos/observable_TEST.py
dvandyk: Understood, thanks!

python/eos/plot/plotter_TEST.py
pluegh: `make check` gives 

```
plotter.py:1491: UserWarning: Matplotlib is currently using pgf, which is a non-GUI backend, so canno│   48         self.plot_args['contents'][0].pop('variable')
t show the figure.                                                                                                                     │   49         with self.assertRaisesRegex(ValueError,
  plt.show()
```

but passes.

python/eos/observable.py
```diff
@@ -78,3 +78,22 @@ def _repr_html_(self):
         result += r'</table>'
 
         return(result)
+
+    @staticmethod
+    def assert_valid_name(name):
+        obs_names = [ob[0] for ob in Observables()]
```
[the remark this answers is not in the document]
pluegh: Thanks!

python/eos/plot/plotter.py
pluegh: Thanks, changed. Also, asserting the kinematic argument is not a responsibility of the `Observables` class, conceptually.

python/eos/plot/plotter.py
pluegh: Implemented.

pluegh:
Pending:
- [ ] Updating the example notebooks
- [ ] Extension of the CPP side and API to check for valid parameter names

python/eos/parameter.py
pluegh: Done!

python/eos/plot/plotter_TEST.py
```diff
@@ -0,0 +1,101 @@
...
+
+        self.plot_args['contents'][0]['kinematics'] = {'q3': 0.1}
+
+        with self.assertRaisesRegex(ValueError,
+            "Kinematic quantity 'q3' does not match known ones for observable " +
+            "'B->Dlnu::dBR/dq2;l=e,q=d': \['q2'\]"):
+            eos.plot.Plotter(self.plot_args).plot()
+
+
+if __name__ == '__main__':
```
dvandyk, line 99: We need to disable these unit tests on machines that have no XeLaTeX. One possibility: Try a simple call to matplotlib and check if a runtime exception is raised at this point. Otherwise, skip this file entirely.

PR #348, opened 2021-07-20
Deploy development releases for 'eoshep' rather than 'eoshep-latest'

dvandyk:


.github/workflows/manylinux-build+check+deploy.yaml
```diff
@@ -161,8 +164,8 @@ jobs:
                 if [[ ${EOS_IS_TAGGED} == 1 ]] ; then
                     cp ../_build/python/setup.py ./
                 else
-                    sed -e "s/\(^.*version=.*\)',/\1.post${GITHUB_RUN_NUMBER}',/" \
-                        -e "s/\(^.*name=.*\)',/\1-latest',/" \
+                    export DEV_VERSION=$(python3 -c "v = [int(n) for n in '${VERSION}'.split('.')] ; v += [0] * (3 - len(v)) ; v[-1] += 1 ; v = [str(n) for n in v] ; print('.'.join(v))")
+                    sed -e "s/\(^.*version=\).*',/\1${DEV_VERSION}.dev${GITHUB_RUN_NUMBER}',/" \
```
dvandyk, line 168: I think this will fail. There is a missing ``'``

Issue #16, opened 2016-02-23
Get rid of pmclib dependency


fredRos:
Implementing the updates is doable in <=200 lines of code. This would greatly clean up `population_monte_carlo_sampler.cc` and relieve the user from building eos with pmclib.

fredRos:
Minimal approach: just replace calls to the library and into its data structures. Don't touch the other methods inside `population_monte_carlo_sampler.cc`

Issue #43, opened 2018-02-12
Trouble installing libpmc


fredRos:
Following the manual in trying to install from source, I did not succed.
```
mkdir /tmp/libpmc
pushd /tmp/libpmc
wget http://www2.iap.fr/users/kilbinge/CosmoPMC/pmclib_v1.01.tar.gz
tar zxf pmclib_v1.01.tar.gz
pushd pmclib_v1.01
./waf configure --m64 --prefix=$SWMOD_INST_PREFIX

bunzip2: Data integrity error when decompressing.
        Input file = t.bz2, output file = t

It is possible that the compressed file(s) have become corrupted.
You can use the -tvv option to test integrity of such files.

You can use the `bzip2recover' program to attempt to recover
data from undamaged sections of corrupted files.

bunzip2: Deleting output file t, if it exists.
Error: Waf cannot be unpacked, check that bzip2 support is present
```

I tried downloading and unpacking twice and got the same `bzip` error again. So I don't believe in an evil bit flipping error. The hard but clean solution would be to remove the dependency on pmclib and implement the few function that are needed by myself. 

That would make the installation experience easier for everyone

dvandyk:
I cannot reproduce the problem right now. However, I completely agree with the proposed solution.

PR #350, opened 2021-07-21
Purge libpmc, libhdf5, and other obsolete things

dvandyk:
- removes ``PopulationMonteCarloSampler``
 - removes ``MarkovChainSampler``
 - removes auxiliary classes of both
 - removes HDF5 class
 - removes configure options ``--enable-pmc``, ``--with-pmc``, ``--with-hdf5``
 - remove obsolete TravisCI settings
 - exclude more auto-generated files from default settings of VS Code's file view


dvandyk:
This will resolve issues #16, #43.

PR #351, opened 2021-07-28
[utils] Add a preliminary expression parser in eos and its test.

mreboud:
This is still preliminary work since Kinematics are not yet fully implemented.

### DIFF
--- a/eos/observable.cc
+++ b/eos/observable.cc
@@ -25,10 +25,12 @@
 #include <eos/rare-b-decays/observables.hh>
 #include <eos/form-factors/observables.hh>
 #include <eos/meson-mixing/observables.hh>
-//#include <eos/utils/concrete_observable.hh>
+#include <eos/utils/expression-fwd.hh>
+#include <eos/utils/expression-observable.hh>
+#include <eos/utils/expression-parser-impl.hh>
 #include <eos/utils/instantiation_policy-impl.hh>
-#include <eos/utils/private_implementation_pattern-impl.hh>
 #include <eos/utils/observable_stub.hh>
+#include <eos/utils/private_implementation_pattern-impl.hh>
 #include <eos/utils/wrapped_forward_iterator-impl.hh>
 #include <eos/observable-impl.hh>
 
@@ -84,9 +86,12 @@ namespace eos
                 return _entries;
             }
 
-            void insert(const QualifiedName & key, const std::shared_ptr<const ObservableEntry> & value)
+            bool insert(const QualifiedName & key, const std::shared_ptr<const ObservableEntry> & value)
             {
-                _entries[key] = value;
+                auto inserted = _entries.insert(std::pair<QualifiedName, std::shared_ptr<const ObservableEntry>>(key, value));
+
+                // true if the insertion was successfull
+                return inserted.second;
             }
     };
 
@@ -260,8 +265,10 @@ namespace eos
     ObservableEntryPtr
     Observables::operator[] (const QualifiedName & qn) const
     {
-        auto i = _imp->observable_entries.find(qn);
-        if (i != _imp->observable_entries.end())
+        const auto & observable_entries = ObservableEntries::instance()->entries();
+
+        auto i = observable_entries.find(qn);
+        if (i != observable_entries.end())
             return i->second;
 
         return ObservableEntryPtr(nullptr);
@@ -270,13 +277,17 @@ namespace eos
     Observables::ObservableIterator
     Observables::begin() const
     {
-        return ObservableIterator(_imp->observable_entries.begin());
+        const auto & observable_entries = ObservableEntries::instance()->entries();
+
+        return ObservableIterator(observable_entries.begin());
     }
 
     Observables::ObservableIterator
     Observables::end() const
     {
-        return ObservableIterator(_imp->observable_entries.end());
+        const auto & observable_entries = ObservableEntries::instance()->entries();
+
+        return ObservableIterator(observable_entries.end());
     }
 
     Observables::SectionIterator
@@ -292,8 +303,39 @@ namespace eos
     }
 
     void
-    Observables::insert(const QualifiedName & name, const ObservableEntry * entry) const
+    Observables::insert(const QualifiedName & name, const std::string & latex, const Options & forced_options, const std::string & input) const
     {
-        ObservableEntries::instance()->insert(name, std::shared_ptr<const ObservableEntry>(entry));
+        eos::exp::Expression expression;
+
+        using It = std::string::const_iterator;
+        ExpressionParser<It> parser;
+
+        It first(input.begin()), last(input.end());
+        bool completed = qi::phrase_parse(first, last, parser, ascii::space, expression) && (first == last);
+
+        if ((! completed) || expression.empty())
+        {
+            throw ParsingError("Could not parse expression '" + input + "'");
+        }
+
+        ExpressionObservableEntry * expression_observable_entry = new
+            ExpressionObservableEntry(name,
+                                      latex,
+                                      Unit::Undefined(),
+                                      expression,
+                                      forced_options
+                                      );
+
+        if (! expression_observable_entry)
+        {
+            throw InternalError("Could not create expression '" + input + "'");
+        }
+
+        bool insertion_success = ObservableEntries::instance()->insert(name, std::shared_ptr<const ObservableEntry>(expression_observable_entry));
+
+        if (! insertion_success)
+        {
+            throw InternalError("Could not insert observable entry '" + name.str() + "'");
+        }
     }
 }
--- a/eos/utils/expression-fwd.hh
+++ b/eos/utils/expression-fwd.hh
@@ -1,0 +1,33 @@
+/*
+ * Copyright (c) 2021 Méril Reboud
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef EOS_GUARD_EOS_UTILS_EXPRESSION_FWD_HH
+#define EOS_GUARD_EOS_UTILS_EXPRESSION_FWD_HH 1
+
+#include <eos/utils/one-of.hh>
+
+namespace eos::exp
+{
+    class BinaryExpression;
+    class ConstantExpression;
+    class ObservableNameExpression;
+    class ObservableExpression;
+
+    using Expression = OneOf<BinaryExpression, ConstantExpression, ObservableNameExpression, ObservableExpression>;
+}
+
+#endif
--- a/eos/utils/expression-observable.hh
+++ b/eos/utils/expression-observable.hh
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2021 Méril Reboud
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef EOS_GUARD_SRC_UTILS_EXPRESSION_OBSERVABLE_HH
+#define EOS_GUARD_SRC_UTILS_EXPRESSION_OBSERVABLE_HH 1
+
+#include <eos/observable-impl.hh>
+#include <eos/utils/expression-fwd.hh>
+#include <eos/utils/expression-parser-impl.hh>
+#include <eos/utils/wrapped_forward_iterator-impl.hh>
+
+
+namespace eos
+{
+    using eos::exp::Expression;
+
+    class ExpressionObservable :
+        public Observable
+    {
+        private:
+            QualifiedName _name;
+
+            Parameters _parameters;
+
+            Kinematics _kinematics;
+
+            Options _options;
+
+            Expression _expression;
+
+        public:
+            ExpressionObservable(const QualifiedName & name,
+                    const Parameters & parameters,
+                    const Kinematics & kinematics,
+                    const Options & options,
+                    const Expression & expression);
+
+            ~ExpressionObservable()
+            {
+            }
+
+            virtual double evaluate() const;
+            virtual ObservablePtr clone() const;
+            virtual ObservablePtr clone(const Parameters & parameters) const;
+
+            virtual const QualifiedName & name() const
+            {
+                return _name;
+            }
+
+            virtual Parameters parameters()
+            {
+                return _parameters;
+            };
+
+            virtual Kinematics kinematics()
+            {
+                return _kinematics;
+            };
+
+            virtual Options options()
+            {
+                return _options;
+            }
+    };
+
+    class ExpressionObservableEntry :
+        public ObservableEntry
+    {
+        private:
+            QualifiedName _name;
+
+            std::string _latex;
+
+            const Unit & _unit;
+
+            const Expression _expression;
+
+            std::vector<std::string> _kinematics_names;
+
+            Options _forced_options;
+
+        public:
+            ExpressionObservableEntry(const QualifiedName & name, const std::string & latex,
+                    const Unit & unit,
+                    const Expression & expression,
+                    const Options & forced_options);
+
+            ~ExpressionObservableEntry()
+            {
+            }
+
+            virtual ObservableEntry::KinematicVariableIterator begin_kinematic_variables() const;
+            virtual ObservableEntry::KinematicVariableIterator end_kinematic_variables() const;
+            virtual ObservablePtr make(const Parameters & parameters, const Kinematics & kinematics, const Options & options) const;
+            virtual std::ostream & insert(std::ostream & os) const;
+
+            virtual const QualifiedName & name() const
+            {
+                return _name;
+            }
+
+            virtual const std::string & latex() const
+            {
+                return _latex;
+            }
+
+            virtual const Unit & unit() const
+            {
+                return _unit;
+            }
+    };
+}
+
+#endif
--- a/eos/utils/expression-parser-impl.hh
+++ b/eos/utils/expression-parser-impl.hh
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2021 Méril Reboud
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef EXPRESSION_PARSER_IMPL_HH
+#define EXPRESSION_PARSER_IMPL_HH 1
+
+#include <boost/spirit/include/qi.hpp>
+#include <boost/fusion/adapted.hpp>
+
+#include <eos/utils/expression.hh>
+#include <eos/utils/expression-parser.hh>
+
+#include <utility>
+
+namespace eos
+{
+    template <typename Iterator>
+    ExpressionParser<Iterator>::ExpressionParser() :
+        ExpressionParser::base_type(expression)
+    {
+        using namespace boost::spirit::qi;
+
+        expression =
+            additive_expr                                [ _val = _1]
+            ;
+
+        auto make_binary = [] (char op, const auto & lhs, const auto & rhs)
+        {
+            return eos::exp::BinaryExpression(op, lhs, rhs);
+        };
+
+        additive_expr =
+            non_additive_expr                            [ _val = _1]
+            >> *(   (char_("+") >> non_additive_expr)    [ _val = phx::bind(make_binary, _1, _val, _2)]
+                  | (char_("-") >> non_additive_expr)    [ _val = phx::bind(make_binary, _1, _val, _2)]
+                )
+            ;
+
+        non_additive_expr =
+            primary_expr                                 [ _val = _1]
+            >> *(   (char_('*') >> primary_expr)         [ _val = phx::bind(make_binary, _1, _val, _2)]
+                  | (char_('/') >> primary_expr)         [ _val = phx::bind(make_binary, _1, _val, _2)]
+                )
+            ;
+
+        auto make_observable = [] (const std::string & name, const KinematicsSpecification & spec)
+        {
+            return eos::exp::Expression(eos::exp::ObservableNameExpression(name, spec));
+        };
+
+        primary_expr =
+              ( '(' >> expression >> ')' )               [ _val = _1 ]
+            | constant                                   [ _val = _1 ]
+            | (observable_name >> kinematics)            [ _val = phx::bind(make_observable, _1, _2) ]
+            | observable_name                            [ _val = phx::bind(make_observable, _1, KinematicsSpecification()) ]
+            ;
+
+        constant = double_ | int_;
+        observable_name =
+              "<<"
+            >> as_string [ lexeme [ *(char_ - ">>")  ] ] [ _val = _1 ]
+            >> ">>"
+            ;
+
+        kinematics =
+              '['
+            >> (   kinematics_alias                      [ phx::bind(_val, _1) ]
+                 | kinematics_value                      [ phx::bind(_val, _1) ]
+               ) % ','
+            >> ']'
+            ;
+
+        kinematics_alias =
+            (
+            as_string [ lexeme [ *~char_(",=>]") ] ]
+            >> "=>"
+            >> as_string [ lexeme [ *~char_(",=]") ] ]
+            )
+            ;
+
+        kinematics_value =
+            (
+            as_string [ lexeme [ *~char_(",=>]") ] ]
+            >> '='
+            >> double_
+            )
+            ;
+    }
+
+    template <typename Iterator>
+    ExpressionParser<Iterator>::~ExpressionParser()
+    {
+    }
+}
+
+#endif
--- a/eos/utils/expression-parser.hh
+++ b/eos/utils/expression-parser.hh
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2021 Méril Reboud
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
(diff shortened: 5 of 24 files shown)


eos/utils/expression-fwd.hh
```diff
@@ -0,0 +1,33 @@
...
+#include <eos/utils/one-of.hh>
+
+namespace eos::exp
+{
+    class BinaryExpression;
+    class ConstantExpression;
+    class ObservableNameExpression;
+    class ObservableExpression;
+
+    typedef OneOf<BinaryExpression, ConstantExpression, ObservableNameExpression, ObservableExpression> Expression;
```
dvandyk, line 30: Please use ``using Expression = OneOf<...>;``.

eos/utils/expression-maker.cc
```diff
@@ -0,0 +1,72 @@
...
+
+#include <iostream>
+
+namespace eos::exp
+{
+
+    // copies one expression tree to a new tree, replacing ObservableNameExpression objects with ObservableExpression objects
+    class ExpressionMaker
+    {
+        Parameters _parameters;
```
dvandyk, line 34: Please add an explicit ``private:`` and indent the members one level.

eos/utils/expression-maker.cc
```diff
@@ -0,0 +1,72 @@
...
+                    )
+                );
+            }
+
+            Expression visit(ObservableExpression &)
+            {
+                throw InternalError("Encountered ObserableExpression in ExpressionMaker::visit");
+            }
+    };
+}
```
dvandyk, line 72: Missing newline at the end of the file.

eos/utils/expression-parser-impl.hh
```diff
@@ -0,0 +1,71 @@
...
+                )
+            ;
+
+        primary_expr =
+              ( '(' >> expression >> ')' )               [ _val = _1 ]
+            | constant                                   [ _val = _1 ]
+            | observable_name                            [ _val = _1 ]
+            ;
+
+        constant = double_ | int_;
```
dvandyk, line 61: I guess we will not need complex-valued numbers, since this works at the level of observables. Or am I wrong?

eos/utils/expression-parser.hh
```diff
@@ -0,0 +1,70 @@
...
+
+        // Constructor
+        ExpressionParser();
+
+        qi::rule<Iterator, eos::exp::Expression()               , ascii::space_type> expression;
+        qi::rule<Iterator, eos::exp::Expression()               , ascii::space_type> additive_expr;
+        qi::rule<Iterator, eos::exp::Expression()               , ascii::space_type> non_additive_expr;
+
+        qi::rule<Iterator, eos::exp::Expression()               , ascii::space_type> primary_expr;
+        qi::rule<Iterator, eos::exp::ConstantExpression()       , ascii::space_type> constant;
```
dvandyk, line 62: Why do you specialize here to ``eos::exp::ConstantExpression``? I would leave the more general ``eos::exp::Expression``.

eos/utils/expression-parser.hh
```diff
@@ -0,0 +1,70 @@
...
+        qi::rule<Iterator, eos::exp::ObservableNameExpression() , ascii::space_type> observable_name;
```
dvandyk, line 63: Ditto.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+#include <eos/utils/parameters.hh>
+
+#include <iostream>
+
+using namespace test;
+using namespace eos::exp;
+using namespace eos;
+
+
+class ExpressionEvaluator
```
dvandyk, line 36: This should be moved somewhere else.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+            throw InternalError("Encountered ObserableNameExpression in ExpressionEvaluator::visit");
+
+            return 0.0;
+        }
+};
+
+
+struct ExpressionTest
+{
+    const std::string _input;
```
dvandyk, line 67: The leading underscores indicate that this (and all further members) are private. But they are not. Please remove the leading underscores or make them private.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+{
+public:
+    ExpressionParserTest() :
+    TestCase("expression_parser_test")
+    {
+    }
+
+    virtual void run() const
+    {
+        {
```
dvandyk, line 96: Please a a comment before this scope, explaining briefly what's being tested. E.g.,
```C++
// testing basic parsing of constants and binary expressions
```
or some such.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+            ExpressionTest test("1+2*3");
+
+            std::stringstream out;
+            ExpressionPrinter printer(out);
+            test._e.accept(printer);
```
dvandyk, line 101: Here the naming rang my alarm bells, since this looks like you're using a private member variable.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+
+            ExpressionEvaluator evaluator;
+
+            TEST_CHECK(test._completed);
+            TEST_CHECK_EQUAL(test._e.accept_returning<double>(evaluator), 7);
```
dvandyk, line 106: ``7.0``

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,138 @@
...
+            TEST_CHECK_THROWS(InternalError, test2._e.accept_returning<double>(evaluator));
+
+            Parameters p = Parameters::Defaults();
+            Kinematics k = Kinematics({{"q2_min", 0.011164}, {"q2_max", 11.62}});
+            Options o;
+            ExpressionMaker maker(p, k, o);
+            Expression evaluable_test2 = test2._e.accept_returning<Expression>(maker);
+
+            // This test is to be improved with correct kinematics and default parameters
+            TEST_CHECK_RELATIVE_ERROR(evaluable_test2.accept_returning<double>(evaluator), 0.295627626, 1e-5);
```
dvandyk, line 134: Hm, this provide a circular dependence on the EOS libraries, but only when testing: ``libeosutils`` must be built before ``libeosbdecays.so``, but the tests of ``libeosutils`` depend on ``libeosbdecays``. Since it works, let's be happy and leave it as is.

eos/utils/expression.cc
```diff
@@ -0,0 +1,40 @@
...
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#include <eos/utils/expression.hh>
+
+//#include <iostream>
+
+namespace eos::exp
+{
+    double BinaryExpression::Add(double a, double b)      { return a+b; }
```
dvandyk, line 24: Please use lower case for all members functions (even static), except for one very particular case: named constructors.
This is not one, so please convert to lower case.

eos/utils/expression.cc
```diff
@@ -0,0 +1,40 @@
...
+    double BinaryExpression::Subtract(double a, double b) { return a-b; }
+    double BinaryExpression::Multiply(double a, double b) { return a*b; }
+    double BinaryExpression::Divide(double a, double b)   { return a/b; }
+
+    BinaryExpression::func
+    BinaryExpression::Method(char op)
```
dvandyk, line 30: Ditto.

eos/utils/expression.cc
```diff
@@ -0,0 +1,40 @@
...
+#include <eos/utils/expression.hh>
+
+//#include <iostream>
+
+namespace eos::exp
+{
+    double BinaryExpression::Add(double a, double b)      { return a+b; }
+    double BinaryExpression::Subtract(double a, double b) { return a-b; }
+    double BinaryExpression::Multiply(double a, double b) { return a*b; }
+    double BinaryExpression::Divide(double a, double b)   { return a/b; }
```
dvandyk, line 27: Come to think of it: ``add``, ``subtract``, ``multiply`` and ``divide`` are not expressions, they are actions. The expressions are ``sum``, ``difference``, ``product`` and ``ratio``. Care to change that?

eos/utils/expression.cc
```diff
@@ -0,0 +1,40 @@
...
+
+    BinaryExpression::func
+    BinaryExpression::Method(char op)
+    {
+        switch(op) {
+            case '+': return BinaryExpression::Add;
+            case '-': return BinaryExpression::Subtract;
+            case '*': return BinaryExpression::Multiply;
+            case '/': return BinaryExpression::Divide;
+            default:  return nullptr;
```
dvandyk, line 37: This should not happen, should it? I would throw an
```C++
eos::InternalError("Unknown binary operator '" + stringify(op) + "' encountered");
```

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+
+namespace eos::exp
+{
+    class BinaryExpression
+    {
+        public:
+            char _op;
+            Expression _lhs, _rhs;
+
+            typedef double(*func)(double, double);
```
dvandyk, line 37: Please use ``using``:
```C++
using func = (double)(*)(double, double);
```

Why are you not using ``const double &`` for the two arguments?

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+
+            static double Add(double a, double b);
```
dvandyk, line 39: See comment above: ``sum`` vs ``add``, etc.

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+            static double Subtract(double a, double b);
+            static double Multiply(double a, double b);
+            static double Divide(double a, double b);
+
+            static BinaryExpression::func Method(char op);
+
+            BinaryExpression() {}
+
+            BinaryExpression(char op, Expression const& l, Expression const& r) :
```
dvandyk, line 48: EOS coding style puts ``const`` before the reference's type, i.e.: ``const Expression & l``, with all the spaces.

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+                _op(op), _lhs(l), _rhs(r)
+            {
+            }
+    };
+
+    class ConstantExpression
+    {
+        public:
+            double _value;
```
dvandyk, line 57: Pleasr remove the leading underscore.

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+#include <memory>
+#include <iostream>
+#include <map>
+
+namespace eos::exp
+{
+    class BinaryExpression
+    {
+        public:
+            char _op;
```
dvandyk, line 34: Please remove the leading underscores.

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+            {
+            }
+    };
+
+    class ObservableNameExpression
+    {
+        public:
+            std::string _observable_name;
+
+            ObservableNameExpression(const std::string& input = "") :
```
dvandyk, line 70: space between ``string`` and ``&``

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+                 _observable_name(input)
+            {
+            }
+
+    };
+
+    class ObservableExpression
+    {
+        public:
+            ObservablePtr _observable;
```
dvandyk, line 80: Underscore.

eos/utils/expression.hh
```diff
@@ -0,0 +1,89 @@
...
+            ConstantExpression(const double& v = 0) :
+                _value(v)
+            {
+            }
+    };
+
+    class ObservableNameExpression
+    {
+        public:
+            std::string _observable_name;
```
dvandyk, line 68: Underscore.

dvandyk (changes requested):
Thank you Meril, well done! It looks very good! I have a few comments, nothing serious.

eos/utils/expression-parser-impl.hh
mreboud: This is something we can easily add if needed. I would keep the parser as simple as possible until we have a complete working basis if you agree.

eos/utils/expression-parser_TEST.cc
mreboud: In principle, this is parsed as integers, so the result is `(int) 7`. In practice, I am happy with `7.0` because `TEST_CHECK_EQUAL` doesn't test the type :)

mreboud: Ah no sorry, my mistake, the evaluator evaluates to `double`.

dvandyk: Hm, but your visitor can only return a double!

eos/utils/expression-parser-impl.hh
dvandyk: I do! No complex values for now.

eos/utils/expression-parser.hh
mreboud: It simply doesn't work with `eos::exp::Expression`. I think this is needed to construct the correct class directly (with `_val=_1`).

mreboud:
> Thank you Meril, well done! It looks very good! I have a few comments, nothing serious.

Thanks Danny for the review; I accounted for all of your comments but those concerning class specialization in the parser.

eos/utils/expression-parser.hh
dvandyk: Understood. Keep it as is then.

dvandyk:
@mreboud Can you please include my two commits for aliasing kinematic variables (``gh-dvandyk/for-meril``) in this PR?

dvandyk (approved):

eos/utils/kinematic.cc
```diff
@@ -153,39 +174,86 @@ namespace eos
             result += ", " + i->first + '=' + stringify(_imp->variables_data[i->second]);
         }
 
+        // TODO: show aliases
```
dvandyk, line 177: Please remove this. We should not show the aliases at all.

eos/utils/kinematic.cc
```diff
@@ -153,39 +174,86 @@ namespace eos
...
+        const auto j(_imp->alias_map.find(name));
+        bool name_exists  = (_imp->variables_map.end() != i);
+        bool alias_exists = (_imp->alias_map.end() != j);
+
+        if (! name_exists)
+            throw UnknownKinematicVariableError(name);
+
+        if (alias_exists)
+        {
+            // TODO: proper error
```
dvandyk, line 195: Can you please create a ``DuplicateKinematicAliasError`` in the same place where ``UnknownKinematicVariableError`` lives, and use it here? I was lazy.

eos/utils/kinematic.cc
```diff
@@ -200,9 +268,10 @@ namespace eos
         return _imp->variables.end();
     }
 
-    KinematicVariable::KinematicVariable(const std::shared_ptr<Implementation<Kinematics>> & imp, unsigned index) :
+    KinematicVariable::KinematicVariable(const std::shared_ptr<Implementation<Kinematics>> & imp, unsigned index, bool is_alias) :
         _imp(imp),
-        _index(index)
+        _index(index),
+        _is_alias(is_alias)
```
dvandyk, line 274: I don't think ``_is_alias`` is required anymore now. However, I'm happy to leave it in, it might be useful later and does it does not put pressure on memory consumption.

eos/utils/kinematic_TEST.cc
```diff
@@ -48,6 +48,30 @@ class KinematicsTest :
...
+                TEST_CHECK_EQUAL(1.0, kinematics["s_min"]);
+                TEST_CHECK_EQUAL(6.0, kinematics["s_max"]);
+                TEST_CHECK_EQUAL(1.0, kinematics["q2_min"]);
+                TEST_CHECK_EQUAL(6.0, kinematics["q2_max"]);
+
+                TEST_CHECK_EQUAL("s_min", kinematics["s_min"].name());
+                TEST_CHECK_EQUAL("s_min", kinematics["s_min"].name());
+
+                // Test clearing the alias map and access values
+                auto saved_alias = kinematics["s_min"];
```
dvandyk, line 70: ``s_min`` is not an alias, it is **aliased*. Please add a test case that saves, in addition, ``q2_min`` and ``q2_max`` and check that they surive clearing the alias map.

eos/utils/expression-maker.cc
```diff
@@ -0,0 +1,90 @@
...
+                    );
+                // Clear alias map
+                _kinematics.clear_aliases();
+                // Return observable
+                return ObservableExpression(observable);
+            }
+
+            Expression visit(ObservableExpression &)
+            {
+                throw InternalError("Encountered ObserableExpression in ExpressionMaker::visit");
```
dvandyk, line 87: Hm, is this correct? Shouldn't this simply pass through? I don't think we have any need to apply this visitor twice in a row, but it would be perfectly valid, wouldn't it?

eos/utils/expression-parser-impl.hh
```diff
@@ -0,0 +1,111 @@
...
+              '{'
+            >> as_string [ lexeme [ *~char_("}") ] ]     [ _val = _1 ]
+            >> '}'
+            ;
+
+        kinematics =
+            '['
+            >> (
+                kinematics_alias    [ phx::bind(_val, _1) ]
+              | kinematics_value    [ phx::bind(_val, _1) ]
```
dvandyk, line 83: I failed to indent this properly w.r.t. to the other attributes in e.g. line 75.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,142 @@
...
+            std::stringstream out;
+            ExpressionPrinter printer(out);
+            test2.e.accept(printer);
+
+            ExpressionEvaluator evaluator;
+
+            TEST_CHECK(test2.completed);
+            TEST_CHECK_EQUAL_STR(
+                "BinaryExpression(ObservableNameExpression(B->Dlnu::BR;l=tau, aliases=[q2_min=>q2_min_tau])\
+ / ObservableNameExpression(B->Dlnu::BR;l=mu, values=[q2_min=0]))",
```
dvandyk, line 96: I prefer a different multiline C string:
```
TEST_CHECK_EQUAL_STR(
    "BinaryExpression(ObservableNameExpression(B->Dlnu::BR;l=tau, aliases=[q2_min=>q2_min_tau])"
    " / "
    "ObservableNameExpression(B->Dlnu::BR;l=mu, values=[q2_min=0]))",
    out.str()
);
```

eos/utils/expression-printer.cc
```diff
@@ -0,0 +1,84 @@
...
+                        _os << "," << v->first << "=" << v->second;
+                    }
+                    _os << "]";
+                }
+                _os << ")";
+            }
+
+            void visit(ObservableExpression &)
+            {
+                throw InternalError("Encountered ObserableExpression in ExpressionPrinter::visit");
```
dvandyk, line 81: Is this correct? We might want to print an expression that includes observables!

dvandyk (changes requested):
Some tiny changes, but some more important are:
  - questions regarding the treatment of ``ObservableExpression`` in ``ExpressionEvaluator`` and ``ExpressionPrinter``;
  - incorrect test case when clearing a ``Kinematics``'s aliases.

mreboud:
There are still a couple of steps before we can merge:
- [ ]  Fill the `_kinematics_names_array` of the class `ExpressionObservableEntry`: can be done via a visitor method.
- [ ] Implement the `uses(...)` in the class `ExpressionObservable`
- [ ] Check that object, esp. of the type `Expression`, are cloned when needed and shared when possible

eos/utils/expression-cloner.cc
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            Expression visit(const ConstantExpression & e)
+            {
+                return ConstantExpression(e);
+            }
+
+            Expression visit(const ObservableExpression & e)
+            {
+                auto kinematics_values = e.kinematics_specification.values;
```
dvandyk, line 59: I would use ``const auto &`` here.

eos/utils/expression-cloner.cc
```diff
@@ -0,0 +1,90 @@
...
+#include <eos/utils/options.hh>
+#include <eos/utils/parameters.hh>
+#include <eos/utils/qualified-name.hh>
+
+#include <iostream>
+
+namespace eos::exp
+{
+
+    // copies one expression tree to a new tree, replacing ObservableNameExpression objects with ObservableExpression objects
```
dvandyk, line 31: This comment does not reflect what the visitor does.

eos/utils/expression-maker.cc
```diff
@@ -0,0 +1,90 @@
...
+
+                return ObservableExpression(observable, e.kinematics_specification);
+            }
+
+            Expression visit(const ObservableExpression & e)
+            {
+                return e;
+            }
+    };
+}
```
dvandyk, line 90: ``ExpressionCloner`` and ``ExpressionMaker`` seem to be very similar. Can we combine them into one class?

eos/utils/expression-observable.hh
```diff
@@ -0,0 +1,193 @@
...
+                    const Kinematics & kinematics,
+                    const Options & options,
+                    const Expression & expression) :
+                _name(name),
+                _parameters(parameters),
+                _kinematics(kinematics),
+                _options(options),
+                _expression(expression)
+            {
+                // Uses to be implemented
```
dvandyk, line 65: Should we make a ``ParameterUser`` a ``ExpressionVisitor``? That would be an elegant solution to implementing parameter usage.

dvandyk (commented):
A few comments...

eos/utils/expression-maker.cc
mreboud: Good point

mreboud: All well thought, this is maybe not a good idea. The `maker` works on `ObservableNameExpression` while the `cloner` focuses on already made `ObservableExpression`. What I can do is gather the visitors into one `expression-visitors.cc` to avoid having so many files.

eos/utils/expression-parser-impl.hh
```diff
@@ -0,0 +1,110 @@
...
+            | constant                                   [ _val = _1 ]
+            | (observable_name >> kinematics)            [ _val = phx::bind(make_observable, _1, _2) ]
+            | observable_name                            [ _val = phx::bind(make_observable, _1, KinematicsSpecification()) ]
+            ;
+
+        constant = double_ | int_;
+        observable_name =
+              '{'
+            >> as_string [ lexeme [ *~char_("}") ] ]     [ _val = _1 ]
+            >> '}'
```
dvandyk, line 76: Hm, we have a (small) problem here. What about
``b->smumu::Re{c9}``? We should probably:
 - use ``{{`` and ``}}`` as lead in and out for an observable name,
 - prohibit ``{{`` and ``}}`` anywhere inside a ``QualifiedName``.

dvandyk (changes requested):

eos/utils/expression-cloner.cc
```diff
@@ -0,0 +1,88 @@
...
+                    _kinematics.declare(value.first, value.second);
+                }
+                for (const auto & alias : kinematics_aliases)
+                {
+                    _kinematics.alias(alias.first, alias.second);
+                }
+
+                // Make observable
+                auto observable = Observable::make(
+                    e.observable->name().full(),
```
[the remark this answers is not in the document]
mreboud: Maybe I misunderstand `full` vs `str`, but I want to be sure that all options specified in the `QualifiedName` are kept in the process.

eos/utils/expression-parser-impl.hh
dvandyk: As pointed out by Meril on Discord: there are Observables that use ``}}`, hence we have to adjust. Using ``<<`` and ``>>`` now as delimiters.

mreboud: I will add a commit to prohibit `<<` and `>>` in the prefix, the rest is already forbidden.

eos/utils/expression-maker.cc
dvandyk: Understood!

python/_eos.cc
```diff
@@ -298,6 +299,30 @@ BOOST_PYTHON_MODULE(_eos)
         .def("__str__", &Options::as_string)
         ;
 
+    // Units
+    class_<Unit, boost::noncopyable>("Unit", R"(
+            Represents the unit of the observables.
+
+            Seven possible entries are currently implemented in EOS:
+            Undefined, None, GeV, GeV2, InverseGeV2, InverseGeV4, InversePicoSecond.
```
dvandyk, line 307: Can you make this comment a list? It will be rendered as such in the documentation. You can check the documentation locally with
```
make install
make -C doc html
```
The doc will be available as ``doc/_build/html/index.html``.

python/_eos.cc
```diff
@@ -589,6 +614,18 @@ BOOST_PYTHON_MODULE(_eos)
...
         .def("__getitem__", &Observables::operator[])
         .def("__iter__", range(&Observables::begin, &Observables::end))
+        .def("insert", &Observables::insert, R"(
+            Insert a new observable to EOS by parsing the input string.
+
+            :param name: The name of the new observable.
+            :type name: eos.QualifiedName
+            :param latex: The latex representation of the new observable.
+            :type latex: std::string
+            :param options: The set of options relevant to this new observable. These options apply to all the observables in the expression.
```
dvandyk, line 624: Can you emphasize the 'all' here for the documentation, i.e., ``**all**``? Generally, we can use all the markup available in reStructuredText, which is also what we're using for the parameter descriptions.

eos/observable.cc
```diff
@@ -86,7 +88,11 @@ namespace eos
 
             void insert(const QualifiedName & key, const std::shared_ptr<const ObservableEntry> & value)
             {
-                _entries[key] = value;
+                auto inserted = _entries.insert(std::pair<QualifiedName, std::shared_ptr<const ObservableEntry>>(key, value));
+                if (!inserted.second)
```
dvandyk, line 92: Coding style: ``! inserted.second``. I wonder if we should rather return the boolean value, rather than ``void`` here, and let the callers of this ``ObservableEntries::insert()`` deal with potential exceptions.

eos/observable.cc
```diff
@@ -292,8 +304,34 @@ namespace eos
     }
 
     void
-    Observables::insert(const QualifiedName & name, const ObservableEntry * entry) const
+    Observables::insert(const QualifiedName & name, const std::string & latex, const Options & forced_options, const std::string & input_expression) const
```
dvandyk, line 307: Hm, the naming is a bit weird. ``input_expression`` is not an expression yet. How about ``input`` and ``expression`` in lieu of ``input_expression`` and ``parsed_expression``? Make the code a bit more readable I think.

eos/utils/expression-parser.hh
```diff
@@ -0,0 +1,59 @@
...
+#ifndef EXPRESSION_PARSER_HH
+#define EXPRESSION_PARSER_HH 1
+
+#include <boost/spirit/include/qi.hpp>
+#include <boost/spirit/include/phoenix.hpp>
+#include <boost/fusion/adapted.hpp>
+
+#include <eos/utils/expression.hh>
+
+namespace qi    = boost::spirit::qi;
```
dvandyk, line 27: Ah, this is dangerous I think, and I overlooked it so far. Can you please move these three namespace declarations inside the ``eos`` namespace?

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,200 @@
...
+        }
+};
+
+
+class ExpressionParserTest :
+    public TestCase
+{
+public:
+    ExpressionParserTest() :
+    TestCase("expression_parser_test")
```
dvandyk, line 60: Coding style: initalizations/constructor calls are indented one level (4 spaces) with respect to the class name.

eos/utils/expression-parser_TEST.cc
```diff
@@ -0,0 +1,200 @@
...
+    {
+    }
+
+    virtual void run() const
+    {
+        // testing parser failure
+        {
+            ExpressionTest test("1 /* 2");
+
+            TEST_CHECK(!test.completed);
```
dvandyk, line 70: Coding style: ``! test.completed``

eos/utils/expression-visitors.cc
```diff
@@ -0,0 +1,338 @@
...
+                {
+                    _parameter_user->uses(*observable);
+                }
+
+                return ObservableExpression(observable, e.kinematics_specification);
+            }
+
+            Expression visit(const ObservableExpression & e)
+            {
+                return ObservableExpression(e);
```
dvandyk, line 268: Ahh, bug! The return value will not necessarily be bound to the same Parameters (``_parameters``). I think we need to clone the existing observable, using ``_parameters`` as the argument to clone. Maybe you were right in the first place that this case should return an error...

eos/utils/one-of.hh
```diff
@@ -189,7 +189,11 @@ namespace eos
             OneOf(const OneOf & other) :
                 _value(other._value)
             {
+            }
 
+            bool empty() const
+            {
+                return !_value;
```
dvandyk, line 196: Can you split this into a different commit please?

dvandyk (changes requested):

mreboud:
@dvandyk I rebased on master. Unless you have further comments, I suggest we merge this round and open a new issue for the future improvements, including:
 - Manage units in inserted observable entries
 - Ensure that expression observables are cacheable
 - Allow expression observable to be also added at compile time

dvandyk:
> @dvandyk I rebased on master. Unless you have further comments, I suggest we merge this round and open a new issue for the future improvements, including:
> 
> * Manage units in inserted observable entries
> * Ensure that expression observables are cacheable
> * Allow expression observable to be also added at compile time

Sounds good!

Issue #352, opened 2021-07-29
Command line support in Python


dvandyk:
- [x] #353
 - [x] #354
 - [x] #355

dvandyk:
Done.

PR #356, opened 2021-08-02
PR for issue #352 and its tasks

dvandyk:


dvandyk:
Regardig #355, I still don't know how to consistently provide the parameters. For the time being, ``dict`` is passed as keyword arguments. However, for running ``sample-mcmc``, this would mean that the user needs to provide
```YAML
steps:
  - command: 'sample-mcmc'
  - params:
      pre_N: 500
```
in the analysis file. Note the ``pre_N``, which is different from the name in the CLI script (``-n`` or ``--number-of-prerun-samples``). I would be grateful for suggestions.

dvandyk:
> Regardig #355, I still don't know how to consistently provide the parameters. For the time being, `dict` is passed as keyword arguments. However, for running `sample-mcmc`, this would mean that the user needs to provide
> 
> ```yaml
> steps:
>   - command: 'sample-mcmc'
>   - params:
>       pre_N: 500
> ```
> 
> in the analysis file. Note the `pre_N`, which is different from the name in the CLI script (`-n` or `--number-of-prerun-samples`). I would be grateful for suggestions.

I have now setup two maps that rename the arguments such that the freestanding Python functions understand them. This happens in ``eos.AnalysisFile.run()``.

python/eos/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+import numpy as _np
+import os
+import pypmc
+
+def sample_mcmc(analysis_file, posterior, chain, base_directory='./', pre_N=150, preruns=3, N=1000, stride=5):
+    """
+    Samples from a named posterior PDF using Markov Chain Monte Carlo (MCMC) methods.
+
+    The output file will be stored in EOS_BASE_DIRECTORY/POSTERIOR/mcmc-IDX.
+
```
mreboud, line 29: What does IDX refer to?

python/eos/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+    :type N: int, optional
+    :param stride: The ratio of samples drawn over samples stored. For every S samples, S - 1 will be discarded. Defaults to 5.
+    :type stride: int, optional
+    """
+
+    output_path = os.path.join(base_directory, posterior, 'mcmc-{:04}'.format(int(chain)))
+    _set_log_file(output_path, 'log')
+    _analysis_file = eos.AnalysisFile(analysis_file)
+    analysis = _analysis_file.analysis(posterior)
+    rng = _np.random.mtrand.RandomState(int(chain) + 1701)
```
mreboud, line 52: Should we set the seed as an optional parameter?

python/eos/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+    :type threshold: float > 1.0, optional
+    :param K_g: The number of mixture components per cluster. Default to 1.
+    :type K_g: int >= 1, optional
+    """
+
+    import pathlib
+    output_path = os.path.join(base_directory, posterior, 'clusters')
+    input_paths = [str(p) for p in pathlib.Path(os.path.join(base_directory, posterior)).glob('mcmc-*')]
+    chains    = [eos.data.MarkovChain(path).samples for path in input_paths]
+    means     = [_np.mean(chain, axis=0) for chain in chains]
```
mreboud, line 85: `_np.mean(chains, axis=0)` should give the same result in a more efficient way, I think

python/eos/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+    variances = [_np.var(chain,  axis=0) for chain in chains]
```
mreboud, line 86: ditto

python/eos/tasks.py
```diff
@@ -0,0 +1,196 @@
...
+    :type step_N: int > 0, optional
+    :param steps: The number of adaptation steps, which ared used to adapt the PMC proposal to the posterior. Defaults to 10.
+    :type steps: int > 0, optional
+    :param final_N: The number of samples to be stored in the output file. Defaults to 5000,
+    :type final_N: int > 0, optional
+    """
+
+    _analysis_file = eos.AnalysisFile(analysis_file)
+    analysis = _analysis_file.analysis(posterior)
+    rng = _np.random.mtrand.RandomState(1701)
```
mreboud, line 117: ditto

mreboud (approved):

doc/advanced.rst
```diff
@@ -0,0 +1,94 @@
...
+   :nodescription:
+   :nodefault:
+
+   The ``eos-analysis`` script provides several subcommands that
+
+    * inspect the analysis file;
+    * sample from a posterior density with Monte Carlo methods;
+    * perform auxiliary tasks on intermediate results.
+
+   The output of these commands are stored on disc as directories filled with YAML files
```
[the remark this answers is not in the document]
dvandyk: Done!

python/eos/tasks.py
dvandyk: It's the chain index. Replaced it with ``CHAIN`` instead. Good catch.

python/eos/tasks.py
dvandyk: Not at this point. I want calls to sample-mcmc to be reproducable, hence the chain-index based seed.

python/eos/tasks.py
dvandyk: Hm, isn't it ``_np.mean(chains, axis=1)``? That seems to work for me. Changed.

python/eos/tasks.py
dvandyk: Again, using ``axis=1``

python/eos/tasks.py
dvandyk: Want this to be reproducible. Let's leave it as is, and discuss how to ensure reproducability, e.g., by setting the seed in the analysis.yaml.

python/eos/tasks.py
mreboud: Yes you are right!

python/eos/tasks.py
mreboud: If the argument is optional with a default value, the reproducibility is ensured, no ?

PR #361, opened 2021-09-06
Display unit

pluegh:
Add a representation of an observables unit to the Jupyter output and documentation.

### DIFF
--- a/doc/observables.rst.py
+++ b/doc/observables.rst.py
@@ -44,11 +44,16 @@ for section in observables.sections():
         print('     - Kinematic Variables')
         for qn, entry in group:
             latex = entry.latex()
+            unit_string = entry.unit().latex()
+            if unit_string == '1':
+                unit_string = ''
+            else:
+                unit_string = r'\, \left[ {unit_string} \right]'.format(unit_string=unit_string)
             if 0 == len(latex):
                 continue
 
             entry_name = str(qn)
-            entry_desc = latex_to_rst(latex)
+            entry_desc = latex_to_rst(latex + unit_string)
             entry_kv   = ', '.join([('``' + kv + '``') for kv in entry.kinematic_variables()])
             print('   * - ``{qn}``'.format(qn=entry_name))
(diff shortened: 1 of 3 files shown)


doc/observables.rst.py
```diff
@@ -44,11 +44,16 @@ def latex_to_rst(s):
         print('     - Kinematic Variables')
         for qn, entry in group:
             latex = entry.latex()
+            unit_string = entry.unit().latex()
+            if unit_string == '1':
+                unit_string = ''
+            else:
+                unit_string = r'\, \left[ {unit_string} \right]'.format(unit_string=unit_string)
```
dvandyk, line 51: I think this does not do what you expect it to do. The unit's latex representation does not contain the math delimiters. You have to add them yourself. Also, ```\left[``` only works in a math environment.

python/eos/observable.py
```diff
@@ -62,11 +63,16 @@ def _repr_html_(self):
                 group_result = '    <tr><th style="text-align:left" colspan=2>{group}</th></tr>\n'.format(group=group.name())
                 for qn, entry in group:
                     latex = entry.latex()
+                    unit_string = entry.unit().latex()
+                    if unit_string == '1':
+                        unit_string = ''
+                    else:
+                        unit_string = r"\, \left[ {unit_string} \right]".format(unit_string=unit_string)
```
dvandyk, line 70: See above.

dvandyk: Ahh, here it works, due to the explicit ```$$``` below in line 75.

dvandyk (changes requested):

doc/observables.rst.py
dvandyk: I agree on the result, but not your explanation. The magic happens in line 59, which I overlooked.
This seems fine. Did you look at the documentation it produces? ``make -C doc html`` and look into ``doc/_build/index.html`` I think.

dvandyk (approved):
Approval subject to a check from @pluegh that the documentation renders correctly.

dvandyk:
Looks good.

PR #368, opened 2021-09-14
Issue367

mreboud:


mreboud:
This is really stupid, I tried to be smart by building this PR on top of the previous one, I thought it would avoid a merge conflict... but it did not work, I had to rebase on master although those are *exactly* the same commits...

Issue #374, opened 2021-09-21
A bug (?) concerning uppercase Option Values (and maybe Keys)


DomLeljak:
Running an analysis that uses `b-to-vec-l-nu.cc` I notice that providing an unallowed value for `SwitchOption` defined with a key `'U'` doesn't yield an error message according to the line 115 of `options-impl.hh` (which happens correctly if I provide an unallowed value for options with the key `'l'`or `'q'`). This is fixed by defining the key with a lowercase.

This is weird because the allowed option keys in line 106 of `qualified-name.cc` seem to be both lowercase and uppercase letters (even though it says right above it that only lowercase letters are allowed in a comment). If I provide a symbol not provided in line 106 of `qualified-name.cc` the error is correctly diplayed.

Can someone confirm that providing an unallowed value to the option with the key 'U' doesn't yield an error message, and that this is fixed by using lowercase letters as option keys?

mreboud:
I can reproduce the error for b-to-d-l-nu (see branch issue374). What puzzles me for the moment is that implementing an equivalent incorrect switch option in options_TEST.cc did not rise any issue...

mreboud:
After a discussion with @dvandyk, we understood that there are no issue here a priori :)

You are probably using an observable such as `B->Dlnu`, where the `U` option is forced to a given value (`c` in this case see https://github.com/eos/eos/blob/ce65b8746149739f81f75939d797c8d0e56aa00f/eos/b-decays/observables.cc#L194). So when you change the value of `U`, nothing happens because it is overwritten by the `U` option of the observable. This has nothing to do with upper- vs. lowercase: `l` or `q` are indeed options that you can modify.

@dvandyk is working on a commit to warn the user that an option was overwritten.

If you agree on all of this, can you close the issue?

DomLeljak:
Thank you, all is clear now. The options in the analysis get overwritten by the options provided by the observable.

Closing.

PR #376, opened 2021-09-24
Fix issue #374

dvandyk:
@DomLeljak this commit fixed the problem you encountered, in all likelihood. Let's discuss please.

### DIFF
--- a/eos/utils/concrete-cacheable-observable.hh
+++ b/eos/utils/concrete-cacheable-observable.hh
@@ -23,6 +23,7 @@
 #include <eos/observable-impl.hh>
 #include <eos/utils/apply.hh>
 #include <eos/utils/join.hh>
+#include <eos/utils/log.hh>
 #include <eos/utils/tuple-maker.hh>
 #include <eos/utils/wrapped_forward_iterator-impl.hh>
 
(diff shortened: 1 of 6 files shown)


eos/utils/concrete-cacheable-observable.hh
```diff
@@ -311,6 +312,15 @@ namespace eos
 
             virtual ObservablePtr make(const Parameters & parameters, const Kinematics & kinematics, const Options & options) const
             {
+                for (const auto & fo : _forced_options)
+                {
+                    const auto & key = std::get<0>(fo);
+                    if (options.has(key))
+                    {
+                        Log::instance()->message("[ConcreteObservableEntry.make]", ll_error)
```
mreboud, line 320: Shouldn't this rather be `[ConcreteCacheableObservableEntry.make]`?

eos/utils/expression-observable.cc
```diff
@@ -148,6 +149,16 @@ namespace eos
             throw InternalError("Empty expression encountered in ExpressionObservableEntry::make!");
         }
 
+        for (const auto & fo : _forced_options)
+        {
+            const auto & key = std::get<0>(fo);
+            if (options.has(key))
+            {
+                Log::instance()->message("[ConcreteObservableEntry.make]", ll_error)
```
mreboud, line 157: Again I would rather print `[ExpressionObservableEntry.make]`

eos/utils/concrete-cacheable-observable.hh
dvandyk: Indeed.

eos/utils/expression-observable.cc
dvandyk: Yes

PR #377, opened 2021-09-28
Fix latex strings

SKuerten:
Fixes some latex representations


mreboud (commented):
Thanks for this fix, both were my mistakes...

Issue #264, opened 2021-01-19
Add an option in analysis.optimize to start at a random point of the parameters space


mreboud:

dvandyk:
I think it's easier than that:

 - ``eos/statistics/log-posterior.hh`` and ``eos/statistics/log-posterior.cc`:
   - add forward iterator: ``using PriorIterator = ForwardIterator<PriorIteratorTag, const LogPriorPtr>;``
   - add ``PriorIterator begin_priors() const`` and ``PriorIterator end_priors() const``, which simply ``return _priors.begin();`` and ``return _priors.end();``
 - ``python/_eos.cc``, line 468
```C++
        .def("log_priors", range(&LogPosterior::begin_priors, &Log:posterior::end_priors))
```

Then it's as easy as running
```Python
    point = [p.inverse_cdf(rng.uniform()) for p posterior.priors()]
```

dvandyk:
@mreboud What's the progress on this one?

mreboud:
It is on my non urgent todo list, I will try to take care of it in the week

PR #378, opened 2021-10-04
Issue264

mreboud:
Add a method to start the optimization of an analysis at a random point of the parameter space.
Closes issue #264

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -213,15 +213,21 @@ class Analysis:
         return eos.GoodnessOfFit(self.log_posterior)
 
 
-    def optimize(self, start_point=None, **kwargs):
+    def optimize(self, start_point=None, rng=np.random.mtrand, **kwargs):
         """
         Optimize the log(posterior) and returns a best-fit-point summary.
 
-        :param start_point: Parameter point from which to start the optimization, with the elements in the same order as in eos.Analysis.varied_parameters. If not specified, optimization starts at the current parameter point.
+        :param start_point: Parameter point from which to start the optimization, with the elements in the same order as in eos.Analysis.varied_parameters.
+                            If set to "random", optimization starts at the random point in the space of the priors.
+                            If not specified, optimization starts at the current parameter point.
         :param start_point: iterable, optional
+        :param rng: Optional random number generator
+
         """
         if start_point == None:
             start_point = [float(p) for p in self.varied_parameters]
+        elif start_point == "random":
+            start_point = [p.inverse_cdf(rng.uniform()) for p in self.log_posterior.log_priors()]
 
         default_kwargs = { 'method': 'SLSQP', 'options': { 'ftol': 1.0e-13 } }
         if kwargs is None:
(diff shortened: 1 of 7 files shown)


python/eos/analysis.py
```diff
@@ -217,11 +217,15 @@ def optimize(self, start_point=None, **kwargs):
         """
         Optimize the log(posterior) and returns a best-fit-point summary.
 
-        :param start_point: Parameter point from which to start the optimization, with the elements in the same order as in eos.Analysis.varied_parameters. If not specified, optimization starts at the current parameter point.
+        :param start_point: Parameter point from which to start the optimization, with the elements in the same order as in eos.Analysis.varied_parameters.
+                            If set to "Random", optimization starts at the random point in the space of the priors.
```
dvandyk, line 221: Can we use ``random`` (all lower case)?

python/eos/analysis.py
```diff
@@ -217,11 +217,15 @@ def optimize(self, start_point=None, **kwargs):
...
+                            If not specified, optimization starts at the current parameter point.
         :param start_point: iterable, optional
         """
         if start_point == None:
             start_point = [float(p) for p in self.varied_parameters]
+        elif start_point == "Random":
```
dvandyk, line 227: See above. This should now also be adopted for the ``eos-analysis`` script. I think that will require a little bit of modifcations, such that either a list or a string are accepted as the start point.

python/eos/analysis.py
```diff
@@ -217,11 +217,15 @@ def optimize(self, start_point=None, **kwargs):
         """
```
dvandyk, line 217: I was wondering if we should add a ``rng`` keyword argument here, such that for a random starting point we still have a deterministic behaviour.

dvandyk (commented):
Left a few inline comments.

python/eos/analysis.py
mreboud: Definitely!

python/eos/analysis.py
mreboud: I added a new option, let me know if this looks ok

src/scripts/eos-analysis
```diff
@@ -386,6 +392,20 @@ def cmd_find_mode(args):
...
+        _gof = eos.GoodnessOfFit(analysis.log_posterior)
+        _chi2 = _gof.total_chi_square()
+        if _chi2 < min_chi2:
+            gof = _gof
+            bfp = _bfp
+            min_chi2 = _chi2
+    elif args.seed:
+        seed = args.seed
+        rng = RNG(seed)
+        starting_point = [p.inverse_cdf(rng.uniform(0, 1)) for p in analysis.log_posterior.log_priors()]
```
dvandyk, line 405: Isn't this duplicating code? I would have just called ``analysis.optimize("random", rng=rng)``

dvandyk (commented):
Approved w/ exception of the call to ``eos.Analysis.optimize``, which seems to duplicate code.

src/scripts/eos-analysis
mreboud: You are correct, I don't know what I had in mind...

PR #382, opened 2021-10-08
Generalize B->Vlnu to b->ulnu transitions and add B->rholnu and B->omegalnu

dvandyk:
Changes:
 - [x] fixed failure of ``eos/references_TEST``: the ``BPR:2021A`` entry had a colon in an unquoted string.
 - [x] fixed error when detemining WET sector: the scale's parameter name had a ``"cb"`` prefix, rather than ``U + "b"``.
 - [x] fixed error in processes table: "B_s->K^*" has isospin I=1/2.


dvandyk:
@DomLeljak can you take care of Meril's request?

PR #406, opened 2021-10-27
Observables doc

pluegh:
Update observables.rst generation

 - use jinja2 template to separate logic and representation
 - handle case where latex string is not available
 - handle group titles that require substitution to render as TeX


dvandyk:
@pluegh Tracking these changes now in #410, closing.

PR #423, opened 2021-11-19
[inclusive-b-to-s-gamma.cc] Added c8prime contribution to bremsstrahl…

cBobeth:
…ung (see hep-ph/312260 eq.(29))


dvandyk:
Merged via #425.

PR #429, opened 2021-11-26
Remove Fred's dead email address

fredRos:
My old MPP account is not active anymore. I don't think a mail address is needed. We could use my github user name instead


dvandyk:
Thanks Fred, I have included your commit in the `for-v1.0.1` branch. Closing.

PR #433, opened 2021-12-01
Improve cli

mreboud:


python/eos/analysis.py
```diff
@@ -220,7 +220,7 @@ def optimize(self, start_point=None, rng=np.random.mtrand, **kwargs):
         :param start_point: Parameter point from which to start the optimization, with the elements in the same order as in eos.Analysis.varied_parameters.
                             If set to "random", optimization starts at the random point in the space of the priors.
                             If not specified, optimization starts at the current parameter point.
-        :param start_point: iterable, optional
+        :type start_point: iterable, optional
```
dvandyk, line 223: Thanks, I think that one was on me!

dvandyk (approved):
Thanks, this looks good! Can be merged, once the tests pass.

PR #436, opened 2021-12-01
[python] Allow to set scales for the plot axes

### LINKED ISSUE #435: Allow logscale in plots
I think it would be very useful to be able to set logarithmic scales to the plots. To my understanding it is fairly trivial using matplotlib's `scale.LogScale` method.

mreboud:
Set default to 'linear' and rename old 'scale' option to 'scaling_factor'

Closes #435

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -76,6 +76,8 @@ class Plotter:
         mytitle = ''
         myylabel = ''
         myxlabel = ''
+        myyscale = 'linear'
+        myxscale = 'linear'
         if 'title' in myplot:
             mytitle = myplot['title']
 
@@ -108,6 +110,9 @@ class Plotter:
             self.ax.xaxis.set_ticks_position('both')
 
             if 'scale' in myx:
+                myxscale = myx['scale']
+
+            if 'scaling_factor' in myx:
 
                 if 'format' in myx:
                     self.xformat = myx['format']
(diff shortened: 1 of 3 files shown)


python/eos/plot/plotter.py
```diff
@@ -108,14 +110,17 @@ def setup_plot(self):
...
+            if 'scaling_factor' in myx:
 
                 if 'format' in myx:
                     self.xformat = myx['format']
                 else:
                     self.xformat = '${x}$'
                     eos.warn("Argument plot:x:format might be required when using plot:x:scale to avoid side effects")
 
-                self.xscale = float(myx['scale'])
+                self.xscale = float(myx['scaling_factor'])
```
pluegh, line 123: The code would be clearer if this variable was named `x_scaling_factor`.

mreboud: Yes definitely

python/eos/plot/plotter.py
```diff
@@ -76,6 +76,8 @@ def setup_plot(self):
         mytitle = ''
         myylabel = ''
         myxlabel = ''
+        myyscale = 'linear'
```
pluegh, line 79: It would be clearer what are possible and default values for `myXscale` if the default value was not assigned and then overriden as an option, but instead assigned if the argument is not supplied by the user via an `else` below.

mreboud: True

mreboud: On the other hand, it wouldn't work with the current code since labels and scales need to be defined even if the `x` or the `y` argument are not there

pluegh:
I changed the documentation via a fixup of your commit.

PR #437, opened 2021-12-01
Use a common parser as parent parser in eos-analysis

### LINKED ISSUE #434: `argparse` bug
There is an issue with the python parser in the cli. It appears that the main parser consumes the `-v` and `-f` arguments without saving the actual arguments. This yields a fail in the test cases, where the default analysis file is not found.

The fix can probably be implemented form this https://docs.python.org/3/library/argparse.html#parents

dvandyk:
Github: resolves #434


mreboud (approved):
Looks good, thanks!

PR #439, opened 2021-12-02
[github] Add concurrency in the jobs

mreboud:
When a PR is forced push, the previous workflow is automatically cancelled. This will avoid having queued jobs due to anyway useless failing jobs.

### DIFF
--- a/.github/workflows/macos-build+check.yaml
+++ b/.github/workflows/macos-build+check.yaml
@@ -6,6 +6,10 @@ on:
 
 name: Build/Check on MacOS
 
+concurrency:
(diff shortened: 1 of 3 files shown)


dvandyk (approved):
Was first suspicious due to the spaces in the group names, but according to the documentation:
```
A concurrency group can be any string or expression. The expression can only use the github context.
```
Accordingly, this should work. Thanks Meril!

mreboud:
It seems to work since it did cancel the runs when I forget the `MacOS`, `manylinux` and `Ubuntu` specifications.

PR #440, opened 2021-12-02
Changes for Meril's presentation in the Bonn group

dvandyk:


eos/constraint.cc
```diff
@@ -198,7 +198,7 @@ namespace eos
             {
                 if (options.has(key) && (value != options[key]))
                 {
-                    Log::instance()->message("[GaussianConstraintEntry.make]", ll_warning)
+                    Log::instance()->message("[GaussianConstraintEntry.make]", ll_debug)
```
mreboud, line 201: Should I open an issue on the correct way to do this?

mreboud (approved):

PR #441, opened 2021-12-04
Szego polynomial, HTML representation for `eos.AnalaysisFile`, and `eos.version()` -> `eos.__version__`.

dvandyk:


eos/utils/szego-polynomial.hh
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			SzegoPolynomial(const SzegoPolynomial &) = default;
+			SzegoPolynomial(SzegoPolynomial &&) = default;
+			~SzegoPolynomial() = default;
+			SzegoPolynomial & operator= (SzegoPolynomial &&) = default;
+			SzegoPolynomial & operator= (const SzegoPolynomial &) = default;
+
+			// Evaluate the normalized polynomials at z on the real axis, in the interval [-1, +1].
+			// Note that, contrary to the literature [S:2005A], we use an integral measure dmu, which
```
mreboud, line 68: Since you mention this reference can you add it to the references.yaml?

mreboud (approved):
Thanks Danny! Once the tests pass, I approve :)

python/eos/analysis_file.py
```diff
@@ -220,3 +220,31 @@ def predictions(self):
...
+            result += fr'''
+                <tr>
+                    <td>{p}</td>
+                </tr>
+            '''
+
+        result += r'''
+            </tbody>
+        </table>
+        '''
```
mreboud, line 248: This only prints the name of the posteriors, right? What about the rest of the analysis file?

python/_eos.cc
```diff
@@ -809,5 +803,5 @@ BOOST_PYTHON_MODULE(_eos)
     // }}}
 
     // EOS version
-    def("version", impl::version);
+    scope().attr("__version__") = PACKAGE_VERSION;
```
mreboud, line 806: This seems to trigger an error `module 'eos' has no attribute 'version'` for me

eos/utils/szego-polynomial.hh
dvandyk: I did not find the book anywhere publicly. So I changed there reference to an arXiv paper, which is now added as `[S:2004B]`.

python/_eos.cc
dvandyk: Fixed. The documentation generators and the `eos.data` module use `eos.version()`.

python/eos/analysis_file.py
dvandyk: For the time being, I'm happy with listing the available posteriors. I do not want to output each analysis, and right now we do not support creating "only" a log likelihood or log prior from an analysis file; only log posteriors.

eos/references.yaml
```diff
@@ -1278,6 +1278,12 @@ S:2004A:
     id: oai:arXiv.org:hep-ph/0403185
   inspire-id: Seidel:2004jh
   title: Analytic two loop virtual corrections to $b \to d \ell^+ \ell^-$
+S:2004B:
+  authors: Simon, Barry
+  eprint:
+    archive: arXiv
+    id: oai:arXiv.org:math/0405111
+  title: Orthogonal polynomials on the unit circle: New results
```
mreboud, line 1286: The `:` is the source of the failure it seems

dvandyk: Fixed!

PR #444, opened 2021-12-07
Move part of `eos/utils/` into `eos/maths/` and `eos/models/` libraries

dvandyk:


eos/b-decays/lambdab-to-lambdac-l-nu.cc
```diff
@@ -21,17 +21,17 @@
 #include <eos/form-factors/baryonic.hh>
 #include <eos/rare-b-decays/charm-loops.hh>
 #include <eos/b-decays/lambdab-to-lambdac-l-nu.hh>
-#include <eos/utils/complex.hh>
+#include <eos/maths/complex.hh>
```
mreboud, line 24: I know it is a bit annoying to do, but I would like to have the header in alphabetic order and classified by sub-directories...

eos/maths/polylog.hh
```diff
@@ -3,7 +3,7 @@
 #ifndef EOS_GUARD_EOS_UTILS_POLYLOG_HH
 #define EOS_GUARD_EOS_UTILS_POLYLOG_HH 1
```
mreboud, line 4: You changed the name of the flag for some files, but not all of them. Is there a reason for that?

eos/b-decays/lambdab-to-lambdac-l-nu.cc
dvandyk: Done.

eos/maths/polylog.hh
dvandyk: I changed the ones I saw. Went through all files and fixed all include guards.

PR #445, opened 2021-12-07
For master

mreboud:


.vscode/settings.json
```diff
@@ -1,24 +1,28 @@
...
+        "configure": true,
+        "config.status": true,
+        "libtool": true,
+        "stamp-h1": true,
+        "autom4te.cache": true,
+        "config/": true,
+    },
+    "C_Cpp.intelliSenseCacheSize": 0,
+
+    // Insert spaces when pressing Tab. This setting is overriden
```
dvandyk, line 25: JSON does not allow comments, does it? Does this work? Is VSCode tollerating it?

.vscode/settings.json
```diff
@@ -1,24 +1,28 @@
 {
-	"files.exclude": {
```
dvandyk, line 2: I would not bother with the indentation, since this file will be modified by VS Code when changing the workspace settings.

eos/utils/szego-polynomial.hh
```diff
@@ -15,90 +15,122 @@
  * Place, Suite 330, Boston, MA  02111-1307  USA
  */
 
+#ifndef EOS_GUARD_SRC_UTILS_SZEGO_POLYNOMIALS_HH
+#define EOS_GUARD_SRC_UTILS_SZEGO_POLYNOMIALS_HH 1
```
dvandyk, line 19: Thanks. I must have accidentally started from an existing source file, rather than a header file.

eos/utils/szego-polynomial.hh
```diff
@@ -15,90 +15,122 @@
...
+                phi_star[0] = 1.0;
+
+                // we use real-valued Verblunsky coefficients only.
+                for (unsigned n = 1 ; n <= order_ ; ++n)
+                {
+                    // cf. [S:2004B], eq. (1.4), p.2
+                    phi[n]      = z * phi[n - 1]  - _verblunsky_coefficients[n - 1] * phi_star[n - 1];
+
+                    // cf. [S:2004B], eqs. (1.4) and (1.5) in combination;
+                    phi_star[n] = phi_star[n - 1] - _verblunsky_coefficients[n - 1] * z * phi[n - 1];
```
dvandyk, line 122: Did you test this? I'm not so sure if this is correct. In particular, calculating `phi_star`.

eos/utils/szego-polynomial_TEST.cc
```diff
@@ -37,45 +37,56 @@ class SzegoPolynomialTest :
...
+                    TEST_CHECK_RELATIVE_ERROR(p3, -1.76188,            1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR(p4, +2.82970,            1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR(p5, -4.54691,            1.0e-5);
+                }
+
+                {
+                    const auto [p0, p1, p2, p3, p4, p5] = p(complex<double>(0.4, 0.916515139));
+
+                    TEST_CHECK_RELATIVE_ERROR(p0, +0.6351351032391984, 1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR_C(p1, complex<double>(-0.35653430,  0.90040363),  1.0e-5);
```
dvandyk, line 84: Has this been checked against your notebook?

.vscode/settings.json
mreboud: It does :) and it works fine for me

eos/utils/szego-polynomial.hh
mreboud: I did! I don't know why MacOs as trouble with that, it passes for the others

eos/utils/szego-polynomial_TEST.cc
```diff
@@ -37,45 +37,56 @@ class SzegoPolynomialTest :
...
+                    TEST_CHECK_RELATIVE_ERROR(p2, +1.09668,            1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR(p3, -1.76188,            1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR(p4, +2.82970,            1.0e-5);
+                    TEST_CHECK_RELATIVE_ERROR(p5, -4.54691,            1.0e-5);
+                }
+
+                {
+                    const auto [p0, p1, p2, p3, p4, p5] = p(complex<double>(0.4, 0.916515139));
+
+                    TEST_CHECK_RELATIVE_ERROR(p0, +0.6351351032391984, 1.0e-5);
```
dvandyk, line 83: Ah, I should have seen this earlier. You need to check the relative error for ``std::real(p0)`` only. macOS complains that ``p0`` cannot be converted to double when passing to ``std::fabs(const double &)``, while the others apparently have an overload for ``std::fabs(const complex<double> &)``. This puzzles me. I think this should fail!

dvandyk:
```
  template<typename _Tp>
    inline _Tp
    /// fabs(__z) [8.1.8].
    //  Effects:  Behaves the same as C99 function cabs, defined
    //            in subclause 7.3.8.1.
    fabs(const std::complex<_Tp>& __z)
    { return std::abs(__z); }
```
macOS is on the wrong side of the standard here. As of TR1 the C++ standard demands that ``fabs`` works for a complex number, to ensure compatibility with C99.

eos/utils/szego-polynomial_TEST.cc
mreboud: Sure :)

eos/utils/szego-polynomial_TEST.cc
mreboud: Ok makes sense, thanks :)

PR #446, opened 2021-12-10
[python] Allow the use of weights in uncertainty plots

mreboud:


python/eos/plot/plotter.py
```diff
@@ -15,6 +15,7 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59 Temple
 # Place, Suite 330, Boston, MA  02111-1307  USA
 
+from yaml import error
```
mreboud, line 18: Hmm :/ I don't know where this is coming from...

python/eos/plot/plotter.py
```diff
@@ -63,6 +64,34 @@ def next_color(self):
         self.__next_color = (self.__next_color + 1) % len(self.colors)
         return(result)
 
+    @staticmethod
+    def weighted_quantiles(values, quantiles, sample_weight=None):
```
dvandyk, line 68: Naming is a bit weird. Suggest ``values`` and ``weights``, or ``samples`` and ``sample_weights``. Former seems shorter :smile:

python/eos/plot/plotter.py
```diff
@@ -63,6 +64,34 @@ def next_color(self):
...
+        """ Compute the quantiles of a weighted sample, ignoring NaN values.
+        :param values: (*tuple* of *float*) -- Sample from which the quantiles are computed.
```
dvandyk, line 70: ``iterable`` of floats, isn't it?

dvandyk: Or array-like, just as below.

python/eos/plot/plotter.py
```diff
@@ -450,6 +479,7 @@ class Uncertainty(BasePlot):
          * ``samples`` (*list* of tuples of *float*) -- The list of samples of the predictive distribution. Each tuple of samples
            corresponds to an evaluation of the observables at the kinematic configuration corresponding to the ``xvalues``
            entry with the same index.
+         * ``weights`` (*tuple* of *float*, optional) -- The weights of the samples, on a linear scale. Defaults to uniform weights.
```
dvandyk, line 482: ``*list*`` or ``*array-like*`` of floats

python/eos/plot/plotter.py
```diff
@@ -450,6 +479,7 @@ class Uncertainty(BasePlot):
          * ``samples`` (*list* of tuples of *float*) -- The list of samples of the predictive distribution. Each tuple of samples
```
dvandyk, line 479: above: xvalues: ``*array-like* of *float*``

python/eos/plot/plotter.py
```diff
@@ -63,6 +64,34 @@ def next_color(self):
...
 
+    @staticmethod
+    def weighted_quantiles(values, quantiles, sample_weight=None):
+        """ Compute the quantiles of a weighted sample, ignoring NaN values.
+        :param values: (*tuple* of *float*) -- Sample from which the quantiles are computed.
+        :param quantiles: (*tuple* of *float*) -- Quantiles to compute, the values must be in [0, 1]
+        :param sample_weight: array-like of the same length as values
+        :return: numpy.array with computed quantiles.
+        """
+        values = np.array(values)[~np.isnan(values)]
```
dvandyk, line 75: I would introduce
```python
valid_indices = ~np.isnan(values)
values = np.array(values)[valid_indices]
[... later on:]
sample_weight = np.array(sample_weight)[valid_indices]
```

dvandyk: Question: How do you treat isnan in weights?

python/eos/plot/plotter.py
```diff
@@ -63,6 +64,34 @@ def next_color(self):
...
+        if (np.any(sample_weight < 0)):
+            eos.error('The sample weights cannot be negative')
+        if (np.sum(sample_weight) == 0):
+            eos.error('The sum of the sample weights evaluated to zero')
+
+        sorter = np.argsort(values)
+        values = values[sorter]
+        sample_weight = sample_weight[sorter]
+
+        weighted_quantiles = np.cumsum(sample_weight) - 0.5 * sample_weight
```
dvandyk, line 91: Comment would be nice to indicate that each sample's weight is half assigned to the the left and half assigned to the right of the point.

python/eos/plot/plotter.py
```diff
@@ -488,9 +518,12 @@ def __init__(self, plotter, item):
 
             self.xvalues = None
             self.samples = None
+            self.weights = None
             if 'data' in item:
                 self.xvalues = np.array(item['data']['xvalues'])
                 self.samples = item['data']['samples']
+                if 'weights' in item['data'].keys():
```
dvandyk, line 525: I think we dispense with the ``.keys()`` for dictionaries, usually.

python/eos/plot/plotter.py
dvandyk, line 68: Can you make this a private method? ``_weighted_quantiles``? Should we move it to somewhere else? Is it relevant to code beyond ``eos.Plotter``?

dvandyk (commented):
One bug found: mask is incorrect, since you removed the NaNs from values in the first place.

Otherwise looking good!

python/eos/plot/plotter.py
mreboud: I keep it in `eos.Plotter` for the time being. We can always move it later.

PR #448, opened 2021-12-13
[python] Add the pruning option to the cli and correct bug in the computation of test statistics

mreboud:


dvandyk (approved):
Looks good!

PR #451, opened 2021-12-16
Fix for issue #450 

### LINKED ISSUE #450: Charged leptonic B decays
Three brief comments:

1. There should probably either be a separate `.yaml` file for the charged leptonic decay constraints (currently they're in `constraints/Bq-to-ll.yaml`), e.g. `Bq-to-lnu.yaml`, or the existing name of the file should be adapted?
2. The reference for the `B^+->tau^+nu::BR@Belle:2015A` constraint is wrong. Actually, it should be added and changed in the constraint name.
3. The expression for the decay width in `b-decays/b-to-l-nu.cc` will fail for massless leptons, as the `m_l` parameter is in the denominator (line 120).

All of these are by me, so I'll fix them, I just need input on 1, @dvandyk.

DomLeljak:
This should fix #450.

Can you check if it's neat enough @mreboud, please?


DomLeljak:
Ah forgot the makefile :/

mreboud (approved):

Issue #452, opened 2021-12-16
Kinematic Variables in the expression Parser


mreboud:
If an expression observable uses another expression observable, the kinematic variables are not correctly propagated. E.g. 
```
                make_expression_observable("B->D^*lnu::R_D^*", R"(R_{D^*})",
                                Unit::None(),
                                R"(
                                <<B->D^*lnu::BRavg;l=tau>>[q2_max=>q2_tau_max,q2_min=>q2_tau_min]
                                /
                                <<B->D^*lnu::BRavg;l=mu>>[q2_max=>q2_mu_max,q2_min=>q2_mu_min]
                                )"),
```
```
                make_expression_observable("B->D^*lnu::BRavg", R"(\bar{\mathcal{B}}(B\to \bar{D}^*\ell^-\bar\nu))",
                                Unit::None(),
                                R"(
                                0.5 * (
                                       <<B->D^*lnu::BR;cp-conjugate=false>>
                                       +
                                       <<B->D^*lnu::BR;cp-conjugate=true>>
                                       )
                                )"),
```
```
                make_observable("B->D^*lnu::BR", R"(\mathcal{B}(B\to \bar{D}^*\ell^-\bar\nu))",
                                Unit::None(),
                                &BToVectorLeptonNeutrino::integrated_branching_ratio,
                                std::make_tuple("q2_min", "q2_max"),
                                { { "U", "c" }, { "I", "1/2" } }),
```
raises an error for `R_D^*`: `Unknown kinematic variable: 'q2_max'`

mreboud:
@dvandyk I have to reopen this issue for the following reason: If I create an two-step observable like 
```
eos.Observables().insert('B->Kll::NormalizeddBR/ds', R'()',
                         eos.Unit.Unity(),
                         eos.Options(),
                         '( <<B->Kll::dBR/ds>> / <<B->Kpsi::BR>> )')
```
then
```
obs_entry = eos.Observables._get_obs_entry("B->Kll::NormalizeddBR/ds")
[kv for kv in obs_entry.kinematic_variables()]
```
returns `[]` while I expected `['q2']`. Any idea?

mreboud:
@dvandyk, when you have a bit of time :)

Issue #442, opened 2021-12-07
CP (a)symmetries shouldn't use `save.hh`


mreboud:
These observables should be implemented as `ExpressionObservable` or in any other smarter way. These changes are needed in:
  - [ ] `rare-b-decays`
  - [ ] `b-decays`

PR #453, opened 2021-12-17
Issue 442

mreboud:
Solves #452, thanks to @dvandyk 

Solves #442; `BR`, `F_L`, `F_H`, `A_FB` now refer to the CP-averaged quantities. The not averaged one are labeled `*_CP_specific`

Solves #456; `B->K^*gamma` isospin asymmetry is now defined via an expression observable.


dvandyk:
I think we should use ``BR`` as the CP-averaged branching ratio, always, and provide ``A_CP`` in addition. If people want to extract a CP specific rate, they can do so manually.

Ditto for all the other observables. ``S_i`` are always CP-averaged, by definition. ``A_I`` are always CP-asymmetries, by definition. ``F_L``, ``A_FB``, and similar angular observables with well-defined quantum numbers should be CP averages, too, I think.

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
```diff
@@ -226,23 +229,22 @@ class BToKstarGammaBobethCompatibilityTest :
...
-            // This line is commented out since it requires a fix, see Github issue #456
-            // observables.push_back(Observable::make("B->K^*gamma::A_I", p, k, o));
+            observables.push_back(Observable::make("B->K^*gamma::A_I", p, k, o));
 
             std::string filename(EOS_SRCDIR "/eos/rare-b-decays/exclusive-b-to-s-gamma_TEST-btokstargamma.data");
 #ifdef EOS_GENERATE_TEST_DATA
             {
                 std::cout << "-- GENERATING test case data for B->K^*gamma --" << std::endl;
-                RandomNumberGenerator rng;
+                gsl_rng * rng = gsl_rng_alloc(gsl_rng_taus);
```
dvandyk, line 238: ``gsl_rng_taus`` is deprecated in favour of ``gsl_rng_taus2`` according to the GSL manual.

eos/rare-b-decays/b-to-kstar-ll-gp2004_TEST.cc
```diff
@@ -588,15 +591,15 @@ class BToKstarDileptonLowRecoilBobethCompatibilityTest :
 #ifdef EOS_GENERATE_TEST_DATA
             {
                 std::cout << "-- GENERATING test case data for B->K^*ll at low recoil --" << std::endl;
-                RandomNumberGenerator rng;
+                gsl_rng * rng = gsl_rng_alloc(gsl_rng_taus);
```
dvandyk, line 594: Ditto.

eos/rare-b-decays/b-to-k-ll-gp2004_TEST.cc
```diff
@@ -254,15 +260,15 @@ class BToKDileptonLowRecoilBobethCompatibilityTest :
 #ifdef EOS_GENERATE_TEST_DATA
             {
                 std::cout << "-- GENERATING test case data for B->Kll at low recoil --" << std::endl;
-                RandomNumberGenerator rng;
+                gsl_rng * rng = gsl_rng_alloc(gsl_rng_taus);
```
dvandyk, line 263: See below.

dvandyk (commented):
I approve in principle, with only a comment on the use of ``gsl_rng_taus``.

The test case failure is truly puzzling, needs to be understood first.

eos/rare-b-decays/b-to-kstar-gamma-bfs2004_TEST.cc
mreboud: Good point, I missed that

PR #459, opened 2022-01-11
Use of power_of

### LINKED ISSUE #390: Trade pow(..., n) for power_of<n>(...)
@dvandyk Should we define the product `double * eos.Parameter -> double ` to have an implicit casting of `eos.Parameter` to `double` in `power_of<unsigned, eos.Parameter>` and avoid having to put hundreds of parentheses?

mreboud:
Closes #390


mreboud:
I tried to use markdown in the commit names, but it is not a success ><

dvandyk (approved):
Looks good. I'll merge it.

Issue #462, opened 2022-01-17
sample_pmc has a couple of weaknesses


mreboud:
1. The treatment of weights is inconsistent https://github.com/eos/eos/blob/f6615b6017d323fbc2f6b3050b9feca744daffa2/python/eos/analysis.py#L463-L465 because some weights are smaller than `eps = np.finfo(float).eps`. Maybe we should change the definition of `eps` or use the minimal weight as a minimum value.
2. As noted by @dvandyk we are currently updating the PMC using all the generated samples, which is ok. But using their outdated weights which is definitely wrong.
3. The update of the PMC may be run several times on the same sample step, this iterative process may improve the convergence

dvandyk:
From the discussion at today's meeting:

1. Try to get rid of adjusted weights, since ``self._perplexity`` already masks negative weights.
    If this does not work, use ``eps`` even for samples with ``weight = 0``.
    
2. Use ``pypmc.sampler.importance_sampling.combine_weights`` on the samples from the last ``pmc_lookback + 1`` update steps. Default ``pmc_lookback = 0``.

3. Yes, using default of ``pmc_iterrations = 1``.

PR #467, opened 2022-01-18
Issue 462

mreboud:
Fixes issue #462

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -20,6 +20,7 @@ import eos
 import copy as _cp
 import numpy as np
 import scipy
+import pypmc
 
 class BestFitPoint:
     """
@@ -191,7 +192,8 @@ class Analysis:
 
     @staticmethod
     def _perplexity(weights):
-        """Helper function that computes the perplexity of an array of weights"""
+        """Helper function that computes the perplexity of an array of weights.
+           Non positive and NaN weights are neglected in the calculation."""
         # sum positive finite weights only
         weights_sum = np.sum(weights,
             where = np.logical_or(weights > 0, np.isfinite(weights)))
@@ -206,6 +208,10 @@ class Analysis:
             perplexity = np.exp(entropy) / len(normalized_weights)
             return perplexity
 
+    @staticmethod
+    def _ess(weights):
+        """Helper function that computes the effective sample size of an array of weights"""
+        return pypmc.tools.convergence.ess(weights)
 
     def clone(self):
         """Returns an independent instance of eos.Analysis."""
@@ -314,7 +320,6 @@ class Analysis:
         .. note::
            This method requiries the PyPMC python module, which can be installed from PyPI.
         """
-        import pypmc
         try:
             from tqdm.auto import tqdm
             progressbar = tqdm
@@ -424,7 +429,6 @@ class Analysis:
         .. note::
            This method requires the PyPMC python module, which can be installed from PyPI.
         """
-        import pypmc
         try:
             from tqdm.auto import tqdm
             progressbar = tqdm
@@ -457,8 +461,10 @@ class Analysis:
             origins = sampler.run(step_N, trace_sort=True)
             generating_components.append(origins)
             samples = sampler.samples[:]
-            last_perplexity = self._perplexity(np.copy(sampler.weights[-1][:, 0]))
-            eos.info('Perplexity of the last samples after sampling in step {}: {}'.format(step, last_perplexity))
+            last_weights = np.copy(sampler.weights[-1][:, 0])
+            last_perplexity = self._perplexity(last_weights)
+            last_ess = self._ess(last_weights)
+            eos.info(f'Convergence diagnostics of the last samples after sampling in step {step}: perplexity = {last_perplexity}, ESS = {last_ess}')
             if last_perplexity < 0.05:
                 eos.warn("Last step's perplexity is very low. This could possibly be improved by running the markov chains that are used to form the initial PDF for a bit longer")
             weights = sampler.weights[:][:, 0]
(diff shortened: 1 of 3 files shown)


cBobeth (approved):
Could not spot anything.

mreboud:
Let me rebase and add the CLI support for `sample_pmc`.

python/eos/analysis.py
```diff
@@ -463,7 +473,9 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
             adjusted_weights = np.where(
                     np.logical_or(adjusted_weights <= 0, np.isnan(adjusted_weights)),
                     eps, adjusted_weights)
-            eos.info('Perplexity of all previous samples after sampling in step {}: {}'.format(step, self._perplexity(adjusted_weights)))
+            eos.info('Convergence diagnostics of all previous samples after sampling in step {}: perplexity = {}, ESS = {}'.format(
```
dvandyk, line 479: Ditto.

python/eos/analysis.py
```diff
@@ -495,7 +507,8 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
             samples = np.apply_along_axis(self._x_to_par, 1, sampler.samples[:])
             weights = sampler.weights[:][:, 0]
         perplexity = self._perplexity(np.copy(weights))
-        eos.info('Perplexity after final samples: {}'.format(perplexity))
+        ess = self._ess(np.copy(weights))
+        eos.info('Convergence diagnostics after final samples: perplexity = {}, ESS = {}'.format(perplexity, ess))
```
dvandyk, line 527: Ditto.

python/eos/analysis.py
```diff
@@ -476,8 +480,10 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
             eos.info('Convergence diagnostics of all previous samples after sampling in step {}: perplexity = {}, ESS = {}'.format(
                 step, self._perplexity(adjusted_weights), self._ess(adjusted_weights)
                 ))
-            pypmc.mix_adapt.pmc.gaussian_pmc(samples, sampler.proposal, adjusted_weights, mincount=0, rb=True, copy=False)
+            pmc = pypmc.mix_adapt.pmc.PMC(samples, sampler.proposal, adjusted_weights, mincount=0, rb=True)
+            pmc.run(iterations=pmc_iterations, prune=0.0, rel_tol=pmc_rel_tol, abs_tol=pmc_abs_tol)
             # Normalize the weights and remove components with a weight smaller than weight_threshold
+            sampler.proposal = pmc.density
             sampler.proposal.normalize()
             sampler.proposal.prune(threshold = weight_threshold)
```
dvandyk, line 497: This prunes components with weight below ``weight_threshold`` only **after** the PMC updates have been performed. I have no intuition if this is worse than pruning components **during** the update steps. However, a comment in the code would be nice.

python/eos/analysis.py
```diff
@@ -403,6 +403,7 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
         :param pmc_iterations: (advanced) Maximum number of update of the PMC, changing this value may make the update unstable.
         :param pmc_rel_tol: (advanced) Relative tolerance of the PMC. If two consecutive values of the current density log-likelihood are relatively smaller than this value, the convergence is declared.
         :param pmc_abs_tol: (advanced) Absolute tolerance of the PMC. If two consecutive values of the current density log-likelihood are smaller than this value, the convergence is declared.
+        :param pmc_lookback: (advanced) Use the samples and the reevaluated weights of the pmc_lookback previous steps to update the mixture. The default value of 1 disables this feature.
```
dvandyk, line 406: This is hard to understand. Let me try:

```
Use reweighted samples from the previous update steps when adjusting the mixture density.
The parameter determines the number of update steps to "look back".
The default value of 1 disables this feature.
```

python/eos/analysis.py
```diff
@@ -455,14 +456,14 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
         # create PMC sampler
         sampler = pypmc.sampler.importance_sampling.ImportanceSampler(log_target, log_proposal, save_target_values=True, rng=rng)
         generating_components = []
-
-        eps = np.finfo(float).eps
+        proposals = [sampler.proposal]
```
dvandyk, line 459: Hm. You are not copying the proposal here. Can you confirm that subsequent updates do not modify the old proposals?
If yes, please add a comment that this has been explicitly excluded.
If not, please copy the proposal explicitly.

python/eos/analysis.py
```diff
@@ -471,21 +472,30 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
+                 samples.reshape(-1, step_N, len(self.varied_parameters)),
+                 weights.reshape(-1, step_N),
+                 proposals[-pmc_lookback:]
+                )[:][:,0]
+
+            pmc = pypmc.mix_adapt.pmc.PMC(samples, sampler.proposal, reevaluated_weights, mincount=0, rb=True)
             pmc.run(iterations=pmc_iterations, prune=0.0, rel_tol=pmc_rel_tol, abs_tol=pmc_abs_tol)
-            # Normalize the weights and remove components with a weight smaller than weight_threshold
             sampler.proposal = pmc.density
+            proposals.append(sampler.proposal)
```
dvandyk, line 493: See above. Please check if making a copy of the proposal is necessary, and if no, record it in a comment.

dvandyk (changes requested):

python/eos/analysis.py
mreboud: Hmm no clue, this would need a dedicated study... I would say that the current version is the most conservative but it maybe unnecessarily time consuming...

python/eos/analysis.py
```diff
@@ -202,6 +203,11 @@ def _perplexity(weights):
             perplexity = np.exp(entropy) / len(normalized_weights)
             return perplexity
 
+    @staticmethod
+    def _ess(weights):
+        """Helper function that computes the effective sample size of an array of weights"""
+        from pypmc.tools.convergence import ess
```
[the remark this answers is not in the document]
dvandyk: I meant: you could simply use
```python
return pypmc.tools.convergence.ess(weights)
```
which would be the full name of the ``ess`` function. Importing it might take a performance penalty, but I'm not sure.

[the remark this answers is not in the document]
mreboud: This require importing the entire pypmc in the helper function which I wanted to avoid. For the mooment it is imported only in `sample_pmc`

dvandyk (approved):

PR #474, opened 2022-01-31
[python] Fix bug in manual constraints

### LINKED ISSUE #472: Fix `eos.Analysis`'s usage of ``manual_constraint` and add test cases
@cBobeth noticed that I recently broke support for `manual_constraints` in `eos.Analysis`. Needs to be fixed by checking if a constraint is a manual constraint, and if so, deserialize an `eos.ConstraintEntry` for it and use that to create the constraint.

mreboud:
Closes #472

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -131,18 +131,18 @@ class Analysis:
             p.set_max(maxv)
             self.varied_parameters.append(p)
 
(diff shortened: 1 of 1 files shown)


cBobeth:
Looks fine to me, and I tested that the new code does not lead to  the error in my application. Seems to work again.

PR #480, opened 2022-02-01
[python] Improved documentation of kde plotter

cBobeth:


python/eos/plot/plotter.py
```diff
@@ -1254,6 +1254,7 @@ class KernelDensityEstimate1D(BasePlot):
          * ``bandwidth`` (*float*) -- The factor by which the automatically determined kernel bandwidth is scaled. See the SciPy documentation
            for ``gaussian_kde``, ``bw_method='silverman'``. Defaults to 1.
          * ``range`` (*tuple* of two *float*) -- The minimum and maximum value of the x coordinate for which the smooth histogram is plotted.
+         * ``level`` (*float*) -- The probability level of the contour. Defaults to ``68.29``.
```
mreboud, line 1257: What is this `68.29`?

cBobeth: The default value in the code ... a rough value of the percentage for 1-sigma range if it were normally distributed

mreboud: Ah I see it now. Can you change it to 68.27, the effect is completely negligible, but this value is rather weird, isn't it?

mreboud (approved):

PR #482, opened 2022-02-01
[python] Catch linalg error due to components that can't be rescaled

mreboud:
In `sample_pmc`, components with very small weights can sometimes be so close to singular that they can't be rescaled. This commit catches the corresponding error and fix the weight of such components to 0

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -451,12 +451,25 @@ class Analysis:
         log_target = pypmc.tools.indicator.merge_function_with_indicator(self.log_pdf, ind, -np.inf)
 
         # rescale log_proposal arguments to [-1, 1]
-        for component in log_proposal.components:
+        for n, component in enumerate(log_proposal.components):
             rescaled_mu = self._par_to_x(component.mu)
             rescaled_sigma = np.array([[
                 4 * component.sigma[i, j] / (bj[1] - bj[0]) / (bi[1] - bi[0]) for j, bj in enumerate(self.bounds)
                 ] for i, bi in enumerate(self.bounds)])
-            component.update(rescaled_mu, rescaled_sigma)
+            try:
+                component.update(rescaled_mu, rescaled_sigma)
+            except np.linalg.LinAlgError:
+                # The failure is probably due to a non positive definite matrix. We add the problematic smallest eigenvalue
+                # to the diagonal to ensure that all eigenvalues are now strictly positive
+                smallest_eigenvalue = np.abs(np.min(np.linalg.eigvalsh(rescaled_sigma)))
+                rescaled_sigma += (1.0 + np.finfo(float).eps) * np.diag(smallest_eigenvalue * np.ones(len(rescaled_mu)))
+                try:
+                    component.update(rescaled_mu, rescaled_sigma)
+                    eos.warn(f"Components {n} was singular, {-smallest_eigenvalue} is manually added to the diagonal.")
+                except np.linalg.LinAlgError:
+                    log_proposal.weights[n] = 0
+                    eos.warn(f"Components {n} could not be rescaled, the corresponding weight was {log_proposal.weights[n]} "
+                            f"and is now set to zero.")
 
         # create PMC sampler
         sampler = pypmc.sampler.importance_sampling.ImportanceSampler(log_target, log_proposal, save_target_values=True, rng=rng)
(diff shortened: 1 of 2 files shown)


python/eos/analysis.py
```diff
@@ -450,12 +450,16 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
+        for n, component in enumerate(log_proposal.components):
             rescaled_mu = self._par_to_x(component.mu)
             rescaled_sigma = np.array([[
                 4 * component.sigma[i, j] / (bj[1] - bj[0]) / (bi[1] - bi[0]) for j, bj in enumerate(self.bounds)
                 ] for i, bi in enumerate(self.bounds)])
-            component.update(rescaled_mu, rescaled_sigma)
+            try:
+                component.update(rescaled_mu, rescaled_sigma)
+            except np.linalg.LinAlgError:
+                eos.warn(f"Components {n} could not be rescaled, the corresponding weight was {log_proposal.weights[n]} and is now set to zero.")
```
pluegh, line 461: Just a nitpick: the long string would be nicer with a line break. For example:
```
 eos.warn(
    f"Components {n} could not be rescaled, the corresponding weight was {log_proposal.weights[n]} "
    f"and is now set to zero.")
```

pluegh (commented):
Looks technically good to me! But I don't know if it's safe to set the weights to zero.

mreboud:
> Looks technically good to me! But I don't know if it's safe to set the weights to zero.

It is, components with 0 weight will be pruned out after the optimization, and they are not accounted for in the optimization process.

mreboud:
Screenshot from 2022-02-02 08-49-07
I added a commit on local p-value

python/eos/analysis.py
```diff
@@ -473,14 +478,17 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
             last_weights = np.copy(sampler.weights[-1][:, 0])
             last_perplexity = self._perplexity(last_weights)
             last_ess = self._ess(last_weights)
-            eos.info(f'Convergence diagnostics of the last samples after sampling in step {step}: perplexity = {last_perplexity}, ESS = {last_ess}')
+            eos.info(f'Convergence diagnostics of the last samples after sampling in step {step}: '
+                     f'perplexity = {last_perplexity}, ESS = {last_ess}')
             if last_perplexity < 0.05:
-                eos.warn("Last step's perplexity is very low. This could possibly be improved by running the markov chains that are used to form the initial PDF for a bit longer")
+                eos.warn("Last step's perplexity is very low. This could possibly be improved by running "
```
pluegh, line 484: Does that work for regular strings?

mreboud: Yes :)

mreboud:
@dvandyk let me know, when you have time, if you agree with the treatment of faulty components :)

python/eos/analysis.py
```diff
@@ -451,12 +451,24 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
             rescaled_mu = self._par_to_x(component.mu)
             rescaled_sigma = np.array([[
                 4 * component.sigma[i, j] / (bj[1] - bj[0]) / (bi[1] - bi[0]) for j, bj in enumerate(self.bounds)
                 ] for i, bi in enumerate(self.bounds)])
-            component.update(rescaled_mu, rescaled_sigma)
+            try:
+                component.update(rescaled_mu, rescaled_sigma)
+            except np.linalg.LinAlgError:
+                # Try to increase the diagonal by the smallest eigenvalue
+                smallest_eigenvalue = np.min(np.linalg.eigvalsh(rescaled_sigma))
```
dvandyk, line 463: This can be a negative number, correct? I would add ``np.abs(smallest_eigenvalue)``.

I think adding the smallest eigenvalue is a bit too much, but it could work. It's better than aborting the PMC in any case!

python/eos/analysis.py
```diff
@@ -510,12 +525,24 @@ def sample_pmc(self, log_proposal, step_N=1000, steps=10, final_N=5000, rng=np.r
...
             rescaled_mu = self._x_to_par(component.mu)
             rescaled_sigma = np.array([[
                 component.sigma[i, j] * (bj[1] - bj[0]) * (bi[1] - bi[0]) / 4 for j, bj in enumerate(self.bounds)
                 ] for i, bi in enumerate(self.bounds)])
-            component.update(rescaled_mu, rescaled_sigma)
+            try:
+                component.update(rescaled_mu, rescaled_sigma)
+            except np.linalg.LinAlgError:
+                # Try to increase the diagonal by the smallest eigenvalue
+                smallest_eigenvalue = np.min(np.linalg.eigvalsh(rescaled_sigma))
```
dvandyk, line 537: Ditto, missing ``np.abs``.

dvandyk (changes requested):
Looks good, except for the problem with the absolute value I think.

python/eos/analysis.py
mreboud: Ah there is a missing minus sign, thanks. If the matrix is not positive definite the smallest eigenvalue is negative, and it is the minimal number that we can add, no?

dvandyk: Hm, I'm not sure that this statement is true. I assume here that the true eigenvalue is small but positive, and the numerically it accidentally turns negative.

However, I do not see the exact relation between the term added to the diagonal and the shift applied to the smallst eigenvalue! Adding to the diagonal **usually** works, but I usually try something of the order of the sqrt(machine precision).

mreboud: What I mean is that if `A.v = lambda.v` then `(A-lambda.Id).v = 0`. Do we need non-zero eigenvalues?

mreboud: Or do you mean that `np.linalg.eigvalsh(rescaled_sigma)` will return only positive values?

dvandyk: The Cholesky decompostion requires that A is **positive definite**, which is equivalent that all eigenvalues are strictly positive. So yes, we need non-zero eigenvalues!

mreboud: Should I subtract twice the smallest eigenvalue or the eigenvalue times `(1 + 1e-20)`?

PR #490, opened 2022-03-03
Split importance samples output from PMC sampler output

dvandyk:


dvandyk:
Needs to be revisited, since I did not update the plotting classes!

python/eos/data/native.py
```diff
@@ -264,21 +254,19 @@ def evaluate_component(comp, x):
...
         :type parameters: list or iterable of eos.Parameter
-        :param samples: Samples as a 2D array of shape (N, P).
-        :type samples: 2D numpy array
-        :param weights: Weights on a linear scale as a 2D array of shape (N, 1).
-        :type weights: 1D numpy array
-        :param proposal: Mixture density that describes the PMC sampler's proposal.
-        :type proposal: pypmc.density.mixture.MixtureDensity
         :param sigma_test_stat: (optional) If provided, the inverse CDF of -2*log(PDF) will be evaluated, using the provided values as the respective significance.
         :type sigma_test_stat: list or iterable
+        :param samples: Samples as a 2D array of shape (N, P). Needed to gerenate the test statistic.
```
SKuerten, line 266: typo: generate

python/eos/plot/plotter.py
```diff
@@ -1298,7 +1310,31 @@ def __init__(self, plotter, item):
                     self.weights = np.exp(item['data']['log_weights'])
                 else:
                     self.weights = None
+            elif 'data-file' in item:
+                if 'variable' not in item:
+                    raise KeyError('no variable specificed')
 
+                dfname = item['data-file']
+                eos.info('   plotting KDE for "{}"'.format(dfname))
+                prefix = os.path.split(dfname)[-1]
```
SKuerten, line 1319: why -1 and not 1? Is that for safety somehow? Can there be more than 1 head as output of os.path.split?

SKuerten (commented):
I think this is fine :)

python/eos/data/native.py
dvandyk: Fixed, thank you!

python/eos/plot/plotter.py
dvandyk: I had the wrong functionality of ``os.path.split`` in mind, i.e., splitting into a list of arbitrary length at every path separator. Luckily, it worked here. Taking the last element is anyway what I intended.

PR #498, opened 2022-04-21
Improve documentation and autogenerate documentation from example notebooks

### LINKED ISSUE #347: Refactoring the use of examples in the documentation
- Philip thinks we should use raw HTML tables in the doc when representing Jupyter notebook outputs.
   - Stephan mentioned that some Jupyter outputs are not possible with the ``list-table`` RST directive.
   - Philip mentioned that Jupyter supports exporting notebooks to RST.

We should sync the notebook tables to the documentation. Low-priority issue.

mreboud:
Closes #347


mreboud:
Hmm I've added `nbsphinx` to the docker, but the test still cannot find it, @dvandyk what am I missing?

dvandyk:
Did you build the image and push it to the docker hub? I think i need to automatize this...

dvandyk (approved):
Enthusiastically approved!

PR #501, opened 2022-05-11
[python] Handle plot item documentation, validation and implementation safety

pluegh:


python/eos/plot/plotter.py
```diff
@@ -377,77 +377,65 @@ class Observable(BasePlot):
...
-            # create (empty) options
             options = eos.Options()
 
+            # process user input
+            # need oname for some specs
+            try:
+                oname = item['observable']
+            except KeyError:
+                raise ValueError("Missing key: 'observable'")
+            self.specs['variable'].observable_name = oname
```
pluegh, line 404: This is not very elegant, but explicit and safe; forgetting these steps results in an implementation error that will be caught in the notebooks.

dvandyk: Why do you rely on the ``KeyError`` exception?

```python3
if not 'observable' in item:
    raise ValueError(...)
```

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
...
-
-            # create (empty) options
             options = eos.Options()
 
+            # process user input
+            # need oname for some specs
+            try:
+                oname = item['observable']
+            except KeyError:
+                raise ValueError("Missing key: 'observable'")
```
mreboud, line 391: Shouldn't this exception been handled by `ObservableSpec` l.368?

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
...
+            self.specs['variable'].observable_name = oname
+            self.specs['kinematics'].observable_name = oname
+            self.specs['parameters'].observable_name = oname
+
+            self.specs.validate_set(item)
+
+            with self.specs as specs:
+                # process all spec values
+
+                oname = specs['observable'].value
```
mreboud, line 401: Hmm that's weird, no? You just assigned `oname` to all `specs` didn't you?

python/eos/plot/specs.py
```diff
@@ -0,0 +1,546 @@
...
+        # might want to force default value other than None of spec is optional
+
+    @property
+    def value(self):
+        # is treated like an attribute via property dec
+        self.was_used = True
+        if self._value is None:
+            return self.default
+        else:
+            return self._value
```
mreboud, line 32: `return None` is faster, no?

mreboud (commented):
A few questions before we discuss all of this in more details

python/eos/plot/plotter.py
pluegh: Yes, I don't like it either. But we have to obtain the oname for the next few lines. I don't see a different solution.

At least this is safe: if `observable_name` is not set before validation, an `ImplementationError` is raised, which will be caught by the tests.

python/eos/plot/plotter.py
pluegh: The value still has to be accessed. I should add a comment.

python/eos/plot/specs.py
pluegh: I don't understand. It's supposed to either return the default or the user-specified value.

python/eos/plot/plotter.py
dvandyk: Hm, can we pass the ``ObservableSpec`` into those other specs that require the name? As a guaranteed variable?

pluegh: I think a good approach is to pass a reference of `Specs` (let's rename that to `SpecCollection`) to each owned Spec.
Then, they can access the other members and it's very easy to have dependencies. For example, making the Specs/keys 'parameters' and 'parameters-file' aware of each other. All of that without tightly coupling the `Specs` themselves.

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
            }
         """)
 
+        from .specs import Specs, ParametersSpec, KinematicsSpec,\
+        RangeSpec, ObservableSpec, VariableSpec, ParameterFileSpec
+
+        specs_prototype = Specs([
+            ObservableSpec(name='observable'),
+            VariableSpec(name='variable'),
```
dvandyk, line 369: Why is there not ``RangeSpec``, which is mandatory for a plot item of type ``observable``?

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
...
+            # need oname for some specs
+            try:
+                oname = item['observable']
+            except KeyError:
+                raise ValueError("Missing key: 'observable'")
+            self.specs['variable'].observable_name = oname
+            self.specs['kinematics'].observable_name = oname
+            self.specs['parameters'].observable_name = oname
+
+            self.specs.validate_set(item)
```
dvandyk, line 396: Why don't we validate in the ``__init__``? Consider the case that you modify your plot file and make a mistake.
Now you run through half an hour of plotting, before you hit your mistake at the end of the plot file. Not optimal...

This is why we distinguish between the ``__init__`` part and the ``plot`` method.

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
...
+
+            with self.specs as specs:
+                # process all spec values
+
+                oname = specs['observable'].value
+
+                var_name = specs['variable'].value
```
dvandyk, line 403: Can we make this more convenient? Why do we not use
```
def __init__(self, item):
    # ...
    self.variable  = VariableSpec(...)
```
and ensure that ``VariableSpec`` can be converted to a ``str``?

python/eos/plot/plotter.py
```diff
@@ -374,80 +361,75 @@ class Observable(BasePlot):
...
+                if specs['variable'].type == 'kinematic':
+                    var = kinematics.declare(var_name, np.nan)
```
dvandyk, line 405: This should be done as part of the ``__init__``, too. Create the kinematics and the parameters in ``__init__``!

dvandyk (changes requested):
See my comment. I think the current interface and order of validation is a bit inconvenient.

python/eos/plot/plotter.py
pluegh: I think we cannot do without the Specs/SpecCollection. The collection makes it easy to generate the documentation, it's a central place to specify the item requirements and it can be used to implement (weak) dependencies between the single specifications.

python/eos/plot/plotter.py
pluegh: Okay, good point. RIght now, it's just a drop-in replacement for the existing code.

python/eos/plot/plotter.py
pluegh: It's not implemented in the current code, but that's fixed easily.

python/eos/plot/plotter.py
dvandyk: I did not ask to remove the collection, as discussed earlier.

python/eos/plot/plotter.py
```diff
@@ -361,20 +361,22 @@ class Observable(BasePlot):
...
         """)
 
-        from .specs import Specs, ParametersSpec, KinematicsSpec,\
-        RangeSpec, ObservableSpec, VariableSpec, ParameterFileSpec
+        # from .specs import SpecCollection, ParametersSpec, KinematicsSpec,\
+        # RangeSpec, ObservableSpec, VariableSpec, ParameterFileSpec
+        from .specs import SpecCollection, ObservableSpec, VariableSpec
 
-        specs_prototype = Specs([
+        specs_prototype = SpecCollection([
```
dvandyk, line 368: Do we need to mark this as private and thus let the interpreter mangle it? I think mangling is needed here.
(Also, can we get rid of ``_prototype``?)

```python
class BasePlot:
    self.specs = deepcopy(BasePlot._specs)

class Observable:
    self.specs += deepcopy(Observable._specs)

Comment (pluegh) [reply python/eos/plot/plotter.py:368]: Do I understand this correctly that you suggest the following?

```python3
class BasePlot:
    _specs = ... # class member

    __init__(self):
       self.specs = deepcopy(BasePlot._specs) # instance member
```

It is supposed to be accessible from the outside for the documentation, though. I would consider this as api.
If python would allow it, the class member should be const whereas the copies would be mutable. But all we can do is somewhat mark this via `_prototype` or similar.

pluegh:
This PR is obsolete by #671, which majorly refactors the code instead of extending the existing base.

PR #504, opened 2022-05-20
[python] Return fig, ax

pluegh:
Enables further manipulation of the figure after creating a plot.

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -1961,7 +1961,12 @@ class Plotter:
 
 
     def plot(self):
-        """Produces the plot"""
+        """Produces the plot
+
+        :returns:
+            - fig (:py:class:`matplotlib.figure.Figure`) - the created figure
+            - ax (:py:class:`matplotlib.axes.Axes`) - the created axes
+        """
         self.setup_plot()
         self.plot_contents()
 
@@ -1969,6 +1974,8 @@ class Plotter:
 
         if self.output:
             plt.savefig(self.output, bbox_inches='tight', dpi=300)
+
+        return self.fig, self.ax
 
 
 def variable_to_latex(variable):


mreboud:
Despite the simple workaround, I think that this PR is still needed, do you agree @dvandyk @pluegh?

mreboud:
@pluegh Didn't you say you wanted to adapt the examples notebooks?

dvandyk:
> Despite the simple workaround, I think that this PR is still needed, do you agree @dvandyk @pluegh?

Yes please!

python/eos/plot/plotter.py
```diff
@@ -1961,7 +1961,9 @@ def plot_contents(self):
 
 
     def plot(self):
-        """Produces the plot"""
+        """Produces the plot
+
+        returns fig, ax from matplotlib.pyplot.subplots()"""
```
dvandyk, line 1966: Can we please use the Sphinx syntax for documenting return values?

python/eos/plot/plotter.py
```diff
@@ -1970,6 +1972,8 @@ def plot(self):
         if self.output:
             plt.savefig(self.output, bbox_inches='tight', dpi=300)
 
+        return self.fig, self.ax
```
dvandyk, line 1975: OK for now. Will need to revisit if we break API and permit more than one axis per plot.

dvandyk (changes requested):

pluegh:
> Despite the simple workaround, I think that this PR is still needed, do you agree @dvandyk @pluegh?

I don't think it's necessary as it's documented in the user reference and an "advanced" feature that would clutter the notebooks.

python/eos/plot/plotter.py
pluegh: Thanks for the suggestion, I changed it and tested it in sphinx.

python/eos/plot/plotter.py
pluegh: Indeed.

dvandyk (approved):

mreboud:
I still don't see the modification to the notebooks, so this will 'break' the documentation

PR #505, opened 2022-05-21
Split the observable database, correct header, adapt the test of the expression parser

### LINKED ISSUE #452: Kinematic Variables in the expression Parser

mreboud:
Closes #452


eos/utils/expression-visitors.cc
```diff
@@ -385,20 +386,46 @@ namespace eos::exp
...
+
+            kinematic_set.insert(entry->begin_kinematic_variables(), entry->end_kinematic_variables());
+
+            const auto & kinematics_values = e.kinematics_specification.values;
+            const auto & kinematics_aliases = e.kinematics_specification.aliases;
+
+            // Set or alias kinematic specifications
+            for (const auto & value : kinematics_values)
+            {
+                kinematic_set.insert(value.first);
```
mreboud, line 405: Shouldn't we *remove* the kinematic variable from the list since it has been fixed?

eos/utils/expression-visitors.cc
```diff
@@ -385,20 +386,46 @@ namespace eos::exp
...
+            }
+            for (const auto & alias : kinematics_aliases)
+            {
+                kinematic_set.insert(alias.second);
```
mreboud, line 409: Ditto with removing `alias.first`?

eos/utils/expression-visitors.cc
```diff
@@ -385,20 +386,46 @@ namespace eos::exp
...
+            }
+
+            return kinematic_set;
         }
-        for (const auto & alias : kinematics_aliases)
+
+        // otherwise check if 'e' matches the name of a parameter
+        if (kinematic_set.empty())
```
dvandyk, line 416: I think the "is it a parameter" check should be put into an ``else`` block. Only look for this being a parameter if there is no observable entry. This is how ``Observable::make`` treats it.

mreboud: It is already the case since the previous `if` ends with a `return` statement. But I will make it explicit

mreboud:
I don't understand what's happening, I can't reproduce the bug,
```
make -j all
make install
make -j check
```
run w/o problem locally on jammy, even after running `make clean` and removing everything in the build directory...

dvandyk (approved):

PR #510, opened 2022-06-01
Fix the find_mode script and make it a task

### LINKED ISSUE #503: Make find_mode a task

mreboud:
Closes #503

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -23,6 +23,8 @@ import logging
 import numpy as _np
 import os
 import pypmc
+import scipy
+import sys
 
 from .ipython import __ipython__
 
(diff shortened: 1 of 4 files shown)


mreboud:
As @Danny commented, the code doesn't need to be duplicated, we can use the task in lieu of the script.

python/eos/tasks.py
```diff
@@ -79,6 +81,97 @@ def task_wrapper(*args, **kwargs):
...
+    eos.info(f'Starting minimization in {optimizations} points')
+
+    if not start_point is None:
+        starting_point = _np.array(start_point)
+        eos.info('Starting optimization from user-provided point')
+
+        _bfp = analysis.optimize(start_point = starting_point)
+        _gof = eos.GoodnessOfFit(analysis._log_posterior)
+        _chi2 = _gof.total_chi_square()
+    elif from_mcmc >= 0:
```
dvandyk, line 137: If we use a numerical value to id the markov chain, then we should call it the same as in the ``sample_mcmc`` task.
That would be ``chain``.

python/eos/tasks.py
```diff
@@ -79,6 +81,97 @@ def task_wrapper(*args, **kwargs):
...
+        starting_point = [float(p) for p in analysis.varied_parameters]
+        _bfp = analysis.optimize(start_point = starting_point)
+        _gof = eos.GoodnessOfFit(analysis._log_posterior)
+        _chi2 = _gof.total_chi_square()
+        if _chi2 < min_chi2:
+            gof = _gof
+            bfp = _bfp
+            min_chi2 = _chi2
+
+    print('best-fit point:')
```
dvandyk, line 165: Ok, this does not work anymore for a task. Can you convert it to ``info`` log messages?

python/eos/tasks.py
```diff
@@ -79,6 +81,97 @@ def task_wrapper(*args, **kwargs):
...
+    for p, v in zip(analysis.varied_parameters, _bfp.point):
+        print('  - {} -> {}'.format(p.name(), v))
+    print('total chi^2 = {}'.format(min_chi2))
+    print('p value = {:.1f}%'.format(100 * (1.0 - scipy.stats.chi2(gof.total_degrees_of_freedom()).cdf(gof.total_chi_square()))))
+    print('individual test statistics:')
+    for n, e in gof:
+        print('  - {}: chi^2 / dof = {:f} / {}'.format(n, e.chi2, e.dof))
```
dvandyk, line 172: This should now be written into a YAML file in ``{posterior}/modes``

dvandyk (changes requested):

python/eos/tasks.py
mreboud: It does when I tested it, but in any case we need to write the bfp somewhere. I will use info.

python/eos/tasks.py
dvandyk: Also, should this be **appended** to that file? ``modes`` implies more than one mode. If yes, how do we avoid stale modes lingering past their lifetime?

python/eos/tasks.py
mreboud: Yes

python/eos/tasks.py
mreboud: You can indeed have several modes depending on your starting_point. Doing this exercise correctly would require keeping all the putative bfp, comparing them to see if they really are different modes etc... Is that what we want?

dvandyk: For this task, I would equate BFP with mode. Distinguishing would not be needed here. I'm just concerned that several calls to ``find-mode`` with changes to the analysis inbetween could make a mess...

mreboud: If the starting_point is fixed, it shouldn't, right?

dvandyk: I am concerned about incompatible analyses. E.g., changing the likelihood, or changing the dimensionality of the parameter space.

dvandyk: @mreboud and I discussed offline and agreed to label the outputs. Modes will be written into ``POSTERIOR/mode-LABEL``, where the label is user configurable and defaults to ``default``.

dvandyk (approved):

PR #511, opened 2022-06-02
[doc] Use jupyter with venv

pluegh:
* Describe how to use jupyter with venv
* Emphasize keywords to improve readability

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -13,7 +13,7 @@ How Do I ...
 ... use EOS within a virtual Python environment?
 ================================================
 
-We highly recommend to use EOS within a virtual Python environment. To do so, you
+We highly recommend to use EOS within a **virtual Python environment**. To do so, you
 need to install the ``virtualenv`` package, and then create a new virtual environment:
 
 ::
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -22,25 +22,39 @@ need to install the ``virtualenv`` package, and then create a new virtual enviro
     virtualenv --python=python3 ~/eos
 
 
-The new virtual environment will now be activated.
+The new virtual environment will now be **activated**.
+That means, the `python` (and `pip`) command will now execute the local binary instead of a system-wide one.
```
dvandyk, line 26: [... the local binary] contained within the virtual environment, instead of the binaries installed system-wide.

doc/faq.rst
```diff
@@ -22,25 +22,39 @@ need to install the ``virtualenv`` package, and then create a new virtual enviro
...
+    source ~/eos/bin/activate
 
-After activation, you can use install a binary distribution of EOS using ``pip3 install eoshep``.
+
+While activated, you can **install** a binary distribution of EOS using ``pip install eoshep``.
 If you plan to install EOS from source, as discussed in the :ref:`installation section<installation-from-source>`, you must set the prefix
 to match the path to your virtual environment, here: ``~/eos``.
 
-To deactivate the virtual environment, run
+You most likely want to install **jupyter and notebook**.
```
dvandyk, line 38: [...] install the packages ``jupyter`` and ``notebook``.

Let's highlight that these are packages?

doc/faq.rst
```diff
@@ -22,25 +22,39 @@ need to install the ``virtualenv`` package, and then create a new virtual enviro
...
+To be able to access the virtual environment from the notebook interface, you also need to *install it* (details `here <https://ipython.readthedocs.io/en/stable/install/kernel_install.html#kernels-for-different-environments>`_).
```
dvandyk, line 39: [...], you also need to make it known to jupyter (i.e., to *install it*; details `here ...`).

Would be nice explain why and how this install is different than the other. "make it known" helps?

dvandyk (changes requested):
Nice, thank you! Three little suggestions for improvement!

doc/faq.rst
pluegh: I stick to that nomenclature deliberately: we want to keep it as short as possible, the details are available in their doc. And I don't know precisely what's happening either; therefore this is safer.

Could I convince you? We could add quotation marks instead of the cursive to avoid confusion: "install it as a kernel".

doc/faq.rst
pluegh: ```**`package`**``` does not render nicely in sphinx. I prefer to keep them as bold keywords. Do you agree?

doc/faq.rst
pluegh: Okay!

doc/faq.rst
dvandyk: Fair enough, they use that nomenclature in their command. Keep as it is!

doc/faq.rst
dvandyk: Ah, I see. No double formatting, right? Keep it as it is!

dvandyk (approved):

PR #519, opened 2022-06-27
[utils] Add CSplineInterpolation

### LINKED ISSUE #512: Memory leak when using interpolation
We currently have a memory leak in the code: whenever a single point is calculated using interpolation, memory is manually allocated and not freed [here](https://github.com/eos/eos/blob/a7bee0cff7174a753d54b5792092071e924a9185/eos/rare-b-decays/charm-loops.cc#L1366).
The GSL library we use relies on C-style malloc.

An elegant solution would be to handle interpolation using singleton objects for each interpolation function.
That way, the object creator does not have the explicit responsibility of keeping a reference for reuse.

pluegh:
Provides a general-purpose class for cubic spline interpolation.

Can be used to fix #512.


eos/utils/interpolation.hh
```diff
@@ -0,0 +1,50 @@
...
+#include <memory>
+#include <gsl/gsl_spline.h>
+
+namespace eos
+{
+    class CSplineInterpolation
+    {
+        public:
+            CSplineInterpolation() = delete;
+            CSplineInterpolation(const std::vector<double> & data_x, const std::vector<double> & data_y,
```
dvandyk, line 32: Nitpicking: is the ``data`` needed here?

eos/utils/interpolation.hh
```diff
@@ -0,0 +1,50 @@
...
+                std::function<double (const double &)> map_x = [](const double & x){return x;});
+
+            /*!
+             * Evaluate with operator()
+             * Instance can be called like a function, i.e. is a functor
```
dvandyk, line 37: That is not a helpful comment. That is the very meaning of having an ``operator()``. I suggest:
```c++
/*!
 * Evaluate the interpolating function.
 *
 * @param x The point at which the function shall be evaluated.
 */
```

eos/utils/interpolation.hh
```diff
@@ -0,0 +1,50 @@
...
+             */
+            double operator()(const double & x) const;
+
+        private:
```
dvandyk, line 41: I prefer to have all the ``private`` members at the top of the class definition.

eos/utils/interpolation.hh
dvandyk: What is ``map_x`` for?

dvandyk: There is no documentation of the constructors.

eos/utils/interpolation.cc
```diff
@@ -0,0 +1,47 @@
...
+        gsl_interp_init(_interp.get(), &_data_x[0], &_data_y[0], _data_x.size());
+    }
+
+    double CSplineInterpolation::operator()(const double & x) const
+    {
+        double res = 0;
+        int gsl_status = 0;
+
+        gsl_status = gsl_interp_eval_e(_interp.get(), &_data_x[0], &_data_y[0], _map_x(x), _acc.get(), &res);
+        if(gsl_status) {
```
dvandyk, line 41: Coding style:
 - ensure a single space between the ``if`` statement and the condition.
 - blocks open on a line of their own.

eos/utils/interpolation_TEST.cc
```diff
@@ -0,0 +1,67 @@
...
+    public:
+        InterpolationTest() :
+            TestCase("interpolation_test")
+        {
+        }
+
+        virtual void run() const
+        {
+            {
+                // standard initialisation
```
dvandyk, line 37: Please document the entire block *above* the block.

dvandyk: Also, leave a bit of space. Think of C++ blocks as paragraphs in a document.

eos/utils/interpolation_TEST.cc
```diff
@@ -0,0 +1,67 @@
...
+                    {0.0, 1.0, 2.0}
+                };
+                TEST_CHECK_THROWS(GSLError, interp(3.0));
+            }
+            {
+                // map domain of x
+                CSplineInterpolation interp = {
+                    {0.0, 1.0, 2.0},
+                    {0.0, 1.0, 2.0},
+                    [](const double & x) {return 2. * x;}
```
dvandyk, line 62: Ahh, I see. This is a bit overkill. This class does not care how it is used or that you map an e.g. infinite domain to a finite one as in your application. Let's please keep the mapping an implementation detail of the user of ``CSplineInterpolation``.

dvandyk (changes requested):
I have a few comments and change requests.

eos/utils/Makefile.am
```diff
@@ -29,6 +29,7 @@ libeosutils_la_SOURCES = \
 	gsl-hacks.cc \
 	indirect-iterator.hh indirect-iterator-fwd.hh indirect-iterator-impl.hh \
 	instantiation_policy.hh instantiation_policy-impl.hh \
+	interpolation.hh interpolation.cc \
```
dvandyk, line 32: Nitpicking: ``.cc`` comes before ``.hh``.

dvandyk (approved):
Looks good, can be merged!

dvandyk (changes requested):
Please move the files from ``eos/utils`` to ``eos/maths`` and re-title the primary commit. Sorry for the late notice!

eos/maths/interpolation.hh
```diff
@@ -0,0 +1,57 @@
...
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef EOS_GUARD_EOS_UTILS_INTERPOLATION_HH
```
mreboud, line 19: Should be `MATHS` :)

mreboud (commented):
A few additional comments

eos/rare-b-decays/charm-loops.hh
```diff
@@ -22,17 +22,26 @@
 
 #include <eos/maths/complex.hh>
 #include <eos/models/model.hh>
+#include <eos/maths/interpolation.hh>
```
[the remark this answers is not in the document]
pluegh: Hmm. One more push? I cannot change the branch for the PR unfortunately.

eos/rare-b-decays/charm-loops.cc
```diff
@@ -3,6 +3,7 @@
 /*
  * Copyright (c) 2010, 2011, 2014, 2017 Danny van Dyk
  * Copyright (c) 2010 Christoph Bobeth
+ * Copyright (c) 2022 Philip Lüghausen
```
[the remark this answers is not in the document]
mreboud: Ah maybe sorry!

eos/rare-b-decays/charm-loops.hh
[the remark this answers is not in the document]
mreboud: one more push, don't worry about the branch

dvandyk:
> Please move the files from `eos/utils` to `eos/maths` and re-title the primary commit. Sorry for the late notice!

This is resolved!

PR #521, opened 2022-06-27
Provide a systematic framework to run analysis steps in a reproducible way

dvandyk:
(addresses discussion #432)

Each ``step`` requires a name, and a list of task descriptions. Optionally, it can provide a list of ``iterations``.
Similar to Github Action's ``strategy.matrix``, a step is run for each element of ``iterations``. In the example below,
the ``sample`` step is run twice: once for the posterior ``CKM-pi``, and once for the posterior ``CKM-rho``.

The example is correctly parsed and interpreted. It currently does not run the tasks. Instead, it prints the equivalent
``eos-analysis`` commands.

```yaml
steps:
  - name: optimize
    iterations:
      - { posterior: CKM-all                                                         }
    tasks:
      - task: sample-mcmc
        arguments:
          posterior: 'CKM-all'
          pre_N: 1000

      - task: find-mode
        arguments:
          posterior: 'CKM-all'


  - name: sample
    iterations:
      - { posterior: CKM-pi                                                          }
      - { posterior: CKM-rho, 'sample-mcmc': { pre_N: 1500, starting_point: 'mode' } }
    tasks:
      - task: sample-mcmc
        arguments:
          chain: 0-1,2
          pre_N: 1000 # overriden for 'CKM-rho' to use 'pre_N': 1500

      - task: find-clusters
```

### DIFF
--- a/python/eos/analysis_file.py
+++ b/python/eos/analysis_file.py
@@ -155,44 +155,134 @@ class AnalysisFile:
             for k, v in params.items()
         }
 
-    def run(self):
+    def steps(self, base_directory=None):
         """Runs predefined analysis steps recorded in the analysis file."""
-        import inspect
-        command_map = {
-            'find-clusters':       eos.find_clusters,
-            'predict-observables': eos.predict_observables,
-            'sample-mcmc':         eos.sample_mcmc,
-            'sample-pmc':          eos.sample_pmc,
-        }
+        from .tasks import _tasks
+        from collections import ChainMap
+        from inspect import signature
+        from itertools import chain, product
+        import networkx as nx
 
-        for idx, step in enumerate(self._steps):
-            if type(step) is not dict:
-                raise ValueError("Step #{} is not a key/value map.")
+        def _expand_arguments(step_name, task, arguments):
+            task_sig = signature(_tasks[task])
 
-            if 'command' not in step:
-                raise ValueError("Step #{} contains no command.")
-
-            command  = step['command']
-            func     = command_map[command]
-            params   = step['parameters'] if 'parameters' in step else {}
-            params   = { params_map[(command, k)]: v for k, v in params.items() }
-            paramstr = ','.join(['{k}={v}'.format(k=k,v=v) for k, v in params])
-
-            func_sig = inspect.signature(func)
-            func_required_args = {}
-            for n, p in func_sig.parameters.items():
-                if p.default() != p.empty():
+            task_required_args = set()
+            for n, p in task_sig.parameters.items():
+                if p.default != p.empty:
                     continue
-                func_required_args += { n }
-            for n in func_required_args:
-                if n in params.keys():
+                task_required_args.add(n)
+            for n in task_required_args:
+                if n in arguments.keys():
                     continue
-                eos.error('Mandatory argument \'{}\' not provided'.format(n))
-                return
+                raise ValueError(f'Task "{task}" in step "{step_name}" requires mandatory argument \'{n}\', which is not provided')
 
-            eos.info('Beginning step #{i}: {cmd}({params})'.format(i=i,cmd=cmd, params=paramstr))
-            func(**params)
-            eos.info('Step #{i} complete'.format(i=i))
+            outer_list = []
+            for key, value in arguments.items():
+                if key not in task_sig.parameters:
+                    eos.warn(f'Task "{task}" in step "{step_name}" does not expect argument "{key}", possibly misspelled?')
+                    continue
+
+                param_type = task_sig.parameters[key].annotation
+
+                # handle special cases
+                if param_type is int:
+                    # integer range?
+                    if type(value) is str:
+                        values = list(chain.from_iterable(range(int(v.split("-")[0]),int(v.split("-")[-1])+1) for v in value.split(",")))
+                        outer_list.append([{key: v} for v in values])
+                    else:
+                        # append
+                        outer_list.append([{key: value}])
+                # otherwise append w/o modification
+                else:
+                    # append
+                    outer_list.append([{key: value}])
+
+            result = []
+            for r in list(product(*outer_list)):
+                _arguments = {}
+                for d in r:
+                    _arguments.update(d)
+
+                # replace format strings
+                arguments = {}
+                for key, value in _arguments.items():
+                    if type(value) is str:
+                        arguments.update({key: value.format(**_arguments)})
+                    else:
+                        arguments.update({key: value})
+
+                result.append(arguments)
+
+            return result
+
+        def _handle_single_step(step_desc, step_arguments):
+            step_name = step_desc['name']
+            tasks     = step_desc['tasks']
+
+            for task_desc in tasks:
+                task = task_desc['task']
+                if task not in _tasks.keys():
+                    raise ValueError(f'Unknown task \"{task}\" encountered in step {step_name}')
+
+                arguments = { 'analysis_file': self }
+                for key, value in step_arguments.items():
+                    if key in _tasks.keys():
+                        continue
+
+                    arguments[key] = value
+
+                if 'arguments' in task_desc:
+                    arguments.update(task_desc['arguments'])
+
+                if task in step_arguments:
+                    arguments.update(step_arguments[task])
+
+                if base_directory:
+                    arguments.update({ 'base_directory': base_directory })
+
+                return [(step_name, step_desc, task, a) for a in _expand_arguments(step_name, task, arguments)]
+
+        # main part starts here
+
+        # check inputs
+        for step_idx, step_desc in enumerate(self._steps):
+            if type(step_desc) is not dict:
+                raise ValueError(f'Description of step #{step_idx} is not a key/value map')
+
+            if 'name' not in step_desc:
+                raise ValueError(f'Description of step #{step_idx} requires a name')
+
+            if 'tasks' not in step_desc:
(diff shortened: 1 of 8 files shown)


python/eos/tasks.py
```diff
@@ -15,6 +15,8 @@
 # this program; if not, write to the Free Software Foundation, Inc., 59 Temple
 # Place, Suite 330, Boston, MA  02111-1307  USA
 
+from __future__ import annotations
```
mreboud, line 18: Can you add a comment why you need this for old version (so that we remove it at some point)?

python/eos/analysis_file.py
```diff
@@ -157,42 +157,104 @@ def _sanitize_params(command, params):
...
+
+                # straight match?
+                if type(value) is param_type:
+                    # append
+                    outer_list.append([{key: value}])
+                # handle integer parameters
+                elif param_type is int:
+                    # integer range?
+                    if type(value) is str and ('-' in value or ',' in value):
+                        values = list(chain.from_iterable(range(int(v.split("-")[0]),int(v.split("-")[-1])+1) for v in value.split(",")))
```
mreboud, line 195: Was this tested?

mreboud (commented):
Looks very promising!

python/eos/analysis_file.py
dvandyk: Yes, but insufficiently. A full test suit should be added. Still thinking about how to do that. I think I would prefer returning the supposed ``eos-analysis`` commands as a string in a "dry-run" mode.

python/eos/tasks.py
dvandyk: Yes

python/eos/analysis_file.py
mreboud: I definitely like the idea of a "dry-run" mode

python/eos/tasks.py
dvandyk: Removed, due to problems with the annotations. Returning to plain ``list`` hints, rather than ``list[float]`` or similar.

python/eos/analysis_file.py
dvandyk: Dry run and test added.

src/scripts/eos-analysis_TEST
```diff
@@ -9,6 +9,16 @@ export EOS_BASE_DIRECTORY=$(mktemp -d)
...
 echo using PYTHONPATH=${PYTHONPATH} >&2
 
+###########################################
+## Run the Analysis's steps automaticall ##
+###########################################
+echo running analysis steps test ... >&2
+$PYTHON ${SOURCE_DIR}/eos-analysis \
+    run \
+    -f ${SOURCE_DIR}/eos-analysis_TEST.d/analysis.yaml
+echo ... success >&2
```
mreboud, line 19: Should this test replace the other tests?

src/scripts/eos-analysis_TEST.d/analysis.yaml
```diff
@@ -79,3 +79,28 @@ predictions:
...
+          label: 'seed-{seed}'
+
+  - name: sample
+    iterations:
+      - { posterior: CKM+FF }
+    tasks:
+      - task: sample-mcmc
+        arguments:
+          chain: '0,1-2,3'
+          pre_N: 500
```
mreboud, line 100: Can we input the seed here is as well? I want to check that the analysis is indeed reproducible.

src/scripts/eos-analysis
```diff
@@ -397,6 +397,10 @@ Runs a list of subcommands.
         help = 'The base directory for the storage of data files. Can also be set via the EOS_BASE_DIRECTORY environment variable.',
         dest = 'base_directory', action = 'store', default = get_from_env('EOS_BASE_DIRECTORY', './')
     )
+    parser_run.add_argument('-d', '--dry-run',
+        help = 'Whether or not to perform a dry run of the analysis.',
```
mreboud, line 401: Can you give a more user-friendly definition? the notion of `dry-run` may be unknown by the users.

src/scripts/eos-analysis
```diff
@@ -397,6 +397,10 @@ Runs a list of subcommands.
...
+        dest = 'dry_run', action = 'store_true', default = False
```
mreboud, line 402: I wonder if we should invert this argument to force the user to check its analysis before it is ran.

python/eos/tasks.py
```diff
@@ -73,16 +77,17 @@ def task_wrapper(*args, **kwargs):
                     display(iaccordion)
                 # use invocation-specific ipython output widget (if available)
                 with ioutput:
-                    result = func(*args, **kwargs)
+                    result = func(**_args)
```
mreboud, line 80: Why did you have to remove `**kwargs` here?

python/eos/tasks.py
```diff
@@ -259,8 +256,8 @@ def find_clusters(posterior, base_directory='./', threshold=2.0, K_g=1, analysis
     eos.data.MixtureDensity.create(os.path.join(base_directory, posterior, 'clusters'), density)
 
 
-@task('{posterior}/product')
-def mixture_product(posterior, posteriors, base_directory='./', analysis_file=None):
+@task('mixture-product', '{posterior}/product')
+def mixture_product(posterior:str, posteriors:list, base_directory:str='./', analysis_file:str=None):
```
mreboud, line 260: Unrelated to your commit, but why do I kept two arguments, `posterior` and `posteriors`? `posterior` could be the first element of posteriors by convention and one would do:
```
mixture_product( [postA, postB] )
```
instead of 
```
mixture_product( postA, [postB] )
```
?

python/eos/tasks.py
```diff
@@ -422,9 +411,47 @@ def run_steps(analysis_file, base_directory='./'):
     :param analysis_file: The name of the analysis file that describes the named posterior, or an object of class `eos.AnalysisFile`.
     :type analysis_file: str or `eos.AnalysisFile`
```
mreboud, line 412: Can you update the documentation?

python/eos/analysis_file.py
```diff
@@ -155,44 +155,134 @@ def _sanitize_params(command, params):
...
+        # return steps and theirs tasks in the order imposed by the dependency graph
+        result = []
+        for step in steps:
+            step_desc = { sd['name']: sd for sd in self._steps }[step]
+            iterations = [{}]
+            if 'iterations' in step_desc:
+                iterations = step_desc['iterations']
+
+            for iteration_arguments in iterations:
+                result.extend(_handle_single_step(step_desc, iteration_arguments))
```
mreboud, line 283: Where is this used? I couldn't find, I was expecting this to appear in the `Executor`, no?

mreboud (changes requested):
I really like it, I have a couple of comments. This will also need a bit of documentation I think :)

src/scripts/eos-analysis_TEST
dvandyk: I thought about this yesteday. I think it should not, since we are both testing the tasks themselves and the CLI.

src/scripts/eos-analysis_TEST.d/analysis.yaml
dvandyk: The task automatically assigns a seed based on the chain index:

https://github.com/eos/eos/blob/67aa0916a26270d0fe8bb3e08b534160e7bef4bf/python/eos/tasks.py#L217

src/scripts/eos-analysis
dvandyk: Will do!

src/scripts/eos-analysis_TEST.d/analysis.yaml
mreboud: hmm true, should we maybe move all the seed assignment to the arguments of the task?

src/scripts/eos-analysis_TEST
mreboud: I was just afraid of time-consumption, but if we only test on a dry-run, it should be ok :)

dvandyk: That's not such a good idea, since I want to test that this really works. It runs one optimization and four short markov chains, which is really not a big time consumption.

src/scripts/eos-analysis_TEST.d/analysis.yaml
dvandyk: For ``sample-mcmc`` I would be opposed. Here, seeding is achieved through the chain index. Let's not make it a 2D problem.

src/scripts/eos-analysis
dvandyk: I was copying what is done in other programms, e.g., ``make``. The ``--dry-run`` or similar arguments are similar to what I implemented, and not inverted. In the end, we have to trust the users.

python/eos/tasks.py
dvandyk: I did not. Actually, everything is a keyword argument now. I have to modify some arguments by name earlier in the code, and assemble a dictionary of all of the tasks arguments. Nothing is provided using positional arguments anymore.

python/eos/tasks.py
dvandyk: I think your example makes no sense, since that would not be a mixture product. The following example makes more sense:
```python
mixture_product('WET+FF', ['WET', 'FF'], ...)
```

python/eos/tasks.py
dvandyk: Yes, sorry.

mreboud (approved):

PR #522, opened 2022-06-30
Remove 'dim' and make 'dof' optional

### LINKED ISSUE #477: `dim` vs. `dof`
I think that there is an ambiguity between `dof` and `dim` in the constraints. For e.g. a `MultivariateGaussian`:

- `dim` should be the number of observables = dimension of all the entries: means, kinematics, options...
- `dof` should only reflect the number of underlying measurements but is a priori not fixed.

E.g. for an angular analysis, `dim` = `dof` = number of angular observables.
Now, suppose that I concatenate several observables in one Gaussian constrain, I would like to have `1 = dim < dof`. Does this make sense?

mreboud:
Closes #477


eos/constraints/B-to-D-l-nu.yaml
```diff
@@ -37,7 +37,6 @@ B^0->D^+e^-nu::BRs@Belle:2015A:
         - [1.096e-09, 5.959e-09, 1.215e-08, 1.602e-08, 1.860e-08, 2.709e-08, 3.533e-08, 1.022e-07, 4.194e-08, 3.744e-08]
```
dvandyk, line 37: Can you change the commit message? "egal" to "equal"?

eos/constraints/B-to-P-form-factors.yaml
```diff
@@ -8,105 +8,86 @@
 # from [FNAL+MILC:2015A] (info on f_+ and f_0). Since the covariance matrices were singular, we
 # applied the modification suggested in 1901.02561 in a procedure explained in Section 4 of [LMvD:2021A]
 B->pi::f_++f_0+f_T@FNAL+MILC:2015C:
-  type: MultivariateGaussian(Covariance)
```
dvandyk, line 11: I think we need a YAML linter pass by the Github Actions workflow and maybe as a pre-commit hook.

eos/constraint.cc
```diff
@@ -1311,7 +1294,7 @@ namespace eos
         {
```
dvandyk, line 1294: Looking at line 1289 (1306 original) above, i wonder if ``dof`` should now only be emitted if ``dim_meas != dim_pred``. Or could we simply use ``dof == dim_pred`` always?

dvandyk (approved):
General approval. See some nitpicking (commit message, maybe only emit ``dof`` conditionally)

eos/constraints/B-to-P-form-factors.yaml
mreboud: I am not sure I know what you mean :)

eos/constraint.cc
mreboud: No, Christoph sent me another example where `dof < dim_meas = dim_pred` https://github.com/eos/eos/issues/477#issuecomment-1170368309. I think that the user may want to use `dof` to modify the test statistics easily. This is also something we may want to use when implementing a combination of two independent measurements, no?

eos/constraints/B-to-P-form-factors.yaml
dvandyk: You spent time and effort to properly format and indent the YAML file. This can and should be automatized. New commits should be rejected if the YAML files are not following the formatting/indentation rules.

eos/constraint.cc
dvandyk: Very good, I'm happy.

Issue #499, opened 2022-05-10
Split ``eos/form-factors/mesonic-impl.hh`` into individual files


dvandyk:
New files should follow the existing naming convention ``parametric-XYZYEAR{-impl.hh,.hh,.cc,_TEST.cc}``.

dvandyk:
Determine outcome:
 - [ ] ``MvD2016`` (baryon)

Keep and split out
 - [ ] ``BCL2008`` (meson)
 - [ ] ``KMPW2010`` (meson)
 - [ ] ``BSZ2015`` (meson)
 - [ ] ``DKMR2017`` (baryon)
 - [ ] ``HQET`` (meson rename -> ``BGJvD2019``)
 - [ ] ``HQET`` (baryon rename -> ``...``)
 
Remove:
 - [ ] ``BZ2004`` (both of them, ``P->P`` and ``P->V``)
 - [ ] ``BFW2010`` (meson)
 - [ ] ``FMvD2015`` (meson)

mreboud:
What about `FvDV2018` (P->PP)?

PR #525, opened 2022-07-05
Remove obsolete form factor parametrizations

dvandyk:
Removed:
 - ``BZ2004`` and ``BZ2004v2``
 - ``BFW2010`` (only the simplified series expansion was implemented)
 - ``FMvD2015``

In furtherance of #499.


eos/form-factors/mesonic.cc
```diff
@@ -3,7 +3,6 @@
 /*
  * Copyright (c) 2010, 2011, 2013, 2014, 2015, 2016, 2018 Danny van Dyk
  * Copyright (c) 2015 Christoph Bobeth
- * Copyright (c) 2010, 2011 Christian Wacker
```
mreboud, line 6: Why did you removed this entry?

mreboud (approved):
Github complains about missing newlines at the end of most of the new files.
I don't have comments, thanks for having taken care of this!

mreboud:
The parametrizations still need to be removed from:

- [ ] The parameters files
- [ ] All the test case (@mreboud 's job)
- For the doc it should be automatic...

mreboud:
Pff for `FMvD2015`, it is also used in the test case. I will trade it for `BSZ2015`, but the default values for `B_s -> K^*` are not in EOS... So I will start by adding the values for the original `BSZ` paper, and then update all the test cases

eos/form-factors/mesonic.cc
dvandyk: Because his code was moved to a different file.

mreboud: Ah I see, good

dvandyk:
> Pff `FMvD2015` is also used in the test case. I will trade it for `BSZ2015`, but the default values for `B_s -> K^*` are not in EOS... So I will start by adding the values of the original `BSZ` paper, and then update all the test cases

Thank you!

PR #526, opened 2022-07-07
Three small commits

### LINKED ISSUE #523: Duplicate entry in ``eos/parameters/hadronic-matrix-elements.yaml``
The following code only works due to a quirk in how ``yaml-cpp`` parses the file:
https://github.com/eos/eos/blob/7bdac93e3f0bef1227d7dae5afe80891661c8c81/eos/parameters/hadronic-matrix-elements.yaml#L4359-L4376

The key to both entries is identical. It is ambiguous if only the first or only the second entry should persevere. Certainly not both.
(Detected by running ``yamlfmt`` on the source tree in preparation of a YAML formatting and linting pre-commit hook.)

mreboud:
- One commit closes #523
- The two others correct typo found in `rare-b-decays`, one for some parameters name the other for the test statistics in `PMCSampler`


eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -4365,13 +4365,13 @@ groups:
...
-      'B->K^*ccbar::%1%_Hhat_at_%2%_%3%@GRvDV2022' :
+      'B->K^*ccbar::A%1%_Hhat_at_%2%_%3%@GRvDV2022' :
           central: 0
           min:     0
           max:     0
-          latex:   '$\mathrm{%1%}\, \hat{\mathcal{H}}_\%3%^{B\to K^*c\bar{c}}(%2%)$'
+          latex:   '$\mathrm{A%1%}\, \hat{\mathcal{H}}_\%3%^{B\to K^*c\bar{c}}(%2%)$'
           matrix:
-              - [ "Abs", "Arg" ]
+              - [ "bs", "rg" ]
```
dvandyk, line 4374: This works here. Have to think about how to solve this problem in general. Add a special case when the value assigned to ``matrix`` is a map rather than a list? Just brainstorming... Nothing to do here right now.

python/eos/data/native.py
```diff
@@ -340,7 +340,7 @@ def create(path, parameters, proposal, sigma_test_stat=None, samples=None, weigh
 
         # The test statistics defaults to two empty lists
         description['test statistics'] = { "sigma": [], "densities": [] }
-        if sigma_test_stat and samples and weights:
+        if not sigma_test_stat is None and not samples is None and not weights is None:
```
dvandyk, line 343: That's not easy to read. Does the following work?
```python
if sigma_test_stat is not None and samples is not None and weights is not None:
```

python/eos/data/native.py
```diff
@@ -340,7 +340,7 @@ def create(path, parameters, proposal, sigma_test_stat=None, samples=None, weigh
...
             sigma_test_stat = _np.array(sigma_test_stat)
             samplesPDF = [x for x in map(lambda x: -2.0 * _np.log(PMCSampler._evaluate_mixture_pdf(proposal, x)), samples)]
```
dvandyk, line 345: Unrelated, but this looks nasty...

```python
samplesPDF = list(map(lambda x: -2.0 * _np.log(PMCSampler._evaluate_mixture_pdf(proposal, x)), samples))
```
does the trick, doesn't it?

dvandyk (commented):
Left a few comments. Nothing really dramatic, but the ``if`` clause could use some work.

python/eos/data/native.py
mreboud: probably :)

python/eos/data/native.py
mreboud: It does :)

Issue #489, opened 2022-02-21
Improve errors and messages in `Observables::insert`


mreboud:
For the moment, a failed insertion is just reported as `Could not create expression ...`. This can happen for several reasons including the fact that the observable already exists. I think that improving the output in this case and maybe catching this specific error may be beneficial.

dvandyk:
- when inserting/overwriting an existing observable, issue a warning via ``Log``
 - only throw a C++ exception when "a real issue" arises

PR #528, opened 2022-07-10
Ensure that inserted observables overwrite existing observables

dvandyk:
Resolves issue #489


mreboud (approved):
Looks good, thank you Danny :)

PR #533, opened 2022-07-21
[eos] Fix the Warning message in the insertion of observables

### LINKED ISSUE #529: Useless warning message in the tests
Due to the commit c7cd161fda7ab045f0021fab3fe8d71092c44e43, we now have an annoying and useless message when testing:
```[WARNING [ObservableEntries.insert_or_assign]] Entry for observable test::obs1 has been replaced.```

This comes from this line https://github.com/eos/eos/blob/c7cd161fda7ab045f0021fab3fe8d71092c44e43/test/test.cc#L117
Can we fix that by e.g. protecting the insertion in an `#ifndef` C++ statement so that it happens only once?

### PR DESCRIPTION
**Pull request by mreboud**
Resolves #529

### DIFF
--- a/eos/observable.cc
+++ b/eos/observable.cc
@@ -71,7 +71,7 @@ namespace eos
     {
         auto result = _entries->insert_or_assign(key, value);
 
-        if (result.second) 
+        if (! result.second)
         {
             Log::instance()->message("[ObservableEntries.insert_or_assign]", ll_warning)
                 << "Entry for observable " << key.str() << " has been replaced.";

### CODE REVIEW

File: eos/observable.cc [hunk 1]
```diff
@@ -71,7 +71,7 @@ namespace eos
     {
         auto result = _entries->insert_or_assign(key, value);
 
-        if (result.second) 
+        if (!result.second)
```
dvandyk, line 74: Coding style: ``!`` is separated from its argument like every operator is.

dvandyk (approved):
Fine except for the coding style.

PR #536, opened 2022-08-10
[github] Upgrade MacOS workflow to python 3.10

mreboud:


.github/workflows/macos-build+check.yaml
```diff
@@ -31,16 +31,16 @@ jobs:
...
                 brew tap eos/eos
-                brew outdated python@3.9 || brew upgrade python@3.9 || true
+                brew outdated python@3.10 || brew upgrade python@3.10 || true
                 rm -Rf '/usr/local/lib/python2.7/site-packages/numpy/'
                 brew outdated boost || brew upgrade boost || true
                 brew outdated boost-python3 || brew upgrade boost-python3 || true
                 brew install autoconf automake boost-python3 gsl libtool pkg-config yaml-cpp
-                brew link --overwrite --force python@3.9
+                brew link --overwrite --force python@3.10
                 $PYTHON -m pip install -U cython h5py matplotlib networkx numpy "pypmc>=1.2" PyYAML scipy wilson
```
dvandyk, line 44: @mreboud can you split this line into two, and see if things work if ``numpy`` is definitely installed before it tries to build ``pypmc``?

PR #541, opened 2022-08-24
Kaon lcdas

cbolognani:
Used reference 1903.08038 for a1 and a2 values at 1 GeV, not updating twist-3 values (f3K, omega3K... naming need to be defined)


eos/form-factors/k-lcdas.cc
```diff
@@ -0,0 +1,493 @@
...
+
+            return f3K_0 * lambda3K_0 * std::pow(c_rge, 139.0 / 18.0)
+                - 14.0 / 67.0 * (std::pow(c_rge, 4.0)        - std::pow(c_rge, 139.0 / 18.0)) * f_K * m_s_0
+                + 14.0 /  5.0 * (std::pow(c_rge, 68.0 / 9.0) - std::pow(c_rge, 139.0 / 18.0)) * f_K * m_s_0 * a1K_0
+                - 4.0  / 11.0 * (std::pow(c_rge, 86.0 / 9.0) - std::pow(c_rge, 139.0 / 18.0)) * f_K * m_s_0 * a2K_0;
+        }
+
+        inline double eta3K(const double & mu) const
+        {
+            return f3K(mu) / (f_K() * muK(mu));
```
SKuerten, line 159: I do not understand why there are paranthesis for the "f_K()" here. Can someone explain this to me?

SKuerten (commented):
I think this is flawless.

eos/form-factors/k-lcdas.cc
dvandyk: The type of the object named ``f_K`` is ``UsedParameter``. This class provides two operators to obtain the parameter's value:
 - a conversion operator to ``double``, which is triggered implicitly by using the parameter in place of a ``double`` anywhere, or explicitly by using ``static_cast<double>``; and
 - a function call operator, which is triggered by treating the object like a function and call it (without any arguments).

dvandyk: https://github.com/eos/eos/blob/19d11d6be8042a1e279146388109a3e0002f7b0a/eos/utils/parameters.hh#L247-L251

eos/form-factors/k-lcdas.cc
```diff
@@ -0,0 +1,493 @@
...
+        // strange quark mass
+        const double m_s = _imp->model->m_s_msbar(mu);
+        const double m_ud = _imp->model->m_ud_msbar(mu) / 2.0;
+
+        // Twist 2 Gegenbauer coefficients
+        const double a1K = _imp->a1K(mu);
+        const double a2K = _imp->a2K(mu);
+
+        // Twist 3 coefficients
+        const double rhopK    = power_of<2>((m_s + m_ud) / _imp->m_K); // EOM constaints, cf. [BBL:2006A], cf. eq. (3.12)
```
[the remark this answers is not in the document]
dvandyk: @cbolognani just fixed it. Thank you @SKuerten!

PR #542, opened 2022-08-24
B->PSD-nu-nu add scalar and tensor modes & tests

lorenzennio:


eos/b-decays/b-to-psd-nu-nu.cc
```diff
@@ -52,6 +52,10 @@ namespace eos
 
         UsedParameter m_P;
 
+        UsedParameter m_b;
```
dvandyk, line 55: You don't need these and should not use them :-)

You need the b and s quark masses at the scale of your WET. You can obtain it from your model using ``model->m_b_msbar(mu)`` and ``model->m_s_msbar(mu)``.

You need to add a new ``UsedParameter`` for the scale though. The name of the parameter should be ``sbnunu::mu``, if it has not yet been added to the list of parameters.

eos/b-decays/b-to-psd-nu-nu.cc
```diff
@@ -143,6 +147,8 @@ namespace eos
             m_B(p["mass::" + _B()], u),
             tau_B(p["life_time::" + _B()], u),
             m_P(p["mass::" + _P()], u),
+            m_b(p["mass::b(MSbar)"], u),
```
dvandyk, line 150: Ditto, remove.

eos/b-decays/b-to-psd-nu-nu.cc
```diff
@@ -176,11 +182,16 @@ namespace eos
         {
             const double m_B = this->m_B(), m_B2 = m_B * m_B;
             const double m_P = this->m_P(), m_P2 = m_P * m_P;
+            const double m_b = this->m_b();
```
dvandyk, line 185: Ditto, see above.

eos/b-decays/b-to-psd-nu-nu.cc
```diff
@@ -189,10 +200,14 @@ namespace eos
                 * sqrt_lambda / (power_of<3>(4.0 * M_PI * m_B));
 
             // first term in square brackets in [FLS:2021A], eq. (8)
-            const double contr_vector = lambda / 24.0 * f_p * f_p * std::norm(wc.cL() +  wc.cR());
+            const double contr_vector = lambda / 24.0 * f_p * f_p * std::norm(wc.cVL() +  wc.cVR());
+            // second line in [FLS:2021A], eq. (8) (ignoring the {bs} Wilson coefficients)
+            const double contr_scalar = q2 * power_of<2>((m_B2 - m_P2)/(m_b - m_s)) / 8 * f_0 * f_0 * std::norm(wc.cSL() +  wc.cSR());
```
dvandyk, line 205: Coding style: please leave a single white space to the left and right of every binary operator, i.e.:
``(...) / (m_b - m_s)`` rather than ``(...)/(m_b - m_s)``.

eos/b-decays/b-to-psd-nu-nu.cc
```diff
@@ -189,10 +200,14 @@ namespace eos
                 * sqrt_lambda / (power_of<3>(4.0 * M_PI * m_B));
 
             // first term in square brackets in [FLS:2021A], eq. (8)
-            const double contr_vector = lambda / 24.0 * f_p * f_p * std::norm(wc.cL() +  wc.cR());
+            const double contr_vector = lambda / 24.0 * f_p * f_p * std::norm(wc.cVL() +  wc.cVR());
+            // second line in [FLS:2021A], eq. (8) (ignoring the {bs} Wilson coefficients)
```
dvandyk, line 204: I think I agree with your reading of the formula, but I'm not 100% certain. @mreboud @mbordone can you wade in to confirm or refute that our interpretation of this is correct?

eos/models/wet.cc
```diff
@@ -471,10 +471,16 @@ namespace eos
 
     /* sbnunu Wilson coefficients */
     WilsonScanComponent<components::WET::SBNuNu>::WilsonScanComponent(const Parameters & p, const Options &, ParameterUser & u) :
-    _re_cl(p["sbnunu::Re{cL}"], u),
-    _im_cl(p["sbnunu::Im{cL}"], u),
-    _re_cr(p["sbnunu::Re{cR}"], u),
-    _im_cr(p["sbnunu::Im{cR}"], u)
+    _re_cvl(p["sbnunu::Re{cVL}"], u),
```
dvandyk, line 474: For some reason the coding style of the original was incorrect. Can you please indent these lines by one level to the right?

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
          * These two operators are defined via
          *
          * H^eff = 4 G_F / sqrt(2) V_tb V_ts^* (
-         *             C_L alpha_e / (2 pi) * [sbar gamma^mu P_L b] [nubar gamma_mu P_L nu]
-         *             C_R alpha_e / (2 pi) * [sbar gamma^mu P_R b] [nubar gamma_mu P_L nu]
+         *             C_L alpha_e / (2 pi) * [nubar gamma_mu P_L nu] [sbar gamma^mu P_L b]
```
dvandyk, line 171: Please keep the original order of [sb] and [nu nubar] currents. We address the operators as ``sbnunu``, not ``nunusb``.

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
-         *             C_L alpha_e / (2 pi) * [sbar gamma^mu P_L b] [nubar gamma_mu P_L nu]
-         *             C_R alpha_e / (2 pi) * [sbar gamma^mu P_R b] [nubar gamma_mu P_L nu]
+         *             C_L alpha_e / (2 pi) * [nubar gamma_mu P_L nu] [sbar gamma^mu P_L b] 
+         *             C_R alpha_e / (2 pi) * [nubar gamma_mu P_L nu] [sbar gamma^mu P_R b] 
          *         )
-         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_L O_L + C_R O_R )
+         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_VLL O_VLL + C_VLR O_VLR )
+         * 
+         * Furthermore, we can add the scalar modes [FLS:2021A], eq. (2)
+         * O_SLL = [bar(nu_L^c) P_L nu][sbar P_L b], where nu_L^c = C bar(nu_L)^T, C = i gamma^2 gamma^0
```
dvandyk, line 177: Your nu_L is incorrect, is it not? ``nu_L^c = (C nubar_L)^T`` ?

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
          * H^eff = 4 G_F / sqrt(2) V_tb V_ts^* (
-         *             C_L alpha_e / (2 pi) * [sbar gamma^mu P_L b] [nubar gamma_mu P_L nu]
-         *             C_R alpha_e / (2 pi) * [sbar gamma^mu P_R b] [nubar gamma_mu P_L nu]
+         *             C_L alpha_e / (2 pi) * [nubar gamma_mu P_L nu] [sbar gamma^mu P_L b] 
+         *             C_R alpha_e / (2 pi) * [nubar gamma_mu P_L nu] [sbar gamma^mu P_R b] 
          *         )
-         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_L O_L + C_R O_R )
+         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_VLL O_VLL + C_VLR O_VLR )
+         * 
+         * Furthermore, we can add the scalar modes [FLS:2021A], eq. (2)
```
dvandyk, line 176: ``modes`` -> ``operators``

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
+         * O_SLL = [bar(nu_L^c) P_L nu][sbar P_L b], where nu_L^c = C bar(nu_L)^T, C = i gamma^2 gamma^0
+         * O_SLR = [bar(nu_L^c) P_L nu][sbar P_R b], where nu_L^c = C bar(nu_L)^T, C = i gamma^2 gamma^0
+         * 
+         * And tensor modes [FLS:2021A], eq. (2)
+         * O_TLL = [bar(nu_L^c) sigma^{mu nu} P_L nu][sbar sigma_{mu nu} P_L b], where nu_L^c = C bar(nu_L)^T, C = i gamma^2 gamma^0
+         * 
+         * 
          *
          * C++ idx -> operator
          * 0       -> O_L
```
dvandyk: Please adjust the notation of the vector operators to reflect your renaming in the code, i.e., use ``VLL` and ``VRL``.

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
          * 1       -> O_R
+         * 2       -> O_SLL
+         * 3       -> O_SLR
```
dvandyk, line 189: This should be ``SRL``, since the flavour quantum numbers are ``sbnunu``, not ``nunusb``.

dvandyk: Ah, it's in the next commit!

dvandyk (changes requested):
Looks very good already, but still requires some changes!

I would prefer if you could split the first commit a bit further, and squash some part of the second commit (fixing the comments in the Wilson coefficients) into an earlier commit.

I think the order of commits should be:
 - rework ``SBNuNu`` interface ``ModelComponent<WET::SBNuNu>`` and similar, along with all users (B->Knunu and B->K*nunu!), without introducing the code for the scalar and tensors yet.
  I.e., this commit should not change any observable.
 - add the SM vs WET test case for the SBNuNu sector.
 - change the B->Knunu code to make use of the new operators

Ping me if you have any questions!

lorenzennio:
Thanks for the comments. I will try to change the commits accordingly. Just so we are on the same page, with reworking the `sbnunu` interface, you mean to introduce all additional WC variables in all parts of the code, not implementing the full [FLS:2021A], eq. (8) (which I then do in commit 3)?

eos/models/wilson-coefficients.hh
lorenzennio: I think it is correct, at least according to [FLS:2021A] page 4, second paragraph. Since C is a matrix and nubar_L is a row vector, so I should transpose it before applying C, if I am not mistaken. Please correct me if I'm wrong (:

dvandyk: You are entirely correct, I had a knot in my brain.

eos/models/wet.cc
lorenzennio: It is the same for `WilsonScanComponent<components::WET::CBLNu>::WilsonScanComponent`, I will change it there too.

eos/models/wet.cc
```diff
@@ -366,56 +366,56 @@ namespace eos
 
     /* b->c Wilson coefficients */
     WilsonScanComponent<components::WET::CBLNu>::WilsonScanComponent(const Parameters & p, const Options &, ParameterUser & u) :
-    _e_re_csl(p["cbenue::Re{cSL}"], u),
```
dvandyk, line 369: Can you please move this change to a separate commit, since it is unrelated to this one?

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
          * H^eff = 4 G_F / sqrt(2) V_tb V_ts^* (
-         *             C_L alpha_e / (2 pi) * [sbar gamma^mu P_L b] [nubar gamma_mu P_L nu]
-         *             C_R alpha_e / (2 pi) * [sbar gamma^mu P_R b] [nubar gamma_mu P_L nu]
+         *             C_L alpha_e / (2 pi) * [sbar gamma^mu P_L b][nubar gamma_mu P_L nu] 
+         *             C_R alpha_e / (2 pi) * [sbar gamma^mu P_R b][nubar gamma_mu P_L nu] 
          *         )
-         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_L O_L + C_R O_R )
+         *       = 4 G_F / sqrt(2) V_tb V_ts^* ( C_VLL O_VLL + C_VRL O_VRL )
+         * 
+         * Furthermore, we can add the scalar operators [FLS:2021A], eq. (2)
```
dvandyk, line 176: I think this comment should include information that the new operators only emerge once you go beyond the SM content of particle, i.e., after including right-handed neutrino fields.

eos/models/wilson-coefficients.hh
```diff
@@ -168,23 +168,37 @@ namespace eos
...
+         * 
+         * And tensor operators [FLS:2021A], eq. (2)
+         * O_TLL = [sbar sigma_{mu nu} P_L b][bar(nu_L^c) sigma^{mu nu} P_L nu], where nu_L^c = C nubar_L^T, C = i gamma^2 gamma^0
+         * 
+         * 
          *
          * C++ idx -> operator
-         * 0       -> O_L
-         * 1       -> O_R
+         * 0       -> O_VLL
```
dvandyk, line 186: Your comment differes from the names used for the class, i.e., you write ``O_VLL`` but the WC is named ``cVL``.
If given the choice, I would prefer the latter convention.

dvandyk (changes requested):
(Very) minor changes requested.

eos/models/wilson-coefficients.hh
lorenzennio: I am still struggling a bit with fully understanding the physical meaning of all the individual operators. This is not extremely well described in the (paper)[https://arxiv.org/pdf/2111.04327.pdf]. I also had a look at (this)[https://link.springer.com/content/pdf/10.1007/JHEP09(2017)158.pdf]  and (this)[https://journals.aps.org/prd/pdf/10.1103/PhysRevD.60.094007] paper, but I am not yet very confident in my understanding. Do you have any good references that explain the physical consequences of the operators?
For example, I don't really see where the right-handed neutrino fields come in - because nu_L^c = (nu^c)_R corresponds to a RH antineutrino according to these (notes)[https://arxiv.org/pdf/hep-ph/0001264.pdf].

mreboud: You are mixing a couple of things here and I am not sure that Github is the best place for this discussion :) Let me try to give a quick answer and if you have further questions we can discuss them live.

- You are working in an effective theory, where all the heavy particles are "integrated out". The integrated fields can however still impact the physics at the low scale -- here the B decays -- indirectly. In the effective theory you therefore need to consider all the possible interactions (= operators) of your particles since some of these interactions may be enhanced by the heavy fields.

- In the SM, things are simple. You don't have right-handed neutrinos (so all the corresponding operators have a zero coefficient) and the effect of the SM heavy particles has been computed.

- You can now consider several NP scenarios: without or without right-handed neutrinos. Without RH neutrinos, you expect the NP heavy particles to modify the coefficients of your operators, by creating or modifying the interactions of your fields at the low scale. This is usually due to loop processes, where the NP particles can now appear. What your are studying is the sensitivity of the B->K(*) nu nu decay to these modifications.

- In other NP scenarios, right-handed neutrinos exist. If they have a small mass, they cannot be integrated out from your effective theory and you have to consider these fields even at the low-scale. These gives you new operators with non-zero coefficients. You are adding them as well because it is interesting to see if these new operators can be constrained by the future measurements.

- Finally, each operators involve its own fields with a specific chiralities and a given spin structure. They therefore have a different impact on the observable, especially on the angular distributions. So fitting the data in EOS allows to get a lot of information on the allowed ranges for the coefficients in the effective theory.

dvandyk:
I think I dropped the ball here. @mreboud can I merge this PR?

PR #543, opened 2022-09-10
Fix various warnings

dvandyk:
- remove an unnecessary `const` type qualifier
 - replace `std::uncaught_exception` with `std::uncaught_exceptions`
 - mark an unused variable as unused in the GSL hacks
 - only check verbosity in `eos-analysis` if a subcommand is invoked

### DIFF
--- a/eos/utils/gsl-hacks.cc
+++ b/eos/utils/gsl-hacks.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2010 Danny van Dyk
+ * Copyright (c) 2010, 2022 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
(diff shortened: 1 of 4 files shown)


eos/utils/gsl-hacks.cc
```diff
@@ -68,7 +68,7 @@ namespace eos
 
         void gsl_error_handler_hack()
         {
-            gsl_error_handler_t * previous_gsl_error_handler = gsl_set_error_handler(&gsl_error_handler);
+            gsl_error_handler_t * previous_gsl_error_handler __attribute__ ((unused)) = gsl_set_error_handler(&gsl_error_handler);
```
mreboud, line 71: If it is unused why do we define it?

mreboud (approved):
Thanks @dvandyk

eos/utils/gsl-hacks.cc
dvandyk: Otherwise we'll get a warning about a discarded return value!

PR #552, opened 2022-10-13
[form-factors] Add a BSZ-like parametrization for D -> P transitions

mreboud:


eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -3045,6 +3045,43 @@ groups:
           unit:     'GeV^-2'
   # }}}
 
+  # Parameters in D -> P Form Factor Parametrizations
+  # {{{
+  - title: 'Parameters in $B\to P$ Form Factor Parametrizations'
```
dvandyk, line 3050: D, rather than B

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -3045,6 +3045,43 @@ groups:
...
+    description: ''
+    parameters:
+      # form factor parameters for D->pi inspired by [BSZ:2015A]
+      'D->Pi::alpha^%1%_%2%@BSZ2015' :
```
dvandyk, line 3054: ``Pi`` -> ``pi``, must match line 97

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -3045,6 +3045,43 @@ groups:
...
+          central:  0.0
+          min:      0.0
+          max:      0.0
+          latex:    '$\alpha_{%1%,%2%}^{D \to \pi,\mathrm{BSZ2015}}$'
+          unit:     '1'
+          matrix:
```
dvandyk, line 3060: Ah, this is going to be a problem. One of the form factor parameters in ``f_0`` is idenfitied with an ``f_+`` parameter. So you create more parameters than needed. Which is bad, since an unknowing user might want to fit it and will not see it affected in a fit!

dvandyk (changes requested):

eos/parameters/hadronic-matrix-elements.yaml
mreboud: Sure but this is treated internally, the spurious parameter will never been called, so the user can try to fit it, it just doesn't appear in the likelihood. I would keep it as it is now, unless you are strongly against. And this will be removed as soon as we provide default values for these parameters :)

dvandyk: I like that solution!

dvandyk (approved):

PR #556, opened 2022-11-03
Corner plot

fnovak42:
Added corner plot as a task and parser.

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -460,6 +460,64 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
     eos.data.DynestyResults.create(os.path.join(base_directory, posterior, 'dynesty_results'), analysis.varied_parameters, results)
 
 
+# Corner plot
+@task('corner-plot', '{posterior}/plots')
+def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
+    """
+    Generates a corner plot of the 1-D and 2-D marginalized posteriors.
+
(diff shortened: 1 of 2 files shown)


src/scripts/eos-analysis
```diff
@@ -569,6 +591,11 @@ def cmd_sample_nested(args):
     return eos.sample_nested(**args_to_dict(args))
 
 
+# Nested sampling
```
dvandyk, line 594: Copy und Paste mistake

python/eos/tasks.py
```diff
@@ -460,6 +460,64 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
...
+    The output files will be stored in EOS_BASE_DIRECTORY/POSTERIOR/plots.
+
+    :param analysis_file: The name of the analysis file that describes the named posterior, or an object of class `eos.AnalysisFile`.
+    :type analysis_file: str or `eos.AnalysisFile`
+    :param posterior: The name of the posterior.
+    :type posterior: str
+    :param base_directory: The base directory for the storage of data files. Can also be set via the EOS_BASE_DIRECTORY environment variable.
+    :type base_directory: str, optional
+    """
+    p = eos.Parameters()
```
dvandyk, line 479: I think we should

1. Create the analysis,
2. Use its parameters object rather than the defaults.

Would make the task usable even when we let an analysis modify the latex representations of its parameters.

dvandyk (changes requested):
Nice, as few comments apply. Also, the 2nd commit's message is not very meaningful.

PR #557, opened 2022-11-05
Corner plot and storing nested sampling results

fnovak42:
The option to visualize marginalized posteriors using a corner plot as an `eos task` was added. A mistake from the previous implementation of the sample-nested task was corrected, namely the calculation of the samples so that they can be stored using `eos.data.ImportanceSamples`.


mreboud:
Thanks @fnovak42! Can you merge your commits 3 and 4 to 1 and 2 respectively? In this end you should only have a python commit for the corner plot, an infra commit for the parser and a python commit for the fix-up. Also, as discussed previously you can improve the label of your commits, e.g. w/o reading your PR, I don't know what your second commit does.

mreboud (approved):
Looks good to me!

PR #566, opened 2022-11-26
[python] Fix typos related to the corner_plot and sample_nested functions

fnovak42:
Typos with regards to the initial implementation of the `corner-plot` functionality in `tasks.py` and the CLI command name was changed from `corner_plot` to `corner-plot`.

Issues related to the storage of nested sampling results were also fixed; they can now be stored as `eos.data.ImportanceSamples` objects, so that they can be used for the plotting functions. The previous version of this code also included a redundant application of the `analysis._x_to_par` function (scaling from a unit hypercube to the parameter space of interest).


python/eos/tasks.py
```diff
@@ -455,7 +455,7 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
     """
     analysis = analysis_file.analysis(posterior)
     results = analysis.sample_nested(bound=bound, nlive=nlive, dlogz=dlogz, maxiter=maxiter)
-    samples = np.array(list(map(analysis._x_to_par, results.samples)))
+    samples = _np.array(list(map(analysis._x_to_par, results.samples)))
```
dvandyk: Was this not supposed to be corrected with respect to the calls to ``_x_to_par``?

python/eos/tasks.py
```diff
@@ -487,15 +488,15 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
+        xmin = _np.min(samples)
+        xmax = _np.max(samples)
         ax.set_xlim((xmin, xmax))
 
         ax.set_xlabel(p[i].latex())
         ax.set_ylabel(p[i].latex())
 
         ax.hist(samples, weights=f.weights, alpha=0.5, bins=100, density=True, stacked=True, color='C1')
-        ax.set_aspect(np.diff((xmin, xmax))[0] / np.diff(ax.get_ylim())[0])
+        #ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff(ax.get_ylim())[0])
```
dvandyk: If this line is not needed, why is it not removed?

python/eos/tasks.py
```diff
@@ -507,17 +508,17 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
             ysamples = f.samples[:, i]
-            xmin = np.min(xsamples)
-            xmax = np.max(xsamples)
-            ymin = np.min(ysamples)
-            ymax = np.max(ysamples)
+            xmin = _np.min(xsamples)
+            xmax = _np.max(xsamples)
+            ymin = _np.min(ysamples)
+            ymax = _np.max(ysamples)
+            #ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff((ymin, ymax))[0])
```
dvandyk: Ditto.

python/eos/tasks.py
```diff
@@ -507,17 +508,17 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
+            ax.hist2d(xsamples, ysamples, weights=f.weights, alpha=1.0, bins=100, cmap='Greys')
             ax.set_xlim((xmin, xmax))
             ax.set_ylim((ymin, ymax))
-            ax.set_aspect(np.diff((xmin, xmax))[0] / np.diff((ymin, ymax))[0])
-            ax.hist2d(xsamples, ysamples, weights=weights, alpha=1.0, bins=100, cmap='Greys')
 
     fig.tight_layout()
-    fig.savefig(os.path.join(base_directory, posterior, 'plots', 'corner-plot.pdf'))
+    fig.savefig(os.path.join(base_directory, posterior, 'plots', 'corner-plot.png'))
```
dvandyk: I think we discussed that this should be an optional keyword for the task, e.g., ``format='pdf'``.

python/eos/tasks.py
```diff
@@ -455,15 +455,15 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
     """
     analysis = analysis_file.analysis(posterior)
     results = analysis.sample_nested(bound=bound, nlive=nlive, dlogz=dlogz, maxiter=maxiter)
-    samples = _np.array(list(map(analysis._x_to_par, results.samples)))
+    samples = results.samples
```
dvandyk, line 458: Ok, this is confusing. Why is this contained in a different commit? Please create three clean commits:
 - one that fixes the task ``sample_nested``
 - one that fixes the task ``corner_plot``
 - one that fixes the naming of the ``eos-analysis`` subcommand.

python/eos/tasks.py
```diff
@@ -455,15 +455,15 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
...
     weights = _np.exp(results.logwt - results.logz[-1])
     eos.data.DynestyResults.create(os.path.join(base_directory, posterior, 'dynesty_results'), analysis.varied_parameters, results)
     eos.data.ImportanceSamples.create(os.path.join(base_directory, posterior, 'samples'), analysis.varied_parameters, samples, weights)
 
 
 # Corner plot
 @task('corner-plot', '{posterior}/plots')
-def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
+def corner_plot(analysis_file:str, posterior:str, base_directory:str='./', file_extension:str='.pdf'):
```
dvandyk, line 466: This should probably be a fourth commit then, since it is not fixing any issues in the ``corner_plot`` task but adds new functionality.

python/eos/tasks.py
```diff
@@ -476,27 +476,28 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
         xmin = _np.min(samples)
         xmax = _np.max(samples)
         ax.set_xlim((xmin, xmax))
-
         ax.set_xlabel(p[i].latex())
         ax.set_ylabel(p[i].latex())
-
-        ax.hist(samples, weights=f.weights, alpha=0.5, bins=100, density=True, stacked=True, color='C1')
-        #ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff(ax.get_ylim())[0])
+        ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff(ax.get_ylim())[0])
```
dvandyk, line 500: Now you are undoing your previous commit. Why?

python/eos/tasks.py
```diff
@@ -512,13 +513,19 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
-            #ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff((ymin, ymax))[0])
             ax.hist2d(xsamples, ysamples, weights=f.weights, alpha=1.0, bins=100, cmap='Greys')
             ax.set_xlim((xmin, xmax))
             ax.set_ylim((ymin, ymax))
+            ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff((ymin, ymax))[0])
 
     fig.tight_layout()
-    fig.savefig(os.path.join(base_directory, posterior, 'plots', 'corner-plot.png'))
+
+    # This could further be modified with a list of allowed output formats
```
dvandyk, line 523: I would prefer ``format`` over ``file_extensions``.
Also, I think this is not very clean. I think the pythonic way to handle this is:
```py
_format = copy(format)
if isinstance(_format, str):
    _format = [ _format ]
for f in _format:
    fig.savefig(os.path.join(base_directory, posterior, 'plots', f'corner-plot.{f}'))
```
Copying from the argument is absolutely needed to avoid changing the default value.

dvandyk (changes requested):
Needs some changes, please see my comments.

dvandyk (approved):
This looks much better, thank you very much! I think we can merge it as is, if @mreboud agrees!

mreboud:
I do!

PR #567, opened 2022-11-30
Address issues 561 and 307

dvandyk:


eos/utils/exception.cc
```diff
@@ -19,17 +19,109 @@
...
+
+    Context::Context(const std::string & entry, const std::source_location location)
+    {
+        impl::context.push_back(std::make_tuple(location, entry));
+    }
+
+    Context::Context(const Context & other) = default;
+
+    constexpr Context &
+    Context::operator= (const Context & other) = default;
```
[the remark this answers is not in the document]
dvandyk: Would be synthesized anyway. This way i'm being explicit!

.github/workflows/ubuntu-build+check+deploy.yaml
```diff
@@ -18,13 +18,12 @@ jobs:
...
                 cfg:
-                    - { os: focal,    cxx: g++,          cxxflags: '-O2 -g -march=x86-64' }
-                    - { os: focal,    cxx: clang++-12,   cxxflags: '-O2 -g -march=x86-64' }
                     - { os: jammy,    cxx: g++,          cxxflags: '-O2 -g -march=x86-64' }
+                    - { os: jammy,    cxx: clang++-14,   cxxflags: '-O2 -g -march=x86-64' }
         name: Build on ${{ matrix.cfg.os }} using ${{ matrix.cfg.cxx }}
         runs-on: ubuntu-20.04
         container:
-            image: eoshep/ubuntu-${{ matrix.cfg.os }}:latest
+            image: eoshep/ubuntu-${{ matrix.cfg.os }}:ae4ca8649bf120510f83ff13b95ef2c0abc49e70
```
mreboud, line 26: So here you are fixing a specific version, right? Why?

dvandyk: a) Versioning of the image is prudent if you make breaking changes. I should have done that before.
b) In this specific case, I made two successive changes to the docker images, but they would have broken ongoing workflows.

dvandyk: It is the full SHA1 id of the commit that produced the docker image over in eos/docker.io.

mreboud: Ok ok, should this be documented somewhere? E.g. this should be adapted when pushing a new release, no?

dvandyk: I just added it to the wiki: https://github.com/eos/eos/wiki/How-to-bump-the-docker-images

mreboud (approved):
Cool, thanks @dvandyk!

PR #568, opened 2022-12-05
[python] Add format option for storage of corner plots

fnovak42:
An optional 'format' input parameter was added, which may contain one string or a list of possible file extensions that specify how corner plots are stored. If left unspecified, it defaults to 'pdf'.

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -25,6 +25,7 @@ import os
 import pypmc
 import scipy
 import sys
+import copy as _copy
 
 from .ipython import __ipython__
 
@@ -463,7 +464,7 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
 
 # Corner plot
 @task('corner-plot', '{posterior}/plots')
-def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
+def corner_plot(analysis_file:str, posterior:str, base_directory:str='./', format:str='pdf'):
     """
     Generates a corner plot of the 1-D and 2-D marginalized posteriors.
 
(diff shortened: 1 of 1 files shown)


python/eos/tasks.py
```diff
@@ -517,7 +520,12 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
             ax.set_aspect(_np.diff((xmin, xmax))[0] / _np.diff((ymin, ymax))[0])
 
     fig.tight_layout()
-    fig.savefig(os.path.join(base_directory, posterior, 'plots', 'corner-plot.pdf'))
+
+    _format = _copy.copy(format)
+    if isinstance(_format, str):
+        _format = [ _format ]
+    for f in _format:
```
mreboud, line 527: Hmm, ok... but can we do without `_copy`?:
```python
for f in np.array([format]).flatten():
    ....
```

dvandyk: That would work. So does the solution with ``copy``. Is there a problem I am not seeing?

python/eos/tasks.py
```diff
@@ -476,8 +476,11 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
     :type posterior: str
     :param base_directory: The base directory for the storage of data files. Can also be set via the EOS_BASE_DIRECTORY environment variable.
     :type base_directory: str, optional
+    :param format: The file extension of the data files. Can also be a list of file extensions.
+    :type format: str or list, optional
```
dvandyk, line 480: This requires a specification of what values ``format`` or its elements may take. A referal to the matplotlib's documentation and output formats would suffice I think.

python/eos/tasks.py
```diff
@@ -476,8 +476,11 @@ def corner_plot(analysis_file:str, posterior:str, base_directory:str='./'):
...
     """
     import matplotlib.pyplot as _plt
+    import copy as _copy
```
dvandyk, line 483: Isn't ``copy`` part of the Python standard library? If yes, please import it at the top of the file rather than inside this task.

dvandyk (changes requested):
Thank you @fnovak42! I have two suggestions for small modifications.

python/eos/tasks.py
dvandyk: Also, ``str or list of str`` is a better description.

mreboud:
@fnovak42 you messed up the rebasing I think

PR #571, opened 2023-01-05
[parameters] Make definition of sbnunu WET Parameters consistent

lorenzennio:


eos/parameters/sm-and-eft.yaml
```diff
@@ -1322,7 +1322,7 @@ groups:
         The list of parameters describing the $\bar{s}b\bar{s}b$ Wilson coefficients in the effective Hamiltonian
         \begin{equation}
           \mathcal{H}^{\bar{s}b\bar{\nu}\nu}_\mathrm{eff}
-          = \frac{4 G_F}{\sqrt{2}} V_{tb}^* V_{ts} \sum_i \mathcal{C}^{\bar{s}b\bar{\nu}\nu}_i \mathcal{O}^{\bar{s}b\bar{\nu}\nu}_i\,.
+          = \frac{4 G_F}{\sqrt{2}} \frac{\alpha_e}{2 \pi} V_{tb}^* V_{ts} \sum_i \mathcal{C}^{\bar{s}b\bar{\nu}\nu}_i \mathcal{O}^{\bar{s}b\bar{\nu}\nu}_i\,.
```
dvandyk, line 1325: Is this in agreement with the current implementation?

lorenzennio: Yes, it is then consistent with 
https://github.com/eos/eos/blob/master/eos/models/wilson-coefficients.hh#L174
and the implementation here
https://github.com/eos/eos/blob/master/eos/b-decays/b-to-psd-nu-nu.cc#L192

dvandyk:
While discussing with @mreboud we noticed that the test cases checking for the identity of the SM values and the WET parameters erroneously do not use the default parameters:
https://github.com/eos/eos/blob/4040eb54223b55da602c0de9e66d7ca97bc1d4a1/eos/models/wet_TEST.cc#L286

lorenzennio:
Ok, everything seems to work again now.
Just as a remark which I had a chat with @mreboud about too - The sbnunu::Re{cVL} parameter value is set here:
https://github.com/eos/eos/blob/master/eos/parameters/sm-and-eft.yaml#L1348-L1350
But every time the SM one changes (because of another default parameter change, as was the case now), this needs to be updated. To avoid this, could we set the default WCs to the calculated, SM matched ones from here
https://github.com/eos/eos/blob/master/eos/models/standard-model.cc#L894-L947
automatically somehow? Or is there a good reason not to do this?

eos/models/wet_TEST.cc
```diff
@@ -283,7 +283,7 @@ class WilsonCoefficientsSBNuNuTest :
             {
                 static const double eps = 1e-8;
 
-                Parameters p = reference_parameters();  
+                Parameters p = Parameters::Defaults();
```
mreboud, line 286: Can you remove `reference_parameters` entirely and trade it for `Parameters::Defaults` wherever it is used in this file?

mreboud (changes requested):
I just a need a small change. The rest is good to be merged.

dvandyk:
> Ok, everything seems to work again now. Just as a remark which I had a chat with @mreboud about too - The sbnunu::Re{cVL} parameter value is set here: https://github.com/eos/eos/blob/master/eos/parameters/sm-and-eft.yaml#L1348-L1350 But every time the SM one changes (because of another default parameter change, as was the case now), this needs to be updated. To avoid this, could we set the default WCs to the calculated, SM matched ones from here https://github.com/eos/eos/blob/master/eos/models/standard-model.cc#L894-L947 automatically somehow? Or is there a good reason not to do this?

I do not see how this is feasible. Maybe we can discuss this in more detail?

lorenzennio:
> > Ok, everything seems to work again now. Just as a remark which I had a chat with @mreboud about too - The sbnunu::Re{cVL} parameter value is set here: https://github.com/eos/eos/blob/master/eos/parameters/sm-and-eft.yaml#L1348-L1350 But every time the SM one changes (because of another default parameter change, as was the case now), this needs to be updated. To avoid this, could we set the default WCs to the calculated, SM matched ones from here https://github.com/eos/eos/blob/master/eos/models/standard-model.cc#L894-L947 automatically somehow? Or is there a good reason not to do this?
> 
> I do not see how this is feasible. Maybe we can discuss this in more detail?

Sure, I'd be happy to. I just thought it would be nice for consistency, but of course there is no rush to implement this.

PR #578, opened 2023-01-20
[python] Use `numpy.double` instead of the decrepeted `numpy.float` and `numpy.float64`

### LINKED ISSUE #576: Incompatibility with numpy v1.24
np.float is not supported anymore for v>=1.24

mreboud:
GitHub: resolves #576


SKuerten (approved):
I agree that these 3 lines are the only occurences of the deprecated `numpy.float` and `numpy.float64`.

dvandyk (approved):

PR #584, opened 2023-01-26
Fix bug in the LaTeX representations of some hadronic parameters

### LINKED ISSUE #583: LaTeX representations of parameters are not tested
@cbolognani ran into an issue with ``eos.tasks.corner_plot`` where a parameter used ``\GeV`` in its LaTeX representation. We should make sure that all parameters have valid representations as part of the test cases.

dvandyk:
GitHub: resolves #583

### DIFF
--- a/eos/parameters/hadron-properties.yaml
+++ b/eos/parameters/hadron-properties.yaml
@@ -627,42 +627,42 @@ groups:
           central: +0.16
           min:     +0.16
           max:     +0.16
-          latex:    '$a^{\pi}_2(1\,\GeV)$'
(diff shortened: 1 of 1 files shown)


mreboud:
Thank you @dvandyk. Your commit is however only half of the solution to #583. Can we find a way of test the latex representation of all the parameters?

dvandyk:
> Thank you @dvandyk. Your commit is however only half of the solution to #583. Can we find a way of test the latex representation of all the parameters?

I've removed the overzealous proposed resolution of #583. We should merge this as is and ASAP, anyway. I will think how to achieve the automatic testing in the mean while.

PR #586, opened 2023-01-29
Errors in eos.data.DynestyResults

fnovak42:
Several errors in the eos.data.DynestyResults object were fixed:
- Loading the data for initialization was not set up properly
- A typo in the parameter description was fixed
- dynesty was not imported

### DIFF
--- a/python/eos/data/native.py
+++ b/python/eos/data/native.py
@@ -18,6 +18,7 @@ import os
 import numpy as _np
 import pypmc
 import yaml
+import dynesty
 from scipy.special import erf
 from scipy.linalg import block_diag
 
@@ -526,7 +527,8 @@ class DynestyResults:
         f = os.path.join(path, 'dynesty_results.npy')
         if not os.path.exists(f) or not os.path.isfile(f):
             raise RuntimeError('Dynesty results file {} does not exist or is not a file'.format(f))
-        self.results = _np.load(f)
+        
+        self.results = dynesty.results.Results(_np.load(f, allow_pickle=True).item())
         self.samples = self.results.samples
         self.weights = _np.exp(self.results.logwt - self.results.logz[-1])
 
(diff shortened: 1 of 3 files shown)


python/eos/data/native.py
```diff
@@ -18,6 +18,7 @@
 import numpy as _np
 import pypmc
 import yaml
+import dynesty
```
mreboud, line 21: @dvandyk do we make `dynesty` an EOS dependence as we did for `pypmc`? Otherwise this needs to be moved to the corresponding function and protected.

python/eos/data/native.py
```diff
@@ -526,7 +527,10 @@ def __init__(self, path):
         f = os.path.join(path, 'dynesty_results.npy')
         if not os.path.exists(f) or not os.path.isfile(f):
             raise RuntimeError('Dynesty results file {} does not exist or is not a file'.format(f))
-        self.results = _np.load(f)
+        
+        load_res = _np.load(f, allow_pickle=True)
+        dict_res = load_res.item()
+        self.results = dynesty.results.Results(dict_res)
```
mreboud, line 533: Since the two commits deal with loading the `dynesty` results, merge them please.

mreboud (changes requested):
@fnovak42 Were you able to test it entirely? Including a check of the result on disk?

python/eos/data/native.py
dvandyk: @fnovak42 can you please adjust the tags for the two docker images in ``.github/workflows``, in the Ubuntu and Manylinux2014 worksflows? They should now point to ``c3b826a70d729805258243fb5d95dbe5ae2fbd2f``, which will be available in ~10min from now.

python/eos/data/native.py
mreboud: Yes

mreboud:
I also found a bug (?) in dynesty. When I run `sample_nested` from the CLI, there seems to be an entry `"blob"` created in the dynesty result. The result can't be imported due to this spurious entry. We should probably test its presence and pop it out from the result dictionary...

fnovak42:
> 

Rather than popping it out of the results dictionary, we could also set `blob=True` in the `eos.sample_nested`, since it might contain some useful information about the likelihood (or at least have the option)

mreboud:
> > 
> 
> Rather than popping it out of the results dictionary, we could also set `blob=True` in the `eos.sample_nested`, since it might contain some useful information about the likelihood (or at least have the option)

I don't think we need further information at this stage. We can always change that in the future.

dvandyk:
This will not cleanly merge. For some reason there's a merge commit in the history. The ``master`` branch does not allow merge commits.

PR #595, opened 2023-02-14
Add pseudo observables for missing BGL coefficients 

### LINKED ISSUE #590: Export pseudo-observable for F1 BGL form factor

dvandyk:
Some of the BGL coefficients in our implementation are non-parametric, due to exact relations amongst the form factors at given phase space points. This PR
 - turns the BGL form factor class into an ``ObservableProvider``;
 - exports the pseudo observables for the ``a_0`` coefficients;
 - fixes a bug in the naming of some pseudo observables for the coefficients of the HQE in B->D transitions.

Resolves #590.

### DIFF
--- a/eos/form-factors/parametric-bgl1997.hh
+++ b/eos/form-factors/parametric-bgl1997.hh
@@ -26,6 +26,7 @@
 #include <eos/form-factors/mesonic.hh>
 #include <eos/form-factors/mesonic-processes.hh>
 #include <eos/models/model.hh>
+#include <eos/utils/reference-name.hh>
 
 namespace eos
 {
@@ -63,6 +64,8 @@ namespace eos
 
             static std::string _par_name(const std::string & ff_name);
 
+            static const std::vector<OptionSpecification> _options;
+
         public:
             BGL1997FormFactors(const Parameters &, const Options &);
             ~BGL1997FormFactors();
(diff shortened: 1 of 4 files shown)


eos/form-factors/parametric-bgl1997.hh
```diff
@@ -113,6 +127,8 @@ namespace eos
 
             static std::string _par_name(const std::string & ff_name);
 
+            static const std::vector<OptionSpecification> _options;
```
mreboud, line 130: Why do you make the options private?

mreboud (commented):
I am a bit puzzled by the error you get, I don't understand what memory problem you encounter.

eos/form-factors/parametric-bgl1997.hh
dvandyk: Because I want to avoid accidentally using them outside the interface of an ``ObservableProvider``, i.e., the interface expected by ``ConcreteObservable``. I am a stout "defaults to private" guy!

dvandyk:
PR fails ``make check`` on Ubuntu/Jammy w/ ``g++``, which I cannot reproduce locally!

eos/form-factors/parametric-bgl1997.hh
mreboud: Ok, this should be extended to the other uses of `OptionSpecification` in this case :)

dvandyk:
> PR fails `make check` on Ubuntu/Jammy w/ `g++`, which I cannot reproduce locally!

I was able to reproduce it within a Docker container, seems to be related to the linker. Putting the static containers for the reference names into the implementation header might have caused this issue, by having to copies of them. Somehow, the destructor was called twice. Moving the container definition into the ``.cc`` file solved the issue.

eos/form-factors/parametric-bgl1997.hh
dvandyk: Off the top of my head: in all other cases, we have a non-private container of ``OptionSpecification`` within the Implementation class of the ``ObservableProvider``. That class is private by construction.

PR #597, opened 2023-02-14
Increase efficiency of the ``sample-nested`` tasks through prior sampling

dvandyk:
This PR makes a few changes to the code base.

 - [x] Each parameter receives an additional floating point value, the generator value, expected to be in ``[0, 1)``. This makes it possible to pass the generator value from Python to the C++ side (via ``evaluate_generator`` and ``set_generator``).
 - [x] The member function ``LogPrior::inverse_cdf`` is retired in lieu of ``LogPrior::sample``. Inverse transform sampling is still possible, and for a univariate prior it simply takes the form
 ```
# replaces
#   p = prior.inverse_cdf(u)
# for a univariate prior
p.set_generator(u)
prior.sample()
p = prior.evaluate()
 ```
 - [x] The member functions ``LogPrior::mean`` and ``LogPrior::variance`` are not used and cannot be adapted to the multivariate case; they are removed!
 - [x] The existing ``LogPrior::Gauss`` has been renamed to ``LogPrior::CurtailedGauss``, to reflect that the Gaussian tails are cut off outside a user-defined interval.
 - [x] A multivariate Gaussian log(prior) is added.
 - [x] Means to create a ``LogPrior`` from an existing ``ConstraintEntry`` are added for
   - [x] ``MultivariateGaussianCovarianceConstraintEntry`` (does not yet check that each "observable" is referring to a parameter!)
   - [x] ``MultivariateGaussianConstraintEntry``
   - [x] For the purpose of this PR, all other ``ConstraintEntry``s now throw an ``InternalError`` if their ``make_prior`` functions are called.
 - [x] Export the functions needed to create a ``LogPrior`` from a constraint
 - [x] Implement and export ``LogPrior::compute_cdf``, which provides access to the CDF for a univariate prior and the principal components' CDFs for a multivariate prior. The results are stored in the parameters' generator values. In short, this provides the inverse to ``LogPrior::sample``. Happy to discuss the naming.
 - [x] The ``eos.Analysis`` class is modified to exclusively use the parameters' u-space representation, i.e., to ``optimize``, ``sample_mcmc``, ``sample_pmc``, and ``sample_nested`` on the generator values / the cummulative prior probability.
 - [x] The internal variable ``eos.Analysis._bounds`` is removed. All bounds are implicity ``[0, 1)`` in u space.
 - [x] The ``eos.Analysis`` class is modified to optionally accept a single constraint as a prior. For example:
```
priors=[
  { 'constraint': 'foo->foo::bar@BAZ:2023A' },
  { 'parameter': 'mass::b(MSbar)', 'min': 4.15, 'max': 4.21, 'type': 'uniform' },
]
```
  - [x] ``eos.Analysis`` should check that there are no overlapping priors, either constraint vs constraint, or constraint vs univariate prior. I think this might already be done, through the ``LogPosterior`` class. If not, ``LogPosterior::add`` should do the checking, ideally by means of a ``std::set<QualifiedName>`` for all the parameter names. **Edit:** Indeed, this is done, see https://github.com/eos/eos/blob/e16907c243ee8bf62b77b98dbedde2ac49ce0fe7/eos/statistics/log-posterior.hh#L161
  - [x] The return value of ``bool LogPosterior::add(LogPriorPtr, bool)`` should be checked by ``eos.Analysis``, and a suitable Python error be thrown to indicate that the analysis description is defective.
  - [x] Create a test case that utilizes (ideally an existing) multivariate gaussian constraint to test that samples drawn using these improvements are correctly distributed. (@fnovak42) See e.g. [here](https://en.wikipedia.org/wiki/Normality_test) for possible tests.
    - [ ] ~Apply this branch to an existing real-world example, e.g. ``CKM-rho`` from eos/data#2023-01.~
    - [x] Check that the log(evidence) is correctly computer after this change. The existing unit test should provide an analytically calculable example.
    - [x] Check if the unit test's covariance matrix is sufficiently regular.

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -151,8 +151,19 @@ class Analysis:
             else:
                 raise ValueError('Prior specification must contains either a parameter or a constraint')
 
+        # check for duplicate entries in the likelihood
+        set_likelihood = set(likelihood)
+        if len(set_likelihood) != len(likelihood):
+            raise ValueError(f'The likelihood contains duplicate entries')
+        set_manual_constraints = set(manual_constraints.keys())
+        if len(set_manual_constraints) != len(manual_constraints):
+            raise ValueError(f'The manual constraints contain duplicate entries')
+        set_intersection = set_likelihood.intersection(set_manual_constraints)
+        if len(set_intersection) > 0:
+            raise ValueError(f'The likelihood contains constraint names also present in the manual_constraints: {set_intersection}')
+
         # record all constraints that comprise the likelihood
-        self._constraint_names = list(likelihood) + list(manual_constraints.keys())
+        self._constraint_names = list(set_likelihood.union(set_manual_constraints))
 
         # create the likelihood
         for constraint_name in self._constraint_names:
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -80,7 +80,10 @@ class Analysis:
             nprior=len(priors), nconst=len(likelihood), nopts=len(global_options), nmanual=len(manual_constraints), nparams=len(fixed_parameters)))
         eos.debug('priors:')
         for p in priors:
-            eos.debug(' - {name} ({type}) [{min}, {max}]'.format(name=p['parameter'], type=p['type'], min=p['min'], max=p['max']))
+            if 'parameter' in p:
+                eos.debug(' - {name} ({type}) [{min}, {max}]'.format(name=p['parameter'], type=p['type'], min=p['min'], max=p['max']))
+            elif 'constraint' in p:
+                eos.debug(' - {name} (constraint)'.format(name=p['constraint']))
         eos.debug('constraints:')
         for cn in likelihood:
             eos.debug(' - {name}'.format(name=cn))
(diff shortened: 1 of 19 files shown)


dvandyk:
One of the problems identified by the tests is as follows:
 - When sampling from a (multivariate) gaussian prior, we would effectively have bounds of ``[-inf, +inf]``, which is incompatible with the current treatment of parameters in the MCMC & PMC sampling and for mode finding. To be discussed when @mreboud is back from vacation.

dvandyk:
The [current version](https://github.com/eos/eos/commit/46803ba3a93d1752ea35673afbaba1048faed38f) of this PR fixed a bug in the sampling from a multivariate gaussian prior.

The last few commits should be reordered and merged to yield
  - [python] Switch ``eos.Analysis`` to work exclusively in u space
  - [python] Support using a known constraint as a multivariate prior

mreboud:
> The [current version](https://github.com/eos/eos/commit/46803ba3a93d1752ea35673afbaba1048faed38f) of this PR fixed a bug in the sampling from a multivariate gaussian prior.
> 
> The last few commits should be reordered and merged to yield
> 
>     * [python] Switch `eos.Analysis` to work exclusively in u space
> 
>     * [python] Support using a known constraint as a multivariate prior

I will do this, as well as correcting the `analysis_TEST.py` file.

dvandyk:
All ticks checked. Branch passes tests locally. Ready for review @mreboud @fnovak42

python/eos/data/native.py
```diff
@@ -115,15 +120,17 @@ def __init__(self, path):
...
-    def create(path, parameters, samples, weights=None):
+    def create(path, parameters, samples, usamples, weights=None):
         """ Write a new MarkovChain object to disk.
 
         :param path: Path to the storage location, which will be created as a directory.
         :type path: str
         :param parameters: Parameter descriptions as a 1D array of shape (N, ).
         :type parameters: list or iterable of eos.Parameter
-        :param samples: Samples as a 2D array of shape (N, P).
+        :param samples: Samples in parameter sppace as a 2D array of shape (N, P).
```
dvandyk, line 130: Typo.

python/eos/data/native.py
```diff
@@ -140,13 +147,17 @@ def create(path, parameters, samples, weights=None):
         if not samples.shape[1] == len(parameters):
             raise RuntimeError('Shape of samples {} incompatible with number of parameters {}'.format(samples.shape, len(parameters)))
 
+        if not usamples.shape[1] == len(parameters):
+            raise RuntimeError('Shape of zsamples {} incompatible with number of parameters {}'.format(samples.shape, len(parameters)))
```
dvandyk, line 151: typo: ``zsamples`` -> ``usamples
bug: ``format(samples.shape, ...)`` -> ``format(usamples.shape, ...)``

python/eos/analysis.py
```diff
@@ -80,7 +80,10 @@ def __init__(self, priors, likelihood, global_options={}, manual_constraints={},
             nprior=len(priors), nconst=len(likelihood), nopts=len(global_options), nmanual=len(manual_constraints), nparams=len(fixed_parameters)))
         eos.debug('priors:')
         for p in priors:
-            eos.debug(' - {name} ({type}) [{min}, {max}]'.format(name=p['parameter'], type=p['type'], min=p['min'], max=p['max']))
+            if 'parameter' in p:
+                eos.debug(' - {name} ({type}) [{min}, {max}]'.format(name=p['parameter'], type=p['type'], min=p['min'], max=p['max']))
+            elif 'constraint' in p:
+                pass
```
dvandyk, line 86: Output the name of the constraint being added.

dvandyk (commented):
Comments based on joint review with @mreboud

python/eos/analysis_TEST.py
```diff
@@ -111,5 +112,118 @@ def test_sanitize_manual_input(self):
...
+        cov    = np.array(entry_data['covariance'])
+        sigmas = np.diag(cov)
+
+        means  = np.append(means,  [1.27, 4.18])
+        sigmas = np.append(sigmas, [0.03, 0.02])
+
+        avg    = np.median(results.samples_equal(), axis=0)
+        delta  = avg - means
+        chi2_1 = np.dot(delta[0:8], np.dot(np.linalg.inv(cov), delta[0:8]))
+        chi2_2 = (avg[8] - means[8])**2 / 0.03**2
```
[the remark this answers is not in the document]
dvandyk: I have also moved the analytic result for log(Z) to a variable in both test cases.

PR #598, opened 2023-02-15
Fix 'fixed_parameters' initialization for predictions

fnovak42:
The syntax error in the loop initialization in `analysis_file.py` was fixed so that the parameters can be specified. A test was also added to `analysis.yaml` to check the modification.

### DIFF
--- a/src/scripts/eos-analysis_TEST.d/analysis.yaml
+++ b/src/scripts/eos-analysis_TEST.d/analysis.yaml
@@ -75,6 +75,8 @@ predictions:
     global_options:
         model: CKM
         form-factors: BCL2008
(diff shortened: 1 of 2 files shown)


src/scripts/eos-analysis_TEST.d/analysis.yaml
```diff
@@ -75,6 +75,8 @@ predictions:
     global_options:
         model: CKM
         form-factors: BCL2008
+    fixed_parameters:
+        mass::e: 5e-4  # This is meant to test the "fixed_parameters" key
```
mreboud, line 79: I think that `5e-4` was not parsed, hence your error

PR #600, opened 2023-02-16
Painfully fix the `begin` and `end` options

### LINKED ISSUE #596: Limiting the range of a multivariate gaussian constraints is not supported for those with correlation matrix

mreboud:
At some point we should implement response matrices in another constraint type.

Fixes #596

### DIFF
--- a/eos/constraint.cc
+++ b/eos/constraint.cc
@@ -860,6 +860,9 @@ namespace eos
             gsl_matrix * response = gsl_matrix_calloc(dim, dim);
             gsl_matrix_set_identity(response);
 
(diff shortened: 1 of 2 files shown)


dvandyk (approved):
Looks good. Test case is more than adequate. Thank you very much for the OCD-ish fixed to the comments, too!

dvandyk:
FYI: This seems to work in the application of branch ``issue238``. Thank you @mreboud !

mreboud:
I am not to happy with the fix though... It may require some refactoring at some point!

PR #604, opened 2023-02-26
Spring cleaning

dvandyk:
This PR collects most changes introduced by running `pre-commit run -a`. Exceptions are:
 - two issues in a YAML file that appear to be valid YAML
 - issues in a Jupyter notebook in `numerics/`

In the process, I delete the old attempt at a manual in `manual/`.

We should probably make using `pre-commit` mandatory. Any suggestions how to best do that?


mreboud:
Thanks @dvandyk!
I would add it in the workflow, as a first check of the code.

mreboud (approved):
All of this looks good. The fact that the tests pass ensures the correctness of the changes I think :)

mreboud:
If we make this check mandatory, we will have to fix the few remaining issues however...

PR #605, opened 2023-02-26
Position space pseudo observable for ``b_lcdas::FLvD2022``

pluegh:
The proposed changes allow to use the position space LCDA of the B meson $\tilde\phi_{B,+}(\tau)$ and its derivative as a constraint for the parametrisation introduced in `[FLvD:2022A]`. Such information can be extracted from theory through an operator product expansion.

### DIFF
--- a/eos/form-factors/b-lcdas-flvd2022.cc
+++ b/eos/form-factors/b-lcdas-flvd2022.cc
@@ -36,8 +36,8 @@ namespace eos
     namespace b_lcdas
     {
         FLvD2022::FLvD2022(const Parameters & p, const Options & o) :
-            opt_q(o, "q", { "u", "s" }, "u"),
-            opt_gminus(o, "gminus", { "zero", "WW-limit" }, "WW-limit"),
+            opt_q(o, options, "q"),
+            opt_gminus(o, options, "gminus"),
             switch_gminus(1.0),
             mu_0(p[parameter("mu_0")], *this),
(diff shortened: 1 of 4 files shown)


eos/form-factors/b-lcdas-flvd2022.cc
```diff
@@ -117,6 +117,11 @@ namespace eos
             throw InternalError("Function not yet implemented");
         }
 
+        double
+        FLvD2022::phitilde_plus(const double & tau) const
+        {
+            return 0.0;
```
dvandyk: I assume that you still need to squash/fixup the last commit into this one?

eos/form-factors/b-lcdas-flvd2022.cc
```diff
@@ -315,7 +320,7 @@ namespace eos
         FLvD2022::options
         {
             { "q",       { "u", "s" },           "u"        },
-            { "g-minus", { "zero", "WW-limit" }, "WW-limit" }
+            { "gminus", { "zero", "WW-limit" }, "WW-limit" }
```
dvandyk, line 378: Is this compatibl with how the options are used in Nico's implementation of the light-cone sum rules?
E.g. ``eos/form-factors/analytic-b-to-p-lcsr-impl.hh``, I think.

dvandyk (changes requested):
- [x] Squash the last two commits into one.
 - [x] Rebase and fix merge conflicts.

Issue #573, opened 2023-01-11
Some additional features for nested sampling-based analyses


mreboud:
- [x] Add a seed to nested sampling, check whether specifying it is implemented in `dynesty`.
- [ ] Find a way of recording the nested sampling settings in the analysis file.
- [x] The `sample_nested` task output should also be fixed. E.g. the `nested` directory is not used anymore and the log file is usually empty
- [x] We should save the state of the sampling after every Nth steps, probably by replacing the existing `run` loop that dynesty provides by our own loop.
- [x] Make sure that the `log` file contains relevant information on the status of the sampling. It should e.g. be clear whether the tasks finished or not.
- [x] Give the possibility to require a minimal number of samples. 

In the long term:
- [ ] Devise a way of interfacing an EOS/pypmc mixture with external software, e.g. `Wilson`.

fnovak42:
- [ ] Rename the `eos.data.ImportanceSamples` to something like "dynesty_samples," not just "samples"

fnovak42:
The problem that occurs when using `rstate=np.random.mtrand.RandomState(SEED)` as an input in the dynesty sampler is `AttributeError: 'numpy.random.mtrand.RandomState' object has no attribute 'integers'` This is not an issue when using e.g. `rstate = np.random.default_rng(SEED)` and not noticeable when doing the same thing with `dynesty.utils.resample_equal`.

mreboud:
Can you investigate a bit further?

fnovak42:
I think the best way to solve this is to specify `rstate = np.random.Generator(np.random.MT19937(seed))` in https://github.com/eos/eos/blob/982f9b2ec255619b7b5247508a67819eb2df9f9c/python/eos/analysis.py#L618

mreboud:
> I think the best way to solve this is to specify `rstate = np.random.Generator(np.random.MT19937(seed))` in...

Yes it seems that dynesty's rng is called `rstate` but they expect a `Generator` while pypmc rng is called `rng` and expect a random state... Your solution looks fine to me!

Can you create a branch issue573 on eos/eos and implement this first commit?

fnovak42:
Regarding points 3 and 5, the log file can be moved to `dynesty_results` with just a simple message such as "A dlogz of ... was reached which falls below the threshold initially specified" or "the sampling was stopped since the maximum number of iterations was reached" for now. Later this can extended using information from point 4. Do you agree or would you add more information to the log file?

dvandyk:
[...]

PR #606, opened 2023-02-28
Add seed to nested sampling

fnovak42:
As discussed in issue #573, a seed was added to the rng used by the `DynamicNestedSampler`

### DIFF
--- a/src/scripts/eos-analysis
+++ b/src/scripts/eos-analysis
@@ -249,6 +249,10 @@ The output file will be stored in EOS_BASE_DIRECTORY/POSTERIOR/nested.
         help = 'The maximum number of iterations. Iterations may stop earlier if the termination condition is reached.',
         dest = 'maxiter', action = 'store', type = int, default = None
     )
+    parser_sample_nested.add_argument('-s', '--use-random-seed',
+        help = 'The seed used to initialize the Mersenne Twister pseudo-random number generator.',
+        dest = 'seed', action = 'store', type = int, default = 10
+    )
     parser_sample_nested.add_argument('-b', '--base-directory',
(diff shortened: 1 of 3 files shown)


dvandyk (changes requested):
This looks reasonable, thanks! However, the change should also be reflected in `eos-analysis`, where the same default value for the seed should be used as in the task.

fnovak42:
For simplicity, I've added just the `int` option in the task and parser specifications for the seed. Technically, we could also use an array or `None`, but I'm not sure if this is practically relevant.

src/scripts/eos-analysis
```diff
@@ -249,6 +249,10 @@ The output file will be stored in EOS_BASE_DIRECTORY/POSTERIOR/nested.
         help = 'The maximum number of iterations. Iterations may stop earlier if the termination condition is reached.',
         dest = 'maxiter', action = 'store', type = int, default = None
     )
+    parser_sample_nested.add_argument('-s', '--rng-seed',
```
mreboud, line 252: Can you use the command `--use-random-seed` for consistency with the `find-mode` task?

mreboud (changes requested):
Have you been able to check that it works as expected? I would run 3 nested samplings: 2 with the default seed and a with another one. Then you can check that the events in the two first samplings are the same and don't match the ones in the third sampling.

fnovak42:
> Have you been able to check that it works as expected? I would run 3 nested samplings: 2 with the default seed and a with another one. Then you can check that the events in the two first samplings are the same and don't match the ones in the third sampling.

All right, that sound like a better way to check it, I will try that. I just ran two samplings, one with the default seed and one with a different one.

dvandyk (approved):
Looks good now, thank you @fnovak42!

PR #608, opened 2023-02-28
Add example notebook on organising an analysis

fnovak42:
Added the basic structure - the notebook and shortened YAML file - for the nested sampling documentation. This has to be further extended with the relevant `eos/doc` entries.


dvandyk:
Merging this PR as is. It adds information to the documentation that is currently not available. Will need to revisit the example notebook at some later point.

PR #617, opened 2023-03-15
[python] Add feature to enable plotting on given axis.

lorenzennio:
Provide users with the option to plot on a given axis:
```
fig, ax = plt.subplots(1,1)
plot_args = {
                'axis': ax,
                'plot': {
                   ...
                },
                'contents': [
                   {
                       ...
                   }
                ]
           }
eos.plot.Plotter(plot_args).plot()
```

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -99,15 +99,19 @@ class Plotter:
         if not 'plot' in self.instructions:
             raise KeyError('no plot metadata specified')
 
-        myplot = self.instructions['plot']
+        if 'axis' in self.instructions:
+            self.ax = self.instructions['axis']
+        else:
+            self.fig, self.ax = plt.subplots()
 
-        self.fig, self.ax = plt.subplots()
+        myplot = self.instructions['plot']
 
         mytitle = ''
         myylabel = ''
         myxlabel = ''
         myyscale = 'linear'
         myxscale = 'linear'
+
         if 'title' in myplot:
             mytitle = myplot['title']
 
@@ -279,7 +283,7 @@ class Plotter:
             self.markersize = item['markersize'] if 'markersize' in item else None
 
         def plot(self):
-            plt.plot(self.x, self.y,
+            self.plotter.ax.plot(self.x, self.y,
                 alpha=self.alpha, color=self.color,
                 label=self.label, linestyle='None',
                 marker=self.marker, markeredgecolor=self.color, markerfacecolor='None',
@@ -304,7 +308,7 @@ class Plotter:
             self.elinewidth = item['elinewidth'] if 'elinewidth' in item else 1.0
 
         def plot(self):
-            plt.errorbar(x=self.x, y=self.y, xerr=self.xerr, yerr=self.yerr,
+            self.plotter.ax.errorbar(x=self.x, y=self.y, xerr=self.xerr, yerr=self.yerr,
                 color=self.color, elinewidth=self.elinewidth, fmt='_', linestyle='none', label=self.label)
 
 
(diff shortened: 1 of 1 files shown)


lorenzennio:
I get the warning 
```
WARNING:EOS:   both data and data-file specified; assuming interactive use is intended
```
whenever I do a kde2d plot. Is this intended?

mreboud:
> I get the warning
> 
> ```
> WARNING:EOS:   both data and data-file specified; assuming interactive use is intended
> ```
> 
> whenever I do a kde2d plot. Is this intended?

Can you show us your plot arguments?

lorenzennio:
> > I get the warning
> > ```
> > WARNING:EOS:   both data and data-file specified; assuming interactive use is intended
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > whenever I do a kde2d plot. Is this intended?
> 
> Can you show us your plot arguments?

Sure:
```
plot_args = {
                'plot': {
                   'x': { 'label': labels[0], 'range': [-30,30] },
                   'y': { 'label': labels[1], 'range': [-30,30] },
                },
                'contents': [
                   {
                       'type': 'kde2D', 'color': 'C1', 'label': 'posterior',
                       'levels': [68, 95], 'contours': ['lines'], 'bandwidth':3,
                       'data': { 'samples': wilsons[:, (0,1)], 'weights':1/nll_norm }
                   }
                ]
           }
```

mreboud:
So `data-file` is absolutely not present, correct?

lorenzennio:
No, not that I know of.

dvandyk:
> Provide users with the option to plot on a given axis:
> 
> ```
> fig, ax = plt.subplots(1,1)
> plot_args = {
>                 'plot': {
>                    'axis': ax,
>                    ...
>                 },
>                 'contents': [
>                    {
>                        ...
>                    }
>                 ]
>            }
> eos.plot.Plotter(plot_args).plot()
> ```

In the interest of future-proofing this change: can we slightly change the code so that ``axis`` becomes another top-level keyword argument (i.e., same level as ``plot`` and ``contents``)?

lorenzennio:
No problem, I will make the corresponding changes.

lorenzennio:
> In the interest of future-proofing this change: can we slightly change the code so that `axis` becomes another top-level keyword argument (i.e., same level as `plot` and `contents`)?
I have changed that and the corresponding comment above.

python/eos/plot/plotter.py
```diff
@@ -99,15 +99,19 @@ def setup_plot(self):
         if not 'plot' in self.instructions:
             raise KeyError('no plot metadata specified')
 
-        myplot = self.instructions['plot']
+        if 'axis' in self.instructions:
+            self.ax = self.instructions['axis']
```
dvandyk, line 103: This leaves ``self.fig`` uninitialized. Can you at least set it to ``None``?

dvandyk (approved):
Conditional approval. If ``Plotter.fig`` is not used anywhere, this is OK to rebase+merge.
Otherwhise, see me comment to set ``fig`` to ``None``.

python/eos/plot/plotter.py
lorenzennio: This should be done here:

https://github.com/eos/eos/pull/617/files#diff-5ac706e3ba490789c67c7e29957bb047c2ba97e7f97d8e316636331751102b01R42

dvandyk: You are absolutely right! I wave my concerns. I think it can be merged now, if all of @mreboud's concerns are removed, too!

PR #624, opened 2023-03-30
Support observables from expressions in analysis file

pluegh:
The small set of changes allows to use the expression parser to include custom observables in an analysis file.


pluegh:
> I approve of the existing code. However, since this adds a side condition to loading an analysis file, I would prefer `eos.AnalysisFile` informed the user of the added observables via `eos.info`.

Thanks for the review. I updated the code with an info log message.

dvandyk (approved):

PR #627, opened 2023-04-14
Implementation of BFW parameterization for b -> sll form factors

mreboud:
I also corrected a few typos here and there :)


eos/form-factors/parametric-bfw2010-impl.hh
```diff
@@ -224,6 +222,38 @@ namespace eos
         return a_T2_0 / (polynomials_T2[0] * x_T1);
     }
 
+    template <typename Process_>
+    double
+    BFW2010FormFactors<Process_, PToV>::_a_A1_0() const
```
dvandyk, line 227: Works. Personally, I would have replaced a_A12_0, but that's not a reason to reject this!

eos/form-factors/parametric-bfw2010-impl.hh
```diff
@@ -451,47 +451,129 @@ namespace eos
...
-        return std::inner_product(_a_A0.begin(), _a_A0.end(), _a_A0.begin(), 0.0);
+        return 0.0;
     }
 
     template<typename Process_>
     double
-    BFW2010FormFactors<Process_, PToV>::saturation_1p_a() const
+    BFW2010FormFactors<Process_, PToV>::saturation_1m_v_perp() const
+    {
+        return 3.0 * std::inner_product(_a_V.begin(), _a_V.end(), _a_V.begin(), 0.0);
```
dvandyk, line 470: The factor of three is convention here, is a bit weird and requires a comment I think.

eos/form-factors/parametric-bfw2010-impl.hh
```diff
@@ -451,47 +451,129 @@ namespace eos
...
+    BFW2010FormFactors<Process_, PToV>::saturation_1m_v_para() const
+    {
+        return 0.0;
+    }
+
+    template<typename Process_>
+    double
+    BFW2010FormFactors<Process_, PToV>::saturation_1m_v() const
+    {
+        return (saturation_1m_v_0() + saturation_1m_v_perp() + saturation_1m_v_para()) / 3.0;
```
dvandyk, line 484: Ditto here.

python/eos/analysis_file.py
```diff
@@ -121,21 +121,28 @@ def analysis(self, _posterior):
                             fixed_parameters=fixed_parameters)
 
 
-    def observables(self, _prediction, parameters):
-        """Creates a list of eos.Observable objects for the named set of predictions."""
+    def observables(self, _posterior, _prediction, parameters):
```
dvandyk, line 124: Good idea, thanks!

dvandyk (approved):
Approved. However, see my few comments.

eos/form-factors/parametric-bfw2010-impl.hh
mreboud: a_A12_0 is already replaced by the relation at q2 = 0 ;)

mreboud:
I removed the factors of 3 to follow the notations in the paper.

PR #630, opened 2023-05-05
[python] Add LogPrior 'scale' to Analysis

pluegh:


dvandyk (commented):
Hm, can you maybe add a test case for using an ``eos.Analysis`` with a ``type: 'scale'`` prior?

pluegh:
> Hm, can you maybe add a test case for using an `eos.Analysis` with a `type: 'scale'` prior?

What extent of a test would be reasonable? Does a test case that constructs an analysis with all possible prior types suffice or could/should it be more elaborate?

PR #632, opened 2023-05-12
Export the charm loop functions to python

mreboud:


eos/rare-b-decays/charm-loops.cc
```diff
@@ -1533,6 +1533,64 @@ namespace eos
...
+        return F27_Qsb(clp);
+    }
+
+    complex<double> F29_Qsb(const complex<double> & s, const double & mu, const double & m_c, const double & m_b)
+    {
+        const agv_2019a::CharmLoopsParameters clp(mu / m_b, s, (m_c * m_c) / (m_b * m_b), 1e-12);
+
+        return F29_Qsb(clp);
+    }
+
```
dvandyk, line 1593: Hrm. Sorry to delay this, but the indentation seems to be incorrect. Shouldn't the helper functions be indented by one further level?

python/_eos.cc
```diff
@@ -864,14 +864,23 @@ BOOST_PYTHON_MODULE(_eos)
         .def("__iter__", range(&SignalPDFs::begin, &SignalPDFs::end))
         ;
 
-    // }}}
+    // Analytic Charm Loops
+    def("F17_Qc",  (complex<double>(*)(const complex<double> &, const double &, const double &, const double &)) &agv_2019a::F17_Qc);
```
dvandyk, line 868: OK as is for Javi's use. Before the next release, we should make sure that these free-standing functions are put into a dedicated class as static methods.

dvandyk (approved):

python/_eos.cc
mreboud: My mistake!

eos/rare-b-decays/charm-loops.cc
mreboud: My mistake!

mreboud:
@dvandyk should I merge?

PR #634, opened 2023-05-19
Fix a bug in B -> V l nu decays

mreboud:


eos/b-decays/b-to-vec-l-nu.cc
```diff
@@ -237,16 +237,16 @@ namespace eos
             if ((q2 >= power_of<2>(m_l)) && (q2 <= power_of<2>(m_B - m_V))) {
                 result.a_0          = isospin * gV_mi * 8.0 * m_B * m_V / sqrtq2 * aff12;
                 result.a_0_T        = isospin * TL / (2.0 * m_V) * ( (m_B * m_B + 3.0 * m_V * m_V - q2) * tff2 - lam * tff3 / (m_B * m_B - m_V * m_V) );
-                result.a_plus       = isospin * (m_B + m_V) * aff1 * gV_mi - sqrt_lam * vff * gV_pl / (m_B + m_V);
-                result.a_minus      = isospin * (m_B + m_V) * aff1 * gV_mi + sqrt_lam * vff * gV_pl / (m_B + m_V);
+                result.a_plus       = isospin * ( (m_B + m_V) * aff1 * gV_mi - sqrt_lam * vff * gV_pl / (m_B + m_V) );
```
dvandyk, line 240: This should be instrumented with a test case.

eos/b-decays/Makefile.am
```diff
@@ -70,6 +70,7 @@ TESTS = \
 	b-to-kstar-nu-nu_TEST \
 	b-to-pi-l-nu_TEST \
 	b-to-pi-pi-l-nu_TEST \
+	b-to-rho-l-nu_TEST \
```
dvandyk, line 73: Uff. The test case was not live?

dvandyk (approved):
Should be merged ASAP.

eos/b-decays/Makefile.am
dvandyk: Ahh, that makes sense!

PR #636, opened 2023-05-23
Throw an error if a form factor or LCDA factory fails to make its target

### LINKED ISSUE #616: Throw an error if a form factor or LCDA factory fails to make its target.
Should we make an explicit error if the key is not found?

_Originally posted by @mreboud in https://github.com/eos/eos/pull/614#discussion_r1136697814_

dvandyk:
Resolves #616.

### DIFF
--- a/eos/b-decays/b-to-vec-nu-nu.cc
+++ b/eos/b-decays/b-to-vec-nu-nu.cc
@@ -38,8 +38,6 @@ namespace eos
     {
         std::shared_ptr<Model> model;
 
-        std::shared_ptr<FormFactors<PToV>> form_factors;
-
         Parameters parameters;
 
         QuarkFlavorOption opt_D;
@@ -70,6 +68,8 @@ namespace eos
         GSL::QAGS::Config int_config;
 
         bool cp_conjugate;
+
+        std::shared_ptr<FormFactors<PToV>> form_factors;
 
         // { D, q, I } -> { process, m_B, m_V, c_I }
         // D: s, d; the quark flavor in the weak transition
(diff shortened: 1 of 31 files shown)


eos/b-decays/b-to-vec-nu-nu.cc
```diff
@@ -217,7 +213,7 @@ namespace eos
     Implementation<BToVectorDineutrino>::options
     {
         Model::option_specification(),
-        FormFactorFactory<PToP>::option_specification(),
+        FormFactorFactory<PToV>::option_specification(),
```
mreboud, line 216: Good catch, thanks :)

eos/form-factors/b-lcdas.cc
```diff
@@ -43,12 +45,10 @@ namespace eos
...
         {
             result.reset( i->second(parameters, options) );
-        }
-        else
-        {
-            throw InternalError("Unknown value for 'name' for BMesonLCDAs::make");
+            return result;
         }
 
+        throw InternalError("Unknown B-meson LCDA model: " + name);
```
mreboud, line 51: Were you able to check that?

eos/form-factors/mesonic.cc
```diff
@@ -485,14 +510,18 @@ namespace eos
     std::shared_ptr<FormFactors<VToV>>
     FormFactorFactory<VToV>::create(const QualifiedName & name, const Parameters & parameters, const Options & options)
     {
+        Context ctx("When creating a V->P form factor");
```
mreboud, line 513: This should be V -> V I think

mreboud (changes requested):
I am mostly happy with the changes. There is one typo, and I would like to see a test for this :)

eos/form-factors/mesonic.cc
dvandyk: Argh. I noticed it in one place, but overlooked the context. Fixing it right now.

eos/form-factors/b-lcdas.cc
dvandyk: After revising: yes. All form factors and LCDAs have been instrumented with a test case ensuring that an exception is raised for nonsense inputs.

mreboud: Excellent, thanks a lot!

PR #638, opened 2023-06-01
Implement charm loops as observables

mreboud:


eos/rare-b-decays/charm-loops.cc
```diff
@@ -1476,6 +1479,195 @@ namespace eos
         return result;
     }
 
+
+    /*
+     * Adaptater that exports the charn loops function as observables
```
dvandyk, line 1484: "Adaptor"

eos/rare-b-decays/charm-loops.cc
```diff
@@ -1476,6 +1479,195 @@ namespace eos
...
+     */
+    template <>
+    struct Implementation<CharmLoopsAdaptater>
```
dvandyk, line 1487: Adaptor :smile:

eos/rare-b-decays/charm-loops.cc
```diff
@@ -1476,6 +1479,195 @@ namespace eos
...
+                 flag_ctQs = 0,
+                 flag_ctQb = 0;
+
+        Implementation(const Parameters & p, const Options & o, ParameterUser & u) :
+            opt_contribution(o, options, "contribution"),
+            m_b(p["mass::b(MSbar)"], u),
+            m_c(p["mass::c"], u),
+            mu(p["sb::mu"], u)
+        {
+            if ("0" == opt_contribution.value())
```
dvandyk, line 1513: This is better done with an ``std::map`` and ``std::tie``, isn't it?

```
static const std::map<std::string, std::tuple<double, double, double>> settings
{
    std::make_pair("all", std::make_tuple(1.0, 1.0, 1.0)),
    ...
};

auto i = settings.find(opt_contribution.value());
if (i == settings.end())
   throw ...;
 
// member variables
double flag_a, flag_b, ...;

std::tie(flag_a, flag_b, flac_) = i->second;
```

eos/rare-b-decays/charm-loops.cc
```diff
@@ -1476,6 +1479,195 @@ namespace eos
...
+
+        inline agv_2019a::CharmLoopsParameters clp(const complex<double> & s) const
+        {
+            return agv_2019a::CharmLoopsParameters(mu / m_b, s / m_b() / m_b(), (m_c * m_c) / (m_b * m_b), 1e-12);
+        }
+
+        complex<double> F17(const complex<double> & s) const
+        {
+            auto params = clp(s);
+            return (double)flag_a * agv_2019a::f17a(params)
```
dvandyk, line 1541: I would declare the flags as ``double`` from the get go, to avoid the casting here.

dvandyk (changes requested):

eos/rare-b-decays/charm-loops.cc
mreboud: Adapter? Like in form-factor-adapter?

PR #639, opened 2023-06-03
[doc] Provide explicit table of common pip3 commands

pluegh:
The proposed changes list the `pip3` commands for installation and upgrade explicitly.
The user can find and copy the right command at a glance, which also better communicates the ease of this installation method.

Please find a preview attached.

PR_preview

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -12,18 +12,30 @@ Linux
 Installation with `pip3`
 ========================
 
-This is the recommended installation method using the 'package installer for Python' `pip3`:
+This is the recommended installation method using the 'package installer for Python' `pip3`.
+We distribute a release version and a development version.
+The changelog of the curated release versions is available `here <https://eos.github.io>`_.
+The development version is automatically up-to-date with the master branch in the Github EOS repository.
+Here we summarize the most common commands using `pip3`:
 
-::
+.. list-table::
+   :widths: 1 1
+   :header-rows: 1
 
-  pip3 install --user eoshep
+   * -
+     - Command
+   * - Install latest release version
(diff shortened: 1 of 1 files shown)


doc/installation.rst
```diff
@@ -12,18 +12,30 @@ Linux
...
-  pip3 install --user eoshep
-
-You can now use the EOS Python module via ``import eos``. In addition, some command-line scripts have become available.
-
-| To **update** EOS, add the flag ``--upgrade`` to the above command.
-| To install the **development version**, add the flag ``--pre``.
-  This version is automatically up-to-date with the master branch in the Github EOS repository.
-| You can **test** and debug the installation of the python module as described :ref:`here <faq-check-installation>`.
+This is the recommended installation method using the 'package installer for Python' `pip3`.
+We distribute a release version and a developer version.
```
pluegh, line 16: It should say *development version* instead, since the distributed versions are not for developers but for users.

doc/installation.rst
```diff
@@ -12,18 +12,30 @@ Linux
...
+The changelog of the curated release versions is available `here <https://eos.github.io>`_.
+The developer version is automatically up-to-date with the master branch in the Github EOS repository.
```
pluegh, line 18: Same here.

pluegh (commented):
Strictly speaking, there is a small typo that should be fixed.

doc/installation.rst
mreboud: I agree

doc/installation.rst
mreboud: Yes

mreboud (approved):
That's very nice :)

PR #642, opened 2023-06-06
Add EOS-specific pre-commit hooks

dvandyk:
Hooks are kept in ``eos/pre-commit-hooks``.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,8 @@ repos:
     - id: rst-backticks
     - id: rst-directive-colons
     - id: rst-inline-touching-normal
+
+ - repo: https://github.com/eos/pre-commit-hooks
+   rev: v1.0.1
+   hooks:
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -23,3 +23,8 @@ repos:
     - id: rst-backticks
     - id: rst-directive-colons
     - id: rst-inline-touching-normal
+
+ - repo: https://github.com/eos/pre-commit-hooks
+   rev: v1.0.0
+   hooks:
+    - id: eos-test-check-argument-order
```
mreboud, line 30: Can you extend the scope of your hook to ``TEST_CHECK_RELATIVE_ERROR`` and ``TEST_CHECK_RELATIVE_ERROR_C``?

dvandyk: Yes, let's move that over to ``eos/pre-commit-hooks``.

PR #644, opened 2023-06-07
Elaborate NS run using a loop

fnovak42:
Use a loop to run a nested sampler internally as outlined here: https://dynesty.readthedocs.io/en/stable/dynamic.html
This is related to one or several points from issue #573.

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -700,7 +700,7 @@ class Analysis:
         return self._u_to_par(u)
 
 
-    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=10, print_progress=True, sample='auto'):
+    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=10, print_progress=True, sample='auto', check_point_interval=None):
         """
         Return samples of the parameters.
 
@@ -718,14 +718,53 @@ class Analysis:
         :type seed: {None, int, array_like[ints], SeedSequence}, optional
         :param sample: The method used for sampling within the likelihood constraints. For valid values, see dynesty documentation. Defaults to 'auto'.
         :type sample: str, optional
+        :param checkpoint_interval: The number of seconds between checkpoints at which the intermediate dynesty sampler results are stored. If None, then no intermediate results are stored. Defaults to 60.
+        :type checkpoint_interval: float, optional
 
         .. note::
            This method requires the dynesty python module, which can be installed from PyPI.
         """
         import dynesty
+        import itertools
+        from dynesty.dynamicsampler import stopping_function, weight_function
         sampler = dynesty.DynamicNestedSampler(self.log_likelihood, self._prior_transform, len(self.varied_parameters), bound=bound, nlive=nlive, rstate = np.random.Generator(np.random.MT19937(seed)), sample=sample)
-        sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress)
-        return sampler.results
+
+        eos.info('Drawing initial samples')
+        for results in sampler.sample_initial(dlogz=dlogz, maxiter=maxiter):
+            pass
+
+        eos.inprogress('Starting nested sampling loop')
+        if check_point_interval is not None:
+            timer = dynesty.utils.DelayTimer(check_point_interval)
+            time = 0
+
+        for iter in itertools.count():
+            if check_point_interval is not None and timer.is_time():
+                time += check_point_interval
+                eos.info(f'Created check-point after {time} seconds')
+                yield sampler.results
+
+            stop = stopping_function(sampler.results)
+
+            if maxiter is not None and iter > maxiter:
+                stop = True
+                eos.sucess("The sampling was stopped because the maximum number of iterations was reached")
+
+            delta_logz = np.logaddexp(0, np.max(sampler.live_logl) + sampler.results.logvol - sampler.results.logz)[-1]
+            if delta_logz < dlogz:
+                stop = True
+                eos.success("The sampling was stopped because the limit of the remaining evidence was reached")
+
+            if not stop:
+                logl_bounds = weight_function(sampler.results)  # derive bounds
+                for results in sampler.sample_batch(logl_bounds=logl_bounds, dlogz=dlogz, maxiter=maxiter):
+                    pass
+                sampler.combine_runs()  # add new samples to previous results
+            else:
+                break
+
+        yield sampler.results
+        return
 
 
     def _repr_html_(self):
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -573,12 +573,17 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
     :type sample: str, optional
     """
     analysis = analysis_file.analysis(posterior)
-    results = analysis.sample_nested(bound=bound, nlive=nlive, dlogz=dlogz, maxiter=maxiter, seed=seed, sample=sample)
-    samples = results.samples
-    posterior_values = results.logwt - results.logz[-1]
-    weights = _np.exp(posterior_values)
-    eos.data.DynestyResults.create(os.path.join(base_directory, posterior, 'nested'), analysis.varied_parameters, results)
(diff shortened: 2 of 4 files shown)


python/eos/analysis.py
```diff
@@ -628,8 +628,23 @@ def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=
            This method requires the dynesty python module, which can be installed from PyPI.
         """
         import dynesty
+        from dynesty.dynamicsampler import stopping_function, weight_function
         sampler = dynesty.DynamicNestedSampler(self.log_likelihood, self._prior_transform, len(self.varied_parameters), bound=bound, nlive=nlive, rstate = np.random.Generator(np.random.MT19937(seed)))
-        sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress)
+        #sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter)
+
+        for results in sampler.sample_initial(dlogz=dlogz, maxiter=maxiter):
```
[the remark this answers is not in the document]
mreboud: Every few samples is probably simpler to implement.

python/eos/analysis.py
```diff
@@ -607,7 +607,7 @@ def _prior_transform(self, u):
         return self._u_to_par(u)
 
 
-    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=10, print_progress=True):
+    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=10, print_progress=True, save_intermediate=False, base_directory='./', posterior=None):
```
dvandyk, line 610: That's a reasonable idea to begin with, but I do not like it very much. So far, the design was to separate the tasks from the lower-level operations. I think we should and can keep doing that.

I would suggest to use a ``yield_intermediate`` keyword argument. Instead of using ``return sampler.results`` once at the end, we rather ``yield sampler.results`` periodically. The task can then run a for loop over this function and save the results itself.

fnovak42:
To tick some of the boxes in issue #573 off:
- the log file now contains the messages generated by `eos.info` and is placed in the `nested` directory as discussed. So I would tick the third box off
- the fourth point is also done; we've replaced the loop
- the log file will now tell you if sampling was stopped due the maximum number of iterations or the `dlogz` limit being reached

Regarding the last point, currently we use the same variables `maxiter` and `dlogz` in the initial sampling (line 641 in `analysis.py`) and the batches (line 667) as well as the overall loop (lines 656, 661). Is it feasible to use different variables for those or should they adhere to the same limit?

I also thought about adding something like this if `maxiter` is not given:
```
import sys
if maxiter is None: maxiter = sys.maxsize
```

python/eos/analysis.py
```diff
@@ -621,9 +621,9 @@ def log_likelihood(self, p, *args):
         try:
             return(self._log_likelihood.evaluate())
         except RuntimeError as e:
-            eos.error('encountered run time error ({e}) when evaluating log(posterior) in parameter point:'.format(e=e))
+            eos.error(f'encountered run time error ({e}) when evaluating log(posterior) in parameter point:')
             for p in self.varied_parameters:
-                eos.error(' - {n}: {v}'.format(n=p.name(), v=p.evaluate()))
+                eos.error(f' - {p.name()}: {p.evaluate()}')
```
dvandyk, line 626: All of the changes above are due to ``pyupgrade`` and should be moved to a separate commit.

python/eos/analysis.py
```diff
@@ -653,14 +653,55 @@ def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=
         :type maxiter: int, optional
         :param seed: The seed used to initialize the Mersenne Twister pseudo-random number generator.
         :type seed: {None, int, array_like[ints], SeedSequence}, optional
+        :param save_intermediate: If set to True, the intemediate dynesty sampler results are stored in each loop iteration.
+        :type save_intermediate: bool, optional
+        :param checkpoint_every: The number of seconds between checkpoints at which the intermediate dynesty sampler results are stored.
```
dvandyk, line 658: I like the checkpointing. Not so sure that I like the naming.

python/eos/analysis.py
```diff
@@ -653,14 +653,55 @@ def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=
...
         .. note::
            This method requires the dynesty python module, which can be installed from PyPI.
         """
         import dynesty
+        import itertools
+        from dynesty.dynamicsampler import stopping_function, weight_function
         sampler = dynesty.DynamicNestedSampler(self.log_likelihood, self._prior_transform, len(self.varied_parameters), bound=bound, nlive=nlive, rstate = np.random.Generator(np.random.MT19937(seed)))
-        sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress)
-        return sampler.results
+        #sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter)
```
dvandyk, line 668: This comment needs to be removed.

python/eos/tasks.py
```diff
@@ -484,12 +484,12 @@ def sample_nested(analysis_file:str, posterior:str, base_directory:str='./', bou
...
-    posterior_values = results.logwt - results.logz[-1]
-    weights = _np.exp(posterior_values)
-    eos.data.DynestyResults.create(os.path.join(base_directory, posterior, 'dynesty_results'), analysis.varied_parameters, results)
-    eos.data.ImportanceSamples.create(os.path.join(base_directory, posterior, 'samples'), analysis.varied_parameters,
+    for iter, results in enumerate(analysis.sample_nested(bound=bound, nlive=nlive, dlogz=dlogz, maxiter=maxiter, seed=seed)):
+        samples = results.samples
+        posterior_values = results.logwt - results.logz[-1]
+        weights = _np.exp(posterior_values)
+        eos.data.DynestyResults.create(os.path.join(base_directory, posterior, f'dynesty_results-{iter:04}'), analysis.varied_parameters, results)
+        eos.data.ImportanceSamples.create(os.path.join(base_directory, posterior, f'samples-{iter:04}'), analysis.varied_parameters,
```
dvandyk, line 492: I like the labelling of the dynesty result by the iteration number. However, doing so for the samples will break backward compatibility with e.g. the plotting framework. Please undo this labelling for ``samples``.

dvandyk (changes requested):
Looks very reasonable! Some comments inline.

python/eos/analysis.py
mreboud: What about `backup` and `backup_frequency`?

pluegh: How about just a `checkpoint_interval` (float, larger zero)? If `None`, no intermediate results are saved.

fnovak42:
For comparison I calculated the chi**2 values from `test_multivariate_priors_1` in `python/eos/analysis_TEST.py`. The values using the code from the main branch (unrolled loop) are:
- chi 1: 0.002682592060968833
- chi 2: 5.803923701460194e-07
- chi 3: 2.52241831039844e-05
- chi 4: 0.2764884467520001

and using this branch:
- chi 1: 0.017353753382656843
- chi 2: 0.00031026173335609383
- chi 3: 0.0009632358001228368
- chi 4: 0.5324848524622118

I haven't yet spotted a significant difference to the dynesty version that would tell me where the difference comes from:  https://github.com/joshspeagle/dynesty/blob/f59d963fad80301e5d28bc5a6e3718c467c1bf94/py/dynesty/dynamicsampler.py#L1824

mreboud (commented):
Can you clean up line 176 of analysis_TEST.py? There is a missing sqrt, but sigma is not used anyway ...

dvandyk:
I rebased the old branch against ``master``. Making the tests cases pass required an increase of live points from ``250`` to ``300``.

@mreboud can you give this a thorough review?

dvandyk:
Something is definitely going awry. Compare the ``analysis-organisation`` example for $V_{ub}$ with the modification:
image
to the one before the modification:
image

mreboud:
Some of the functionalities have already been merged in another PR, some others are in progress. Closing this PR which fails the tests.

PR #651, opened 2023-06-15
[python] Fix the 'end' option in predict_observables

mreboud:


python/eos/tasks.py
```diff
@@ -394,7 +394,7 @@ def predict_observables(analysis_file:str, posterior:str, prediction:str, base_d
     :type base_directory: str, optional
     :param begin: The index of the first sample to use for the predictions. Defaults to 0.
     :type begin: int
-    :param end: The index beyond the last sample to use for the predictions. Defaults to -1.
+    :param end: The index of the last sample to use for the predictions. Defaults to -1.
```
dvandyk, line 397: Hm, I think this is wrong. Take a list ``l = ['a', 'b', 'c']``. Using ``l[:-1]`` yields ``['a', 'b']``. So the description was correct, but the default value is not. The default value should be ``None``.

python/eos/tasks.py
```diff
@@ -412,7 +412,17 @@ def predict_observables(analysis_file:str, posterior:str, prediction:str, base_d
 
     parameters = [_parameters[p['name']] for p in data.varied_parameters]
     observable_samples = []
-    for i, sample in enumerate(progressbar(data.samples[begin:end])):
+
+    if end == -1:
```
dvandyk, line 416: All this can be avoided by using ``None`` as the default. Sorry for this, I am pretty sure that I scewed it up in the first place.

python/eos/tasks.py
```diff
@@ -424,7 +434,7 @@ def predict_observables(analysis_file:str, posterior:str, prediction:str, base_d
     observable_samples = _np.array(observable_samples)
 
     output_path = os.path.join(base_directory, posterior, 'pred-{}'.format(prediction))
-    eos.data.Prediction.create(output_path, observables, observable_samples, data.weights[begin:end])
+    eos.data.Prediction.create(output_path, observables, observable_samples, weights)
```
dvandyk, line 437: Ditto, fixable by defaulting ``end`` to ``None``.

dvandyk (changes requested):

dvandyk (approved):

PR #656, opened 2023-06-20
Improve compiletime and used RAM of NLO charm-loop functions

vikusc:
The files holding the largest charm-loop functions F279b, F279c and F279d were further split and redundancies were removed


dvandyk (changes requested):
I cannot review the contents of this PR and have to rely on the test cases passing.

Comments mostly on the structure of the PR
 - The 1st commit does not fix a typo but a mislabelling of a variable. That wording is less likely to cause a heart attack in reviewer no. 1.
 - The 2nd and 3rd commit do use the gerund rather than the simple present tense.

PR #661, opened 2023-07-13
Various changes

dvandyk:
- adapt analysis file format to preliminary documentation / v1.1 paper draft
 - fix use of back ticks in the documentation
 - clarify minimal version requirements in the documentation
 - modernise `B_q->lnu` and `B->Plnu` observables


doc/installation.rst
```diff
@@ -131,25 +131,25 @@ The dependencies can be roughly categorized as either system software, Python-re
...
 pkg-config
-  the freedesktop.org library helper, in version 0.28 or higher;
+  the freedesktop.org library helper, in version 0.27.1 or higher;
 
 BOOST
   the BOOST C++ libraries (sub-libraries ``boost-filesystem`` and ``boost-system``);
 
 yaml-cpp
-  a C++ YAML parser and interpreter library.
+  a C++ YAML parser and interpreter library, in version 0.5.1 or higher.
```
SKuerten, line 152: The API has changed in version 0.5.0, so I think we should include it - if the reason for the restriction is the API.

doc/installation.rst
```diff
@@ -187,7 +187,7 @@ We recommend you install the above packages via your system's software managemen
 EOS requires the following scientific software:
 
 GSL
-  the GNU Scientific Library \cite{GSL}, in version 1.16 or higher;
+  the GNU Scientific Library \cite{GSL}, in version 1.15 or higher;
```
SKuerten, line 190: Why do we include an earlier version here?

eos/b-decays/b-to-l-nu.cc
```diff
@@ -129,8 +140,10 @@ namespace eos
     const std::vector<OptionSpecification>
     Implementation<BToLeptonNeutrino>::options
     {
-        { "l", { "e", "mu", "tau" }, "mu" },
-	    { "q", { "c", "u" }, "c"}
+        Model::option_specification(),
+        { "cp-conjugate", { "true", "false" },  "false" },
+        { "l",            { "e", "mu", "tau" }, "mu"    },
+	    { "q",            { "c", "u" },         "c"     }
```
SKuerten, line 146: this looks like it should be aligned by shifting 4 spaces to the left

eos/b-decays/b-to-psd-l-nu.cc
```diff
@@ -426,10 +431,11 @@ namespace eos
     {
         Model::option_specification(),
         FormFactorFactory<PToP>::option_specification(),
-        { "l", { "e", "mu", "tau" }, "mu" },
-        { "U", { "c", "u" }, "c" },
-        { "q", { "u", "d", "s" }, "d" },
-    	{ "I", { "1", "0", "1/2" }, "1" },
+        { "cp-conjugate", { "true", "false" },  "false" },
+        { "l",            { "e", "mu", "tau" }, "mu"    },
+        //{ "U",          { "c", "u" },         "c"     },
```
SKuerten, line 436: why keep those comments? The information is also earlier in this file in line 163.

doc/installation.rst
dvandyk: We have always depended on version 0.5.0 or higher, so I would not bother. This is just to clarify the minimal version.

doc/installation.rst
dvandyk: Good catch. Because it actually builds with version 1.15 and that is what we currently use at IPPP.

eos/b-decays/b-to-l-nu.cc
dvandyk: This is weird, let me check what happened.

eos/b-decays/b-to-psd-l-nu.cc
dvandyk: I would like to keep it, since we need to move it back here once ``OptionSpecification`` allows to hide an option to the user. Right now, it is hidden manually.

doc/installation.rst
mreboud: If it works for 1.15, then this change is justified

PR #664, opened 2023-07-20
[python] Add option 'distribution' to corner plot

### LINKED ISSUE #657: The `corner-plot` task should also support plotting posterior predictions
Currently, only the distribution of the posterior samples can be plotted. It would be nice if we could repurpose the code to also plot posterior predictive samples as obtained using the `predict-observables` task.

pluegh:
Closes #657


mreboud (approved):
Very clean, thanks @pluegh

Issue #662, opened 2023-07-13
Remove support for Python 3.7


dvandyk:
As of 2023-06-27, Python 3.7 is no longer supported.

 - [x] Remove `py37` from the `manylinux2014` workflow.
 - [x] Bump minimum Python version to `3.8` in `configure.ac`.
 - [x] Bump minimum Python version to `3.8` in `python/setup.py.in`.

dvandyk:
Resolved by @vikusc through PR #665

PR #665, opened 2023-07-24
[python] Remove support for Python 3.7 - issue #662

vikusc:
Solve issue #662


dvandyk (approved):
This looks very clean. Thanks @vikusc! Let's see if the last few checks pass now.

vikusc:
The last 3 checks should fail since they build with python 3.7, don't they?

PR #668, opened 2023-07-27
Small fixes for the YETI tutorial

mreboud:


python/eos/plot/plotter.py
```diff
@@ -461,10 +461,21 @@ def plot(self):
...
+                eos.warn('    overriding values read from \'parameters-from-mode\' with explicit values in \'parameters\'')
+
             if 'parameters-from-file' in item and type(item['parameters-from-file']) is str:
                 eos.warn('    overriding parameters from file')
                 parameters.override_from_file(item['parameters-from-file'])
 
+            if 'parameters-from-mode' in item and type(item['parameters-from-mode']) is str:
+                eos.warn('    overriding parameters from mode')
+                mode = eos.Mode(item['parameters-from-mode'])
+                if not mode.type == "Mode":
```
dvandyk, line 474: Good to check here. But the check should be part of the data file class, really. If it is not yet, let's open an issue.

dvandyk (approved):
Thanks Meril, looks good for merging!

python/eos/plot/plotter.py
mreboud: It is already :)

dvandyk: Then this can go, right? I would prefer removing it. Suppose we switch from a ``Mode`` to ``Mode-v2`` format in the future, this might lead to unexpected bugs.

mreboud:
I will merge once all tests pass :)

PR #671, opened 2023-08-03
New plotter

pluegh:
The proposed changes introduce a new submodule `eos.figure` to the python interface that should eventually replace the current `eos.plot`. As of now, the module is not feature-complete but can be used in addition/parallel to the existing one.

The changes facilitate a number of improvements:

- They create a clear separation of tasks. For instance, responsibilities like managing the appearance of a figure, generating data, and validating user input are handled separately.
- This separation enables automatic and comprehensive API documentation.
- It promotes consistent use of docstrings, unifying documentation and code within the same context to enhance maintainability of the documentation.
- Documentation and the Python `help()` function are unified, making both equally helpful.
- Overall code maintainability is enhanced.
- The changes also enhance test isolation, making testing more diagnostic and secure.

At the current stage, the code would benefit from thorough review.

### DIFF
--- a/doc/api/python.rst.jinja
+++ b/doc/api/python.rst.jinja
@@ -288,3 +288,80 @@ All item types recognize the following optional keys:
 
 {% endif %}
 {% endfor %}
+
+
+*********************
+Module ``eos.figure``
+*********************
+
+EOS also provides a new plotting framework with a different interface, which will replace
+the existing framework as soon as it is feature complete.
+
+.. note::
+
+   Import :py:mod:`eos.figure` before you do something like ``import matplotlib.pyplot as plt``,
+   because the :py:mod:`eos.figure` module modified the default plot style and the matplotlib backend.
+   All options (except the backend) can be overwritten by updating ``matplotlib.rcParams[...]``;
+   see also the `matplotlib <https://matplotlib.org/>`_ documentation.
+   Note that the default settings use LaTeX to create labels and math expressions,
+   so for this to work latex needs to be available on your system.
+
+
+Figures
+=======
+
+.. automodule:: eos.figure
+   :members:
+
+
+Plot types
+----------
+
+Below we list the available plot types.
+Each key ``plot_type`` corresponds to a class which is instantiated in the background.
+The user can supply keyword arguments which are documented under :py:mod:`eos.figure.plot`.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - ``plot_type``
+     - class
+     - description
+   {% for key, class, description in figure_plot_types %}
+   * - ``{{ key }}``
+     - :py:class:`{{ class }}`
+     - {{ description }}
+   {% endfor %}
+
+The classes including the arguments the user can pass are documented in detail below.
+
+.. automodule:: eos.figure.plot
+   :members:
+
+
+Item types
+----------
+
+Below we list the available item types.
+Each key ``item_type`` corresponds to a class which is instantiated in the background.
+The user can supply keyword arguments which are documented under :py:mod:`eos.figure.item`.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - ``item_type``
+     - class
+     - description
+   {% for key, class, description in figure_item_types %}
+   * - ``{{ key }}``
+     - :py:class:`{{ class }}`
+     - {{ description }}
+   {% endfor %}
+
+The classes including the arguments the user can pass are documented in detail below.
+
+.. automodule:: eos.figure.item
+   :exclude-members: Item
+   :members:
--- a/python/eos/figure/figure.py
+++ b/python/eos/figure/figure.py
@@ -1,0 +1,138 @@
+# Copyright (c) 2023 Danny van Dyk
+# Copyright (c) 2023 Philip Lueghausen
+#
+# This file is part of the EOS project. EOS is free software;
+# you can redistribute it and/or modify it under the terms of the GNU General
+# Public License version 2, as published by the Free Software Foundation.
+#
+# EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+# details.
+#
+# You should have received a copy of the GNU General Public License along with
+# this program; if not, write to the Free Software Foundation, Inc., 59 Temple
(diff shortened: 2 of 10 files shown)


doc/api/python.rst.jinja
```diff
@@ -288,3 +288,80 @@ All item types recognize the following optional keys:
...
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - `plot_type`
+     - class
+     - description
+   {% for key, class, description in figure_plot_types %}
+   * - ``{{ key }}``
```
dvandyk, line 332: double back ticks vs single back ticks

python/eos/figure/figure.py
```diff
@@ -0,0 +1,142 @@
...
+# Place, Suite 330, Boston, MA  02111-1307  USA
+
+from .plot import PlotFactory
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+class Figure:
+    def __init__(self, figure, axes, descriptions:list): # plots -> descriptions
```
dvandyk, line 24: What is the purpose of the comment here?

python/eos/figure/figure.py
```diff
@@ -0,0 +1,142 @@
...
+    This function takes a figure description and handles the contruction of a publication-grade
+    figure in a convenient way.
+    The description contains the arguments used for initialization of the objects in the
+    `eos.figure` module.
+
+    A figure can contain one or more plots, where each plot containts one or more items such
+    as a data series from the evaluation of an observable or a kernel density estimate to
+    visualize a set of statistical samples.
+
+    The description can be either a dictionary (`figure description`) discribing a single plot
```
dvandyk, line 99: ``discribing`` -> ``describing``

python/eos/figure/figure.py
```diff
@@ -0,0 +1,142 @@
...
+    +-------------+----------------------------------+
+
+    +-------------+----------------------------------+
+    | `item description` (dict)                      |
+    +=============+==================================+
+    | `item_type` | valid item type (str)            |
+    +-------------+----------------------------------+
+    | ...         | one or several keyword arguments |
+    +-------------+----------------------------------+
+
```
dvandyk, line 125: A simple example would go a long way here!

python/eos/figure/figure.py
```diff
@@ -0,0 +1,142 @@
...
+    :param nrows: The number of rows in the grid.
+    :type nrows: int
+    :param ncols: The number of columns in the grid.
+    :type ncols: int
+    :param plots: The list of plots' data, as a list of dictionaries each containing the 'description' and the 'contents' for a single plot.
+    :type plot: list[dict] or iterable[dict]
+    """
+
+    def __init__(self, nrows, ncols, descriptions, **kwargs):
```
dvandyk, line 82: Is this a convenient order of parameters? I would probably move ``descriptions`` to before ``nrows, ncols``. But maybe there is a good reason for the current ordering?

python/eos/figure/item.py
```diff
@@ -0,0 +1,178 @@
...
+        self.label = label
+
+
+        # Create kinematics
+        eos_kinematics = eos.Kinematics()
+        if fixed_kinematics:
+            for k, v in fixed_kinematics.items():
+                if k not in valid_kin_vars:
+                    raise ValueError("Kinematic quantity '" + k + "' does not " +
+                    "match known ones for observable '" + observable + "': " + valid_kin_vars.__repr__())
```
dvandyk, line 95: ``quantity`` -> ``variable``
``known ones`` -> ``any of the declared kinematic variables``?

python/eos/figure/item.py
```diff
@@ -0,0 +1,178 @@
...
+                eos_kinematics.declare(k, v)
+
+        # Create parameters
+        eos_parameters = eos.Parameters.Defaults()
+        if fixed_parameters and fixed_parameters_from_file:
+            eos.warn('Overriding values read from \'parameters-from-file\' with explicit values in \'parameters\'')
+            raise RuntimeError("Providing 'fixed_parameters_from_file' is not yet implemented")
+
+        elif fixed_parameters_from_file:
```
dvandyk, line 104: Could you copy what @mreboud had recently added to the old plotting framework in terms of ``parameters-from-mode`` when plotting an observable?

python/eos/figure/plot.py
```diff
@@ -0,0 +1,156 @@
...
+    def draw(self):
+        "Draw nothing"
+        pass
+
+
+class PlotFactory:
+    # Also build the documentation based on ordered registry
+    # Initializer is well-defined for python version >= 3.6
+    registry = OrderedDict({
+        '2dplot': TwoDimPlot,
```
dvandyk, line 127: I would prefer ``2D-plot`` or somesuch, highlighting the "d".

python/eos/figure/plot.py
```diff
@@ -0,0 +1,156 @@
...
+            del result[old_key]
+
+        # print(f'Sanitize result = {result}')
+        return result
+
+
+    @staticmethod
+    def make(ax, contents:list, type:str, **kwargs):
+        if type not in PlotFactory.registry:
+            raise ValueError(f'Unknown plot type: { type }')
```
dvandyk, line 154: Hm, does it make sense to default to a ``2D-plot`` if no plot type is specified?

dvandyk (changes requested):
General approval, well done @pluegh!

Some comments and requests for changes below, including a few typos.

python/eos/figure/item.py
pluegh: Sure thing. But if we add non-trivial features to the bare bones now, it is also a good time to add unit tests. If @mreboud already has a test case at hand, maybe he wants to add it? 😄

python/eos/figure/figure.py
dvandyk: Fair enough, closing this for now.

doc/api/python.rst.jinja
dvandyk: Had another look and fixed the single back ticks. Single back ticks should only be used as arguments to a Sphinx or RST directive, e.g.,
```RST
This is an example linking to the Python class :py:class:`eos.figure.Figure`. It uses, amongst others,
the ``matplotlib.pyplot`` module.
```
Mentions of classes not included in our API documentation are using double back ticks to indicate code.

python/eos/figure/item.py
dvandyk: Suspending this for the time being.

dvandyk (approved):

PR #674, opened 2023-08-05
[python] Allow for a list of kinematic assignements in the prediction task

### LINKED ISSUE #673: Allow for a list of kinematic specifications in the predictions part of an analysis file
I suggest that this should e.g. be a valid syntax:
```
predictions:
  - name: form-factors
    observables:
      - name: B->K^*::V(q2)
        kinematics:
          [ {q2: -15}, {q2: -10}, {q2: -5}, {q2: 0}, {q2: 5} ]
```

mreboud:
Closes #673

### DIFF
--- a/python/eos/analysis_file.py
+++ b/python/eos/analysis_file.py
@@ -164,12 +164,22 @@ class AnalysisFile:
                 if key in global_options and global_options[key] != value:
                     eos.error(f'Global option {key}={global_options[key]} overrides option part specification {key}={value} for observable {o["name"]} in prediction {_prediction}.')
 
-        observables = [eos.Observable.make(
-            o['name'],
-            parameters,
-            eos.Kinematics(**(o['kinematics'] if 'kinematics' in o else {})),
-            options
-        ) for o in prediction['observables']]
+        observables = []
+        for o in prediction['observables']:
+            if 'kinematics' not in o:
+                kinematics = [{}]
+            elif type(o['kinematics']) == dict:
+                kinematics = [o['kinematics']]
+            else:
+                kinematics = [k for k in o['kinematics']]
+
+            for k in kinematics:
+                observables.append(eos.Observable.make(
+                    o['name'],
+                    parameters,
+                    eos.Kinematics(k),
+                    options
+                ))
 
         if None in observables:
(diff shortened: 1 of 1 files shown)


python/eos/analysis_file.py
```diff
@@ -164,12 +164,32 @@ def observables(self, _posterior, _prediction, parameters):
...
+                    ) for kinematics in o['kinematics']]
+                elif type(o['kinematics']) == dict:
+                    observables.append(eos.Observable.make(
+                        o['name'],
+                        parameters,
+                        eos.Kinematics(o['kinematics']),
+                        options
+                    ))
+                else:
+                    eos.error(f'The kinematic specification of the observable {o["name"]} should be a dictionnary or a list of dictionnaries.')
```
dvandyk, line 185: ``dictionnary`` -> ``dictionary``
``dictionnaries`` -> ``dictionaries``

python/eos/analysis_file.py
```diff
@@ -164,12 +164,32 @@ def observables(self, _posterior, _prediction, parameters):
...
 
-        observables = [eos.Observable.make(
-            o['name'],
-            parameters,
-            eos.Kinematics(**(o['kinematics'] if 'kinematics' in o else {})),
-            options
-        ) for o in prediction['observables']]
+        observables = []
+        for o in prediction['observables']:
+            if 'kinematics' in o:
```
dvandyk, line 169: The following might by slightly more readable and maintainable?

```python
if 'kinematics not in o:
    kinematics = [{}]
elif type(o['kinematics']) == dict:
   kinematics = [o['kinematics']]
else:
   kinematics = [k for k in o['kinematics']]

for k in kinematics:
    observables.append(eos.Observable.make(
        o['name'],
        parameters,
        eos.Kinematics(k),
        options
    ))
```

dvandyk (changes requested):
I generally approve the change, except for the two typos in the error message.

My suggested code snippet might be more maintainable, since it only has one site in which ``eos.Observable.make`` is used.

python/eos/analysis_file.py
mreboud: Definitely, thanks!

PR #685, opened 2023-09-01
Allow to declare parameters including metadata

dvandyk:


mreboud (approved):
There are still a couple of places where the old `declare` function is used. The rest looks good, thanks!

dvandyk:
@mreboud Please re-review. I ran into trouble after fixing ``eos/utils/kmatrix_TEST.cc``, with your ``KMatrix`` class complaining about the central value for a new mass parameter that was declared. Turned out I had mixed up the order of the minimal and central value of the parameter's value when storing inside ``eos::Parameters``.

mreboud (approved):

PR #686, opened 2023-09-01
Create EOS - pyhf interface

lorenzennio:


python/_eos.cc
```diff
@@ -46,6 +46,78 @@
...
+
+            virtual double significance() const override
+            {
+                throw InternalError("Not implemented");
+                return 0.0;
+            }
+
+            virtual TestStatistic primary_test_statistic() const override
+            {
+                return test_statistics::Empty();
```
dvandyk, line 116: OK for now. In the future, we might want to return known / implemented test statistics, which will require adding further mandatory member functions to the underlying python implementation.

eos/statistics/log-likelihood.cc
```diff
@@ -1399,7 +1402,7 @@ namespace eos
         {
             double result = 0.0;
 
-            // loop over all likelihood blocks
+            // loop over all constrain-based likelihood blocks
```
dvandyk, line 1405: Typo: ``constrain`` -> ``constraint``

python/_eos.cc
```diff
@@ -673,6 +673,7 @@ BOOST_PYTHON_MODULE(_eos)
             Represents the log(likelihood) of a Bayesian analysis undertaken with the :class:`Analysis <eos.Analysis>` class.
         )", init<Parameters>())
         .def("add", (void (LogLikelihood::*)(const Constraint &)) &LogLikelihood::add)
+        .def("add", (void (LogLikelihood::*)(const LogLikelihoodBlockPtr &)) &LogLikelihood::add)
```
dvandyk, line 676: Uhh, does this work? C++ supports overloading. Does Boost-Python support overloading C++ functions? Checking....

dvandyk: Nevermind, this is exactly as it is intended. See https://stackoverflow.com/questions/7577410/boost-python-select-between-overloaded-methods

python/eos/analysis.py
```diff
@@ -57,6 +56,8 @@ class Analysis:
     :type priors: iterable
     :param likelihood: The likelihood as a list of individual constraints from the internal data base of experimental and theoretical constraints; cf. `the complete list of constraints <../constraints.html>`_.
     :type likelihood: iterable
+    :param external_likelihood: The likelihood as a list of external likelihoods constructed as a `eos.LogLikelihoodBlock.External` likelihood block.
```
dvandyk, line 59:
```
The extenal likelihood blocks as a list or iterable of objects returned by :py:meth:`eos.LogLikelihoodBlock.External` .
```

python/eos/analysis.py
```diff
@@ -57,6 +56,8 @@ class Analysis:
...
+    :type external_likelihood: iterable
```
dvandyk, line 60: ``list or iterable of :py:class:`eos.LogLikelihoodBlock`.``

python/eos/pyhf.py
```diff
@@ -0,0 +1,37 @@
...
+        self.data = np.array(workspace.data(self.model, include_auxdata=False))
+
+        self.varied_parameters = []
+        for pn, v in zip(self.model.config.par_names, self.model.config.suggested_init()):
+            if pn in parameter_map:
+                qn = eos.QualifiedName(parameter_map[pn])
+            else:
+                qn = eos.QualifiedName('pyhf::' + pn)
+
+            p = self.parameters.declare(qn, v)
```
dvandyk, line 32: This is not quite OK. We should only declare this parameter if it is not declared yet!
```
if self.parameters.had(qn):
    p = self.parameters[qn]
else:
    p = self.parameters.declare(qn, v)
```

dvandyk (changes requested):
Looks nice already! Some changes requested, see my comments.

eos/statistics/log-likelihood.cc
dvandyk: Fixed by @lorenzennio

python/eos/analysis.py
dvandyk: Fuck. I typoed this. ``extenal`` -> ``external``.

python/eos/analysis.py
dvandyk: Fixed by @lorenzennio

python/eos/pyhf.py
dvandyk: I fat-fingered this one as well: ``self.parameters.has``.

python/eos/pyhf.py
```diff
@@ -0,0 +1,85 @@
...
+                        'min': param_set.suggested_bounds[i][0], 
+                        'max': param_set.suggested_bounds[i][1],
+                        'type': 'gauss',
+                        'central': param_set.auxdata[i],
+                        'sigma': param_set.width()[i]
+                    }
+                    priors.append(prior)
+
+            elif param_set.pdf_type == 'poisson':
+                raise NotImplementedError
```
dvandyk, line 83: This can now be implemented since you added ``eos::priors::Poisson``, right?

python/eos/pyhf.py
```diff
@@ -0,0 +1,85 @@
+import eos
```
dvandyk, line 1: This file is missing the typical boiler plate. See e.g. ``python/eos/analysis_file.py``.

dvandyk (changes requested):
A few more comments.

dvandyk (approved):

eos/statistics/log-prior.cc
```diff
@@ -619,6 +619,74 @@ namespace eos
...
+                Parameter _parameter;
+
+                std::string _name;
+
+                const double _k;
+
+                const double _ln_norm;
+
+            public:
+                Poisson(const Parameters & parameters, const std::string & name, const double & k) :
```
lorenzennio, line 707: I think we might have made a mistake here. The distribution should be a function of $k$ with $\lambda$ as a parameter. I think we accidentally switched the roles of the two. And the connection to the Gamma distribution should be $Poisson(k;\lambda) = Gamma(\lambda; a=k+1, b=1)$.
But maybe I confused something here.

lorenzennio, line 707: I think we might have made a mistake here. The distribution should be a function of $k$ with $\lambda$ as a parameter. I think we accidentally switched the roles of the two. And the connection to the Gamma distribution should be $Poisson(k;\lambda) = Gamma(\lambda; a=k+1, b=1)$.
But maybe I confused something here.

dvandyk: No mistake, and you description is entirely correct. This log prior returns the PDF for any input $\lambda$ given some fixed value of $k$.

Below, in ``virtual double operator()() const`` you extract the value of $\lambda$ from the parameter object. You simply call it ``x`` instead.

python/_eos.cc
```diff
@@ -797,6 +797,34 @@ BOOST_PYTHON_MODULE(_eos)
             :type lambda: float, strictly positive
         )", args("parameters", "name", "min", "max", "mu_0", "scale"))
         .staticmethod("Scale")
+        .def("Gaussian", &LogPrior::Gaussian, return_value_policy<return_by_value>(), R"(
```
dvandyk, line 800: Bah, there is a typo in the commit message: ``Poission`` -> ``Poisson``.

python/eos/pyhf.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
+# vim: set sw=4 sts=4 et tw=120 :
+
+# Copyright (c) 2020-2023 Danny van Dyk
```
dvandyk, line 4: This needs Lorenz's copyright notice instead! Please remove mine or at least limit it to 2023.

python/eos/pyhf.py
dvandyk: My previous comment still stands. I had already removed the ``min``/``max`` values to use a "true" Gaussian prior w/ infinite support, rather than the curtailed Gaussian we were limited to in past EOS versions.

dvandyk (changes requested):
Only one blocker:
 - emit an entry for a ``poisson`` prior when generating the list of suggested priors for an existing ``pyhf`` workspace.

python/eos/pyhf.py
dvandyk: We just fixed that. Resolved.

eos/statistics/log-prior.cc
dvandyk: This was correct for the general case as anticipated. However, histfactory uses a rescaling by the expected value. Since we aim only to support histfactory likelihoods with this, we use the same convention. This leads all ``poisson``-distributed parameters to peak close to $1$.

python/_eos.cc
dvandyk: @lorenzennio fixed this.

python/eos/pyhf.py
dvandyk: Also fixed.

dvandyk (approved):

PR #687, opened 2023-09-02
Use jinja to generate list of constraints, parameters, and references

### LINKED ISSUE #678: Fully migrate python documentation from scripts to Jinja templates
As of now, the documentation for parts of the Python code is generated based on scripts.
The other parts already take advantage of Jinja templates, to separate processing from representation of information. The purpose is to improve readability and maintainability.

The following scripts need adaption (analogous to e.g. `doc/observables.rst.{py,jinja}`):
- [ ] `doc/constraints.rst.py`
- [ ] `doc/parameters.rst.py`
- [ ] `doc/references.rst.py`

The changes of `doc/Makefile` are already prepared in PR #671 and should be addressed after the PR is merged.
Specifically, the accommodation for the legacy implementation should be removed.

dvandyk:
Resolves #678

### DIFF
--- a/doc/parameters.rst.jinja
+++ b/doc/parameters.rst.jinja
@@ -1,0 +1,39 @@
+##################
+List of Parameters
+##################
+
+
+The following is the full list of parameters and their values as used in EOS v{{ version }}.
+
+{% for section in sections %}
+
+{{ "*" * len(section.name) }}
+{{ section.name }}
+{{ "*" * len(section.name) }}
+
+{% for group in section.groups %}
+{{ group.name }}
+{{ "=" * len(group.name) }}
+
+.. list-table::
+   :widths: 40, 20, 20, 20
+   :header-rows: 1
+
+   * - Qualified Name
+     - Representation
+     - Default Value
+     - Unit
+
+   {% for param in group.parameters %}
+       .. _{{ param.link_key }}:
+   * - ``{{ param.qualified_name }}``
+     - {{ param.latex }}
+     - {{ param.value }}
+     - {{ param.unit }}
+
+   {% endfor %}
+
+{{ group.description }}
+
+{% endfor %}
+{% endfor %}
--- a/doc/references.rst.jinja
+++ b/doc/references.rst.jinja
@@ -1,0 +1,14 @@
+##################
+List of References
+##################
+
+
+The following is the full list of references that were used to implement theory codes and provide likelihood constraints
+or parameter values within EOS v{{ version }}.
+
+{% for handle, reference in references %}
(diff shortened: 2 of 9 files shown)


doc/references.rst.jinja
pluegh, line 1: This produces a delimiter even after the title even if there is no eprint (similar for missing title). On a side note, the whitespace in the RST is all over the place.
I will rebase the branch with a fix in a moment.

pluegh (changes requested):

doc/parameters.rst.jinja
```diff
@@ -0,0 +1,39 @@
...
+   {% for param in group.parameters %}
+       .. _{{ param.link_key }}:
+   * - ``{{ param.qualified_name }}``
+     - {{ param.latex }}
+     - {{ param.value }}
+     - {{ param.unit }}
+
+   {% endfor %}
+
+{{ group.description }}
```
pluegh, line 36: The group description sometimes includes display math (equation, align). I remove this for now, since the implementation is not trivial.

dvandyk: We can translate that as well, can we not?

pluegh (changes requested):
Needs a small fix and it would be nice to suppress the sphinx warnings due to unreferences citations (an intentional use case for us).

doc/parameters.rst.jinja
pluegh: After trying for a couple minutes, I ran into trouble with align, for example, here:

https://github.com/eos/eos/blob/43e942742fb0946a6885fa9c49bb6ea66367c8f0/eos/parameters/sm-and-eft.yaml#L880-L892

dvandyk: Isn't the only thing needed here to replace
```
\begin{align}
   %content
\end{align}
```
with
```
.. math::

   %content
```
or am I missing something?

dvandyk: See also here: https://sphinx-rtd-trial.readthedocs.io/en/latest/ext/math.html#directive-math

dvandyk:
> Needs a small fix and it would be nice to suppress the sphinx warnings due to unreferences citations (an intentional use case for us).

The new commit does not follow the repository template for commit messages.

doc/parameters.rst.jinja
pluegh: It should work now!
There was a minor problem with newlines in the yaml files and aligned equations with only a single line (and a single `&` character).

dvandyk:
@pluegh Please re-review and merge if acceptable.

pluegh (approved):

Issue #615, opened 2023-03-15
Add generic class implementing the Gegenbauer polynomials


dvandyk:
Would it be worth defining these polynomials in a separate file in `eos/maths`? E.g. to improve their evaluation which is not optimal here...

_Originally posted by @mreboud in https://github.com/eos/eos/pull/614#discussion_r1136694858_

dvandyk:
Files that are known to use Gegenbauer polynomials:
 - ``eos/form-factors/pi-lcdas.cc``
 - ``eos/form-factors/k-lcdas.cc``

PR #691, opened 2023-09-12
Solve Issue #615 - Implement Gegenbauer polynomials

vikusc:


eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright (c) 2021 Danny van Dyk
+ * Copyright (c) 2022 Méril Reboud
```
dvandyk, line 3: This is incorrect. Please add your own name for ``2023``

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
...
+    class GegenbauerPolynomial
+    {
+        private:
+            const unsigned int order;
+
+            const double alpha; // alpha > -0.5
+
+            std::vector<double> coefficients;
+
+            const unsigned int halforder;
```
dvandyk, line 37: It would be helpful to add some comments at the top of the class to explain the order and halforder of the polynomial(s).

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
...
+                    }
+
+                    return result;
+            }
+
+        public:
+            GegenbauerPolynomial(const unsigned int order, const double alpha) :
+            order(order),
+            alpha(alpha),
+            halforder(floor(static_cast<double>(order) / 2.0)),
```
dvandyk, line 66: Ah, I see. So, half order does not need to be documented as an external input. Only the ``order`` and ``alpha``.

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
...
+
+            const double alpha; // alpha > -0.5
+
+            std::vector<double> coefficients;
+
+            const unsigned int halforder;
+
+            std::vector<double> calculate_coefficients() const
+            {
+                const unsigned int isodd = order % 2;
```
dvandyk, line 41: Would be helpful to store this in the class object itself upon construction. Maybe a member variable ``order_is_odd``, which is boolean?

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
...
+                const double epsilon = 1e-8;
+                std::vector<double> result;
+
+                for (unsigned i = 0 ; i <= halforder ; i++)
+                {
+                    result[i] = power_of<2*i + isodd>(2) * power_of<halforder - i>(-1.0) * 
+                        tgamma(alpha + halforder + i + isodd) / (tgamma(halforder - i + 1) * tgamma(2*i + isodd + 1)); // denominator: (halforder - i)! * (2i + isodd)!
```
dvandyk, line 48: In what header is ``tgamma`` declared?

dvandyk (changes requested):
Solid first attempt. See comments. Following "global" issues:

 - This is a header-only implementation, however, without any need for it. Should probably be split into a ``.cc`` and a ``.hh`` file.
 - Coding style. Variable identifiers should be ``lower_case_with_underscores``, not ``smushedtogethersonobodycanreadit`` :smile:

eos/maths/gegenbauer-polynomial.hh
vikusc: its in cmath, but I am thinking about expressing the coefficients without the gamma function

eos/maths/gegenbauer-polynomial.hh
vikusc: I don't understand why that would be that helpful, but I can do that. however I need it as integer since it appears in the calculation of the coefficients

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,91 @@
...
+
+                if (order % 2 == 0)
+                    x = 1;
+                else
+                    x = z;
+
+                for(unsigned int i; i <= halforder; i++)
+                {
+                    result += coefficients[i] * x;
+                    x *= z*z;
```
mreboud, line 83: Can you use space in algebraic expressions?

eos/maths/gegenbauer-polynomial.hh
dvandyk: In that case, name it to something that is more obvious to the reader, e.g., ``power_shift`` and set it explicitly to an integer value at the construction time of the polynomial:
```C++
const int power_shift = (1 == order % 2) ? 1 : 0; // we need to shift the power in some expressions by 1 if the order is odd
```

vikusc: I implemented the factor as suggested in the first message, as it appears also in the evaluate method

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,55 @@
...
+#include <eos/maths/power-of.hh>
+
+#include <vector>
+
+namespace eos
+{
+    class GegenbauerPolynomial
+    {
+        private:
+            const unsigned int order; // order of the Gegenbauer polynomial
```
dvandyk, line 30: Coding style: The names of private member variables should be prefixed with an underscore, i.e., ``_order``.

eos/maths/gegenbauer-polynomial.cc
```diff
@@ -0,0 +1,64 @@
...
+#include <cmath>
+#include <vector>
+
+namespace eos
+{
+    std::vector<double> GegenbauerPolynomial::calculate_coefficients() const
+    {
+        const double epsilon = 1e-8;
+        double entry;
+        std::vector<double> result (half_order + 1);
```
dvandyk, line 30: Coding style: please remove the whitespace between the identifier and the call to the constructor.

eos/maths/gegenbauer-polynomial.cc
```diff
@@ -0,0 +1,64 @@
...
+
+#include <cmath>
+#include <vector>
+
+namespace eos
+{
+    std::vector<double> GegenbauerPolynomial::calculate_coefficients() const
+    {
+        const double epsilon = 1e-8;
+        double entry;
```
dvandyk, line 29: C++: Variables should be declared only very close to where they are used. Here, the ``double entry`` should be part of the ``for { ... }`` loop, e.g.,
```C++
for (...)
{
    const double entry = pow(...) * ...;
    result[i] = entry;
}
```

eos/maths/gegenbauer-polynomial.cc
```diff
@@ -0,0 +1,64 @@
...
+
+        for (unsigned int i = 0 ; i <= half_order ; i++)
+        {
+            entry = pow(2.0, 2.0 * i + 1.0 * r) * pow(-1.0, 1.0 * (half_order - i)) *
+                tgamma(alpha + 1.0 * (half_order + i + r)) / (tgamma(1.0 * (half_order - i + 1)) * tgamma(2.0 * i + 1.0 * (r + 1))); // denominator: (half_order - i)! * (2i + r)!;
+
+            result.at(i) = entry;
+        }
+
+        if (std::abs(alpha) <= epsilon)
```
dvandyk, line 40: Coding style: use braces

eos/maths/gegenbauer-polynomial.cc
```diff
@@ -0,0 +1,64 @@
...
+            return result;
+        else
+            for (unsigned int i = 0 ; i <= half_order ; i++)
+            {
+                result[i] = result[i] / tgamma(alpha);
```
dvandyk, line 45: You are making ``N = half_order + 1`` calls to ``tgamma`` here, although you need to evaluate it only once. How about the following.

```C++
double gamma_alpha = (alpha < epsilon) ? 1.0 : tgamma(alpha);
for (unsigned int i = 0 ; ...)
{
    result[i] /= gamma_alpha;
}
return result;
```

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,55 @@
...
+
+            const double alpha; // parameter of the Gegenbauer polynomial with alpha > -0.5
+
+            const unsigned int r; // remainder of (order / 2)
+
+            const unsigned int half_order; // rounded down half of the order: floor(order / 2)
+
+            std::vector<double> calculate_coefficients() const;
```
dvandyk, line 38: Ditto for member functions: ``_calculate_coefficients()``.

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,55 @@
...
+
+            std::vector<double> coefficients;
+
+        public:
+            GegenbauerPolynomial(const unsigned int & order, const double & alpha) :
+                order(order),
+                alpha(alpha),
+                r(order % 2),
```
dvandyk, line 46: Can you please make this explicit? You are currently casting from an integer type to a floating point type. You do not need to.

```C++
_r((order % 2 == 0) : 0.0 : 1.0)
```

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,55 @@
...
+                half_order(floor(static_cast<double>(order) / 2.0)),
+                coefficients(calculate_coefficients())
+            {}
```
dvandyk, line 49: Coding  style: opening and closing braces on lines of their own.

eos/maths/gegenbauer-polynomial.hh
```diff
@@ -0,0 +1,55 @@
...
+
+            double polynomial_evaluation(const double & z);
```
dvandyk, line 51: The name is a bit unwieldy. Why not ``evaluate(const double & z)``?

eos/maths/gegenbauer-polynomial_TEST.cc
```diff
@@ -0,0 +1,65 @@
...
+        {
+            {
+                GegenbauerPolynomial gp0(2, 0.0);
+                GegenbauerPolynomial gp1(3, 0.5);
+                GegenbauerPolynomial gp2(4, 1.5);
+
+                {
+                    const double z [5] = {-1.0, -0.3, 0.0, 0.7, 1.0};
+                    double g [3];
+                    const double v [3][5] = {{1.0, -0.82, -1.0, -0.02, 1.0},
```
dvandyk, line 48: Coding style: ``v [3]`` -> ``v[3]``, similarly above for ``z``.
Coding style:
```C++
    const double v[3][5] = {
        {  1.0, -0.8200, -1.0, -0.02,        1.00 },
        { -1.0,  0.3825,  0.0, -1.5335625,  15.00 },
        ...
    };
```

dvandyk (changes requested):
Thank you @vikusc! I have a lot of coding style comments, please have a look.
Hope to be able to merge it after this round of modifications!

eos/form-factors/k-lcdas.cc
```diff
@@ -269,9 +270,11 @@ namespace eos
     AntiKaonLCDAs::phi(const double & u, const double & mu) const
     {
         // Gegenbauer polynomials C_n^(3/2)
-        const double x = 2.0 * u - 1.0, x2 = x * x;
-        const double c1 = 3.0 * x;
-        const double c2 = (15.0 * x2 - 3.0) / 2.0;
+        const GegenbauerPolynomial g1(1, 3.0 / 2.0);
```
dvandyk, line 273: Hm. I would move this line into the constructor and provide all needed Gegenbauer polynomials as a member variable, e.g.
```C++
class PionLCDAs :
   ...
{
    const GegenbauerPolynomial gp_3o2_1;
    
    
    double
    PionLCDAs
};

PionLCDAs::PionLCDAs(...) :
    ...
    gp_3o2_1(1, 3.0 / 2.0),
    ...
{
    ...
}

double
PionLCDAs::phi(const double & u, const double & mu) const
{
    ...
    const double c1 = gp_3o2_1.evaluate(x), c2 = gp_3o2_2.evaluate(x);
    ...
}
```
In this way, the computationally expensive part (determining all coefficients) is done only once, while the fast part is called whenever is ``phi`` is called.

dvandyk (changes requested):
Almost there! See my single comment on where to store the polynomial objects.

dvandyk (approved):

Issue #680, opened 2023-08-15
[python] Task 'corner plot' is not documented


pluegh:
As of now, the corner plot task is not in the documentation:

https://github.com/eos/eos/blob/e1beea8c83b5e42353a556b0b2a544eeb298b9b6/doc/api/python.rst.jinja#L112-L125

Either a line should be added (quicker) or we should figure out to generate the list based on the code (more sustainable).

PR #692, opened 2023-09-12
[doc] Document eos.tasks functions automatically

pluegh:
Solves #680

### DIFF
--- a/doc/api/python.rst.py
+++ b/doc/api/python.rst.py
@@ -36,9 +36,14 @@ for plot_key, plot_class in reg.items():
         # Do not show in docs when no doc string is present
         pass
 
+# Document eos.tasks automatically
+excluded_tasks = ['run']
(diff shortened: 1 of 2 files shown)


mreboud (approved):
Perfect, thanks!

doc/api/python.rst.py
```diff
@@ -36,9 +36,15 @@
         # Do not show in docs when no doc string is present
         pass
 
+# Document eos.tasks automatically
+# Excluded tasks must be handled explicitly
+excluded_tasks = ['find_mode', 'run']
```
mreboud, line 41: Any reason why you exclude `find_mode` as well?

pluegh: Because it is treated explicitly in the documentation.
Why this is the case, I am not sure! There is not special commentary, so it might just be a remnant of history.

mreboud: As discussed in the group meeting: Can you merge the two `tasks` section since `Mode Finding` does not contain any additional information?

PR #698, opened 2023-09-27
Split ``_eos.cc`` into multiple files

### LINKED ISSUE #689: Split `_eos.cc`
The `_eos.cc` file starts to be a bit messy and would benefit from being split. I would keep the boost instructions in the file and move the rest in another file.

dvandyk:
- moved ``eos.version`` into separate files
- moved miscellaneous wrappers into separate files
- moved ``ExternalLogLikelihoodBlock`` into separate files

Github: resolves #689


mreboud (approved):
Thanks @dvandyk! If it compiles you can merge!

PR #704, opened 2023-10-24
Expand ``ExpressionParser`` to handle undeclared parameters gracefully

### LINKED ISSUE #701: ``ExpressionKinematicReader`` assumes that all parameters are available through ``Parameters::Defaults()``
When working on ANALYSIS-2023-09 with @smeiser, I encountered a problem.
Defining a number of new parameters and new expression observables that use these parameters, EOS shows me 
```
----> 1 eos.Observables().insert(
      2     'aux::R(pi)@Tevatron', R'(R(\pi))',
      3     eos.Unit.Unity(), eos.Options(),
      4     '''
      5         <<hadronization::fs/fd@Tevatron>>
      6         * <<B_s^0->D_s^-pi^+::BR@Parametric>> / <<B^0->D^-pi^+::BR@Parametric>>
      7         * 4.40 * <<D^-->K^+pi^-pi^-::BR@Parametric>> / <<D_s^-->phipi^-::BR@Parametric>>
      8     '''
      9 )

RuntimeError: Unknown Observable Error: Expression 'hadronization::fs/fd@Tevatron' is neither a known Observable nor a Parameter
```

Possible solutions:
 - turn ``eos.Parameters.declare`` into a static method that amends the default set of parameter at run time;
 - or modify the end of https://github.com/eos/eos/blob/118b5ebe6829eeb6d13b114f1d7fdbf35b5f5724/eos/utils/expression-visitors.cc#L436-L488 to avoid throwing an exception and emitting a warning instead.

dvandyk:
- [x] Disallow ``[[`` and ``]]`` in qualified names
 - [x] Add classes ``ParameterNameExpression`` and ``ParameterExpression``
 - [x] Let ``ExpressionParser`` handle parameter names as, e.g., ``[[mass::c]]``
 - [x] Test modifications

Github: resolves #701

### DIFF
--- a/eos/utils/expression-evaluator.hh
+++ b/eos/utils/expression-evaluator.hh
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Méril Reboud
+ * Copyright (c) 2023 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -36,6 +37,10 @@ namespace eos::exp
             double visit(ObservableNameExpression &);
 
             double visit(ObservableExpression & e);
+
+            double visit(ParameterNameExpression &);
+
+            double visit(ParameterExpression & e);
 
             double visit(KinematicVariableNameExpression & e);
 
--- a/eos/utils/expression-kinematic-reader.hh
+++ b/eos/utils/expression-kinematic-reader.hh
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Méril Reboud
+ * Copyright (c) 2023 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -46,6 +47,10 @@ namespace eos::exp
             void visit(const ObservableNameExpression & e);
 
             void visit(const ObservableExpression & e);
+
+            void visit(const ParameterNameExpression &);
+
+            void visit(const ParameterExpression &);
 
             void visit(const KinematicVariableNameExpression & e);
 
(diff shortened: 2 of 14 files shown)


eos/utils/expression-evaluator.hh
```diff
@@ -37,6 +38,10 @@ namespace eos::exp
 
             double visit(ObservableExpression & e);
 
+            double visit(ParameterNameExpression &);
```
mreboud, line 41: For consistency, a `e` could be added here.

eos/utils/expression-kinematic-reader.hh
```diff
@@ -47,6 +48,10 @@ namespace eos::exp
 
             void visit(const ObservableExpression & e);
 
+            void visit(const ParameterNameExpression &);
```
mreboud, line 51: Ditto

mreboud (commented):
I think everything is ok. Maybe a further test would be to use a parameter that does not exist yet?

eos/utils/expression-evaluator.hh
dvandyk: Actually, I disagree. ``ExpressionEvaluator::visit(ParameterNameExpression &)`` does not use its argument, hence it should be unnamed to avoid the warning that the argument is not used.

dvandyk:
> I think everything is ok. Maybe a further test would be to use a parameter that does not exist yet?

That is a very good idea!

eos/utils/expression-evaluator.hh
mreboud: You will probably not get a Warning since it is in the header file, but ok :)

dvandyk:
> > I think everything is ok. Maybe a further test would be to use a parameter that does not exist yet?
> 
> That is a very good idea!

Implemented, and it passes the checks.

Issue #702, opened 2023-10-11
Create parameter reset function for `eos.analysis` object


lorenzennio:
Given an `eos.analysis` object, it would be useful to have a function to reset its parameters. 
@dvandyk suggested the solution:
```python
for d in eos.Parameters()
    analysis.parameters.set(d.name(), d.evaluate())
```

dvandyk:
@lorenzennio Can we get a PR for this? :smile:

lorenzennio:
@dvandyk sure. I am in Marseille at the moment though, so it might take a few days (:

dvandyk:
Resolved by PR #705.

PR #705, opened 2023-10-26
Solve issue #702 - Create parameter reset function for ``eos.analysis`` object

lorenzennio:


python/eos/analysis.py
```diff
@@ -282,10 +282,17 @@ def _perplexity(weights):
...
         return pypmc.tools.convergence.ess(weights)
+    
 
     def clone(self):
         """Returns an independent instance of eos.Analysis."""
         return eos.Analysis(**self.init_args)
+    
+    
+    def reset(self):
+        """Resets the analysis parameters to its initial state."""
```
mreboud, line 293: Can we call it reset_parameters and describe it as `Resets the analysis parameters to their default values.`?

mreboud (changes requested):
Thanks @lorenzennio!
Looks good, I would just change the title and description a bit :)

dvandyk (approved):

Issue #622, opened 2023-03-28
Add loop level of QCD expressions as a parameter


pluegh:
In some cases, the expressions in `eos/utils/qcd.{hh,cc}` are needed to a certain loop order (not as high as possible). Currently we have to duplicate code for those cases. Implementing the loop level as a parameter will improve this.

PR #706, opened 2023-10-27
Solve issue #622 - add loop order option to QCD expressions

vikusc:


vikusc:
There seems to be a bug in

 https://github.com/eos/eos/blob/118b5ebe6829eeb6d13b114f1d7fdbf35b5f5724/eos/utils/qcd.cc#L163

and

https://github.com/eos/eos/blob/118b5ebe6829eeb6d13b114f1d7fdbf35b5f5724/eos/utils/qcd.cc#L171

compared to [CKS:2000] eq. 10 p. 6 a factor of 1/3 is missing

eos/utils/qcd.cc
```diff
@@ -99,14 +99,26 @@ namespace eos
...
+        {
+            throw InternalError("QCD::alpha_s: zero loop order not implemented");
+        }
+        if (loop_order > 4)
+        {
+            throw InternalError("QCD::alpha_s: loop order " + stringify(loop_order) + " not yet implemented");
+        }
+
         double a = alpha_s_0 / M_PI;
+        std::array<int, 4> switches;
```
dvandyk, line 114: I would turn the switches into ``std::array<double, 4>``, to avoid having to case the integers to double on every use.

eos/utils/qcd.cc
```diff
@@ -99,14 +99,26 @@ namespace eos
...
+        std::generate(switches.begin(), switches.end(), [n = 0, l = loop_order] () mutable { return (n++ < l) ? 1 : 0;});
```
dvandyk, line 115: See above, I would turn the literals to ``1.0`` and ``0.0``, respectively.

eos/utils/qcd.cc
```diff
@@ -130,25 +142,36 @@ namespace eos
...
+        }
+        if (loop_order > 4)
+        {
+            throw InternalError("QCD::alpha_s: loop order " + stringify(loop_order) + " not yet implemented");
+        }
+
         double a_mu0 = alpha_s_0 / M_PI;
         double a_mu = alpha_s_mu / M_PI;
+        std::array<int, 4> switches;
+        std::generate(switches.begin(), switches.end(), [n = 0, l = loop_order] () mutable { return (n++ < l) ? 1 : 0;});
```
dvandyk, line 159: See above.

dvandyk (commented):
Excellent work, thank you @vikusc! We still have to figure out which test cases fail and why.

dvandyk:
> There seems to be a bug in the last line of
> [...]

Good catch. I would deal with this in a separate commit, which can be easily handled as part of this PR.

vikusc:
With the bugfix the tests from the qcd class pass, but standard_model_TEST and wet_TEST fail now.

dvandyk:
> With the bugfix the tests from the qcd class pass, but standard_model_TEST and wet_TEST fail now.

Right, they need to be amended to show the correct values. There is a bit of a hack involved. You can run the tests after compiling them with ``#define EOS_GENERATE_TESTS 1`` **before** including ``<test/test.hh>``. The test will not actually check thing. Instead, it will output the various ``TEST_*`` macros with the correct values in place. This requires that the macros are called with the correct ordering of their arguments, though. Give it a try?

vikusc:
@dvandyk should we also implement the looporder like it was done in RunDec? At this point in the commit, we set the parameters of the beta and gamma functions to zero which correspond to a higher looporder than wanted. In RunDec all coefficients are set to zero which are influenced by higher looporders

mreboud:
@vikusc prepared a few fixes and pushed them to https://github.com/vikusc/eos/pull/new/loopoptionsMeril. I will find the time to continue his work.

mreboud:
Moved to #785

Issue #703, opened 2023-10-14
Add documentation for tags


mesmith75:
Would it be possible to add some documentation for the available tags for various processes? Just what's available and which papers they reference. For example for `BToKstarDilepton` - I had to search the code to find what I can use.

PR #723, opened 2023-11-10
Issue703

mreboud:
@dvandyk Can you have a look at the description and tell me what I should had? I think that adding a notebook that reproduces the B -> K ll predictions is necessary if we want to make this clear...

Github: Solves #703

### DIFF
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -23,7 +23,7 @@ need to install the ``virtualenv`` package, and then create a new virtual enviro
 
 
 The new virtual environment will now be **activated**.
-That means, the `python` (and `pip`) command will now execute the local binary instead of a system-wide one.
+That means, the ``python`` (and ``pip``) command will now execute the local binary instead of a system-wide one.
 At any later point, you can (re-)activate the virtual environment by running the following command:
 
 ::
@@ -119,7 +119,7 @@ If you built and installed from the source code
 and the import fails, i.e. you see ``ModuleNotFoundError: No module named 'eos'``,
 the module was either not installed with ``make install`` or the ``PYTHONPATH`` is not configured correctly.
 
-If you installed via `pip`
+If you installed via ``pip``
 and the output description is incorrect or ``None``,
-you have another module installed that is also named `eos`.
+you have another module installed that is also named ``eos``.
 Mind that the package name for the ``pip3 install`` command is ``eoshep``.
--- a/doc/faq.rst
+++ b/doc/faq.rst
@@ -123,3 +123,43 @@ If you installed via ``pip``
 and the output description is incorrect or ``None``,
 you have another module installed that is also named ``eos``.
 Mind that the package name for the ``pip3 install`` command is ``eoshep``.
+
+.. _faq-tags:
+
+... choose which ``tag`` to use for :math:`b \to s \ell \ell` observables
+=========================================================================
+
+The contributions from non-local operators in :math:`b \to s \ell \ell` observables
+can be estimated with different theoretical approaches, usually not valid in the same energy range,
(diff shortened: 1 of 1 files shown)


doc/faq.rst
```diff
@@ -123,3 +123,40 @@ If you installed via ``pip``
...
+   * - ``BFS2004``
+     - Next-to-leading order calculation based on the framework of QCD factorization and valid at small dilepton masses.
+     - [BFS:2001A], [BFS:2004A]
+
+   * - ``GP2004``
+     - Calculation of the non-local contributions valid for large dilepton masses.
+     - [GP:2004A]
+
+   * - ``GvDV2020``
+     - Parametrization based on unitarity and analyticity valid for small dilepton masses.
```
dvandyk, line 161: Maybe add ``below the $\psi(2S)$ state``.

doc/faq.rst
```diff
@@ -123,3 +123,40 @@ If you installed via ``pip``
...
+   * - ``Naive``
+     - Leading-order calculation.
+     -
+
+   * - ``BFS2004``
+     - Next-to-leading order calculation based on the framework of QCD factorization and valid at small dilepton masses.
+     - [BFS:2001A], [BFS:2004A]
+
+   * - ``GP2004``
+     - Calculation of the non-local contributions valid for large dilepton masses.
```
dvandyk, line 157: Maybe: ``Calculation of the non-local contributions in a local OPE. This approach is expected to yield sensible results only for $q^2$-integrated observables above the open-charm threshold.``

doc/faq.rst
```diff
@@ -123,3 +123,40 @@ If you installed via ``pip``
...
+   * - ``tag``
+     - Description
+     - Reference
+
+   * - ``Naive``
+     - Leading-order calculation.
+     -
+
+   * - ``BFS2004``
+     - Next-to-leading order calculation based on the framework of QCD factorization and valid at small dilepton masses.
```
dvandyk, line 153: Maybe: ``Calculation of the non-local contribution based on the framework of QCD factorization and a perturbative treatment of the intermediate quark states. This approach is expected to yield sensible results only for $q^2 \ll 4 m_c^2$.``

doc/faq.rst
```diff
@@ -123,3 +123,40 @@ If you installed via ``pip``
...
+.. list-table::
+   :widths: 15 55 30
+   :header-rows: 1
+
+   * - ``tag``
+     - Description
+     - Reference
+
+   * - ``Naive``
+     - Leading-order calculation.
```
dvandyk, line 149: Hm, I disagree. ``Absence of any non-local contributions.``?

dvandyk (commented):
Thanks @mreboud. I left some comments.

doc/faq.rst
mreboud: Not really, I think we are still using `C9eff` and `C7eff` at leading order in alpha_s, see e.g. https://github.com/eos/eos/blob/f3c073bf2bb9a2aa2af7740ab4a905a68c2c90b8/eos/rare-b-decays/lambda-b-to-lambda1520-ll-naive.cc#L88-L89

dvandyk: The key is in the name of the class: ``c7eff`` and ``c9eff`` are the short-distance quantities, basically ``c7`` and ``c9`` modified to be invariant under the RGE to higher order in $\alpha_s$. As such, they are essential to complete the  local contributions but do not contain any non-local terms (despite arising from four-quark operators).

mreboud: Ah yes sorry I am an idiot

dvandyk (approved):

Issue #710, opened 2023-11-07
rescale-by-width plot option is not documented


MJKirk:
The Constraint plot class takes an optional key of `rescale-by-width` which is used in the examples but not documented

Should be added here
https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/python/eos/plot/plotter.py#L774-L811

mreboud:
Closed by #724

PR #724, opened 2023-11-10
Add documentation for the rescale-by-width plot option

MJKirk:
See #710

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -789,6 +789,10 @@ class Plotter:
            Must identify one of the observables known to EOS; see `the complete list of observables <../observables.html>`_.
            This is only mandatory in multivariate constraints, since these can constrain more than one observable simultaneously.
 
+        The following keys are optional:
+
+         * ``rescale-by-width`` (*bool*) -- Rescales binned constraints by the bin width. Defaults to false.
+
         Example:
 
         .. code-block::
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -791,7 +791,8 @@ class Plotter:
         The following keys are optional:
 
          * ``xrange`` (list of int) -- The interval in which the observable is plotted in the case of a multivariate constraint.
-         * ``rescale-by-width`` (*bool*) -- Rescales binned constraints by the bin width. Defaults to false.
+         * ``rescale-by-width`` (*bool*) -- Rescales binned constraints by the inverse of the bin width. This is often required 
+           to compare theory (integrated) predictions and experimental (averaged) measurements. Defaults to false.
 
         Example:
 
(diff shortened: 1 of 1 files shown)


MJKirk:
The xrange option could also be moved to my new subsection of optional parameters in the docstring now.

As well, as far as I can see, the current docstring is slightly wrong, in that it currently says it applies for multivariate constraints
https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/python/eos/plot/plotter.py#L784
while if you look at the code it would seem to apply to any time of constraint (although currently I don't think there are any plain Gaussian constraints which apply to multiple kinematic values which could use this option).

mreboud:
> The xrange option could also be moved to my new subsection of optional parameters in the docstring now.
> 
> As well, as far as I can see, the current docstring is slightly wrong, in that it currently says it applies for multivariate constraints
> 
> https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/python/eos/plot/plotter.py#L784
> 
> while if you look at the code it would seem to apply to any time of constraint (although currently I don't think there are any plain Gaussian constraints which apply to multiple kinematic values which could use this option).

I agree that this can be moved to your `optional` section, but I don't understand your second comment. In the case of a single Gaussian constraint, the xrange option would not mean much since there is only one kinematic specification. This option allows you to remove some of the measurements when plotting a constraint with many entries.

MJKirk:
> I don't understand your second comment. In the case of a single Gaussian constraint, the xrange option would not mean much since there is only one kinematic specification. This option allows you to remove some of the measurements when plotting a constraint with many entries.

Yeah, sorry, that is me not fully understanding the Constraints format. I was thinking the Multivariate constraints were just for constraints that  applied to multiple observables, not also multiple kinematic variable bins. So ignore

mreboud:
Well that's perfect, it means that this is not 100% clear and could be clarified somewhere in the documentation!

python/eos/plot/plotter.py
```diff
@@ -789,6 +789,10 @@ class Constraint(BasePlot):
            Must identify one of the observables known to EOS; see `the complete list of observables <../observables.html>`_.
            This is only mandatory in multivariate constraints, since these can constrain more than one observable simultaneously.
 
+        The following keys are optional:
+
+         * ``rescale-by-width`` (*bool*) -- Rescales binned constraints by the bin width. Defaults to false.
```
dvandyk, line 794: I think we should be more specific here. Does it rescale by the inverse of the bin width or the bin width? I think it is the former, to turn integrated measurements in averaged measurements, making it possible to show integrated measurements in a differential plot.

dvandyk (changes requested):

python/eos/plot/plotter.py
MJKirk: Better?

mreboud: Definitely! Can you also add something like: `This is often required to compare theory (integrated) predictions and experimental (averaged) measurements.`

MJKirk: Done

dvandyk:
@MJKirk Can you squash both commits into one please? We do not need two commits for one one-line change :smile:

MJKirk:
> @MJKirk Can you squash both commits into one please? We do not need two commits for one one-line change 😄

Microsoft charging by the commit now :stuck_out_tongue_winking_eye:?
Sure, will do

mreboud (approved):

Issue #715, opened 2023-11-07
Add warning not to set parameter values directly


MJKirk:
Add a warning somewhere around here
https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/examples/basics.ipynb#L86

that if you directly modify an eos parameter (e.g. like `m_mu = 1.779` rather than `m_mu.set(1.779)`) nothing immediately breaks, but now `m_mu` is just a number rather than an eos.Parameter

MJKirk:
Something like https://nbsphinx.readthedocs.io/en/0.9.3/markdown-cells.html#Info/Warning-Boxes

mreboud:
Certainly a good idea, let me see if I can do that.

mreboud:
Well, all well thought it is a bit obvious, no?
But if you think it is worth it, I suggest to add

> Caution: If you evaluate ``m_mu = 1.779``, you will just make ``m_mu`` a double instead of an ``eos.Parameter``

Let me know what you think @MJKirk @dvandyk !

MJKirk:
Perhaps it is obvious - I think I was thinking of the masses being properties of a class rather than a class object itself, and therefore you could directly set them using `=`. And I think the way the documentation pulls out `m_mu` as an individual variable encouraged me to play with it directly.

But I think my opinion would be that early on the documentation it doesn't hurt to state the obvious :)

mreboud:
Closed by #726

Issue #714, opened 2023-11-07
Explain parameters used in B->Dlnu theory uncertainty example


MJKirk:
In the [Estimating Theory Uncertainties](https://eos.github.io/doc/predictions.html#Estimating-Theory-Uncertainties) section of the docs, various B->D::alpha parameters are used, but with minimal explanation.
I propose to add brief text saying that they are parameters in some particular parametrisation of the form factors and maybe a link to the paper where they are defined (made marginally more complicated by the fact that they are called BSZ2015 parameters but this is a pseudoscalar decay and this definition is really from a later paper by Danny, not the BSZ 2015 paper)

mreboud:
Good point.

The name BSZ will certainly be changed into SSE (for Simplified Series Expansion) since we started to use it for almost every parameters ><

mreboud:
Closed by #726

PR #726, opened 2023-11-13
Solves issue 714 and 715

mreboud:
Solves #714
Solves #715

### DIFF
--- a/examples/basics.ipynb
+++ b/examples/basics.ipynb
@@ -101,6 +101,13 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "> Caution: If you evaluate ``m_mu = 1.779``, you will just make ``m_mu`` a double instead of an ``eos.Parameter``"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


MJKirk (approved):
Looks good to me!

mreboud:
> Looks good to me!

Thanks, I just remove the cells and python version modification in the notebook, and the we can merge. Unless @dvandyk has a comment?

dvandyk (changes requested):
I think the poles are incorrect in the description.

mreboud:
> I think the poles are incorrect in the description.

They should match BSZ's right?

image

dvandyk:
The form is OK, but the states are listed as ``D^*`` mesons:
image
They should be generic poles or (in this example) ``B_c`` states with a variety of spin/parity quantum numbers.
I would replace this with ``m_{R_i}^2`` and explain that the poles are listed in the respective papers.

mreboud:
> The form is OK, but the states are listed as `D^*` mesons: image They should be generic poles or (in this example) `B_c` states with a variety of spin/parity quantum numbers. I would replace this with `m_{R_i}^2` and explain that the poles are listed in the respective papers.

Ah sure sorry, I am an idiot

dvandyk (approved):

PR #730, opened 2023-11-18
Check physical phase space in B->3lnu observables

### LINKED ISSUE #729: B->l^- nu l'^+ l'^-: branching ratio can be evaluated outside the physical phase space
[First reported by Martino Borsato]

The comments and code in
https://github.com/eos/eos/blob/870d11b1b7cf07ab6ffbab708fbd2654ba2e6fff/eos/b-decays/b-to-3l-nu.cc#L91-L97
should be adjusted to respect the phase space boundary ``m_l^2 <= k2 <= (m_B - sqrt(q2))^2``.

SKuerten:
resolves #729

### DIFF
--- a/eos/b-decays/b-to-3l-nu.cc
+++ b/eos/b-decays/b-to-3l-nu.cc
@@ -90,8 +90,8 @@ namespace eos
 
         /*!
          * differential decay width of 2 kinematic variables
-         * q2 is the invariant mass of the off-shell photon in the range 4 m_l'^2 <= q2 <= (m_B - k2)^2
-         * k2 is the invariant mass of the W-meson in the range m_l^2 <= k2 <= (m_B -q2)^2
+         * q2 is the invariant mass of the off-shell photon in the range 4 m_l'^2 <= q2 <= (m_B - m_l )^2
+         * k2 is the invariant mass of the W-meson in the range m_l^2 <= k2 <= (m_B - sqrt(q2) )^2
          */
         double double_differential_decay_width(const double & q2, const double & k2) const
         {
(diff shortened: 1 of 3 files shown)


eos/b-decays/b-to-3l-nu.cc
```diff
@@ -90,8 +90,8 @@ namespace eos
 
         /*!
          * differential decay width of 2 kinematic variables
-         * q2 is the invariant mass of the off-shell photon in the range 4 m_l'^2 <= q2 <= (m_B - k2)^2
-         * k2 is the invariant mass of the W-meson in the range m_l^2 <= k2 <= (m_B -q2)^2
+         * q2 is the invariant mass of the off-shell photon in the range 4 m_l'^2 <= q2 <= (m_B - sqrt(k2) )^2
```
dvandyk, line 93: That does not reflect what we had written in [KKvDZ:2022A], which is: $4 m_{\ell'}^2 \leq q^2 \leq (m_B - m_\ell)^2$. $k^2$ seems to be correct.

eos/b-decays/b-to-3l-nu.cc
```diff
@@ -683,6 +689,12 @@ namespace eos
             - 2 * k2 * (m_B_sq + q2)) * z_w) + k2 * (-m_B_sq + m_l_sq - q2 + sqrt(k4
             + m_B_sq_q2_diff2 - 2 * k2 * (m_B_sq + q2)) * z_w)));
 
+            if ( power_of<2>(m_B - sqrt(k2)) < q2 || q2 < 4 * m_lprime_sq
+                || power_of<2>(m_B - sqrt(q2)) < k2 || k2 < m_l_sq )
```
dvandyk, line 693: See my earlier comment. This does not reflect what we have in the paper.

dvandyk (changes requested):

dvandyk:
Nitpicking on the commit messages. We typically do not use the file names in the commit messages, so ``B->3lnu`` rather than ``bto3lnu`` in the first commit message. Also, why did you split this into two commits? I think one would suffice.

SKuerten:
> Nitpicking on the commit messages. We typically do not use the file names in the commit messages, so `B->3lnu` rather than `bto3lnu` in the first commit message. Also, why did you split this into two commits? I think one would suffice.

You are just to quick. I was already fixing that.^^

eos/b-decays/b-to-3l-nu.cc
SKuerten: fixed

eos/b-decays/b-to-3l-nu.cc
SKuerten: done

eos/b-decays/b-to-3l-nu.cc
```diff
@@ -184,7 +184,13 @@ namespace eos
             + f25 * f_B * real(F_2) + f35 * f_B * real(F_3) + f45 * f_B * real(F_4)
             + f55 * power_of<2>(f_B);
 
-        return prefactor * amp;
```
dvandyk, line 187: Thanks for fixing this, would have driven me mad(der) :smile:

dvandyk (approved):

PR #736, opened 2023-11-23
Refactor parameters declare

dvandyk:
Summary of change of functionality:

 - default values are loaded at the first creation of a ``Parameter`` objects,
   and stored in the ``ParameterDefaults`` singleton;
 - ``Parameters::declare`` has become a static method returning ``void``;
 - ``Parameters::decalare_and_insert`` has been added to facilitate declaring
   a new parameter globally and inserting it into the current parameter set;
 - test cases and Python interface have been adjusted.

Request review from @lorenzennio for changes to the handling of parameters in ``python/eos/pyhf_likelihood.py``.

### DIFF
--- a/eos/utils/parameters.cc
+++ b/eos/utils/parameters.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2010-2022 Danny van Dyk
+ * Copyright (c) 2010-2023 Danny van Dyk
  * Copyright (c) 2021 Philip Lüghausen
  * Copyright (c) 2010 Christian Wacker
  *
@@ -22,6 +22,7 @@
 #include <config.h>
 
 #include <eos/utils/cartesian-product.hh>
+#include <eos/utils/instantiation_policy-impl.hh>
 #include <eos/utils/log.hh>
 #include <eos/utils/parameters.hh>
 #include <eos/utils/private_implementation_pattern-impl.hh>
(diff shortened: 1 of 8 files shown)


eos/utils/parameters.cc
```diff
@@ -640,9 +691,17 @@ namespace eos
...
     Parameters::declare(const QualifiedName & name, const std::string & latex, Unit unit, const double & value, const double & min, const double & max)
     {
+        ParameterDefaults::instance()->insert(name, Parameter::Template { name, min, value, max, latex, unit });
+    }
+
+    Parameter
+    Parameters::declare_and_insert(const QualifiedName & name, const std::string & latex, Unit unit, const double & value, const double & min, const double & max)
+    {
+        ParameterDefaults::instance()->insert(name, Parameter::Template { name, min, value, max, latex, unit });
+
```
mreboud, line 704: Shouldn't this appear after line 710, after the comment `// create new parameter`?

python/_eos.cc
```diff
@@ -211,9 +211,28 @@ BOOST_PYTHON_MODULE(_eos)
...
+            :param latex: The LaTeX representation for the parameter.
+            :type latex: str
+            :param unit: The unit in which the parameter is expressed.
+            :type unit: eos.Unit
+            :param value: The initial value for the parameter.
+            :type value: float
+            :param min: The minimum value that the parameter can attain.
+            :type min: float
+            :param max: The maximum value that the parameter can attain.
+            :type max: float
```
mreboud, line 229: Shouldn't these specifications be optional?

mreboud (approved):

python/_eos.cc
dvandyk: Hm, probably. How do you do default parameter values in boost-python?

eos/utils/parameters.cc
dvandyk: Well, this declares it but does not insert it yet. I will adapt the comments, which were taken without modification from the previous code.

python/_eos.cc
mreboud: With colons? https://www.boost.org/doc/libs/1_83_0/libs/parameter_python/doc/html/index.html#id10

eos/utils/parameters.cc
mreboud: But the point was that the parameter might already have been declared.

lorenzennio:
Actually, one of my tests just failed:

```bash
======================================================================
ERROR: test_multivariate_priors_2 (__main__.ClassMethodTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/lorenz/eos/python/./eos/analysis_TEST.py", line 235, in test_multivariate_priors_2
    results  = dynesty.utils.merge_runs(results_list)
  File "/home/lorenz/venv_eos/lib/python3.10/site-packages/dynesty/utils.py", line 1750, in merge_runs
    res = _merge_two(rlist_base[0], rlist_base[1], compute_aux=True)
  File "/home/lorenz/venv_eos/lib/python3.10/site-packages/dynesty/utils.py", line 1980, in _merge_two
    if np.all(base_info['bounds'] == new_info['bounds']):
ValueError: operands could not be broadcast together with shapes (17,2) (16,2) 

----------------------------------------------------------------------
Ran 7 tests in 31.456s

FAILED (errors=1)
FAIL eos/analysis_TEST.py (exit status: 1)
```
It might just be an incompatibility with my `dynesty` version or something (2.1.3). Might be worth double-checking for one of you. If needed, I can investigate more (:

eos/utils/parameters.cc
dvandyk: Ahh. Good point. Something is clearly going awry when we use this method to ``declare_and_insert`` an already available parameter, contrary to the pure ``declare`` which might simply override a different analysis file's parameter declarations. I will deal with it.

python/_eos.cc
dvandyk: Nope. https://boostorg.github.io/python/doc/html/tutorial/tutorial/functions.html#tutorial.functions.default_arguments.boost_python_member_function_ove , I am not a fan...

Since this is a low-level function that should ideally not be used by users directly, I think we should make callers of ``eos.Parameters.declare_and_insert`` provide all arguments in the Python API.

dvandyk:
> Actually, one of my tests just failed:
> 
> ```shell
> ======================================================================
> ERROR: test_multivariate_priors_2 (__main__.ClassMethodTests)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/lorenz/eos/python/./eos/analysis_TEST.py", line 235, in test_multivariate_priors_2
>     results  = dynesty.utils.merge_runs(results_list)
>   File "/home/lorenz/venv_eos/lib/python3.10/site-packages/dynesty/utils.py", line 1750, in merge_runs
>     res = _merge_two(rlist_base[0], rlist_base[1], compute_aux=True)
>   File "/home/lorenz/venv_eos/lib/python3.10/site-packages/dynesty/utils.py", line 1980, in _merge_two
>     if np.all(base_info['bounds'] == new_info['bounds']):
> ValueError: operands could not be broadcast together with shapes (17,2) (16,2) 
> 
> ----------------------------------------------------------------------
> Ran 7 tests in 31.456s
> 
> FAILED (errors=1)
> FAIL eos/analysis_TEST.py (exit status: 1)
> ```
> 
> It might just be an incompatibility with my `dynesty` version or something (2.1.3). Might be worth double-checking for one of you. If needed, I can investigate more (:

Yes, this looks like a problem with ``dynesty``. We have pinned it to ``==2.0.3`` in the Docker images.
Let me open a new issue to figure out what the fuck is going on there.

PR #741, opened 2023-11-29
Some fixes

mreboud:


MJKirk (commented):
I believe if you use `:meth:` here rather than `:code:`, you get a hyperlink to the relevant part of the API. Similarly to what is done on line 93 with `:class:eos.Options`

.github/workflows/pypi-build+check+deploy.yaml
MJKirk, line 1: Looks fine to me (in the sense I think it corresponds to what was done before [this commit](https://github.com/eos/eos/commit/868a426c067d213f05ee2cefb55f6a03e8701e75) ) but I'm not an expert in the github workflow syntax to be 100%

PR #742, opened 2023-11-29
[rare-b-decays] implement scalar and tensor sectors for b-to-vec-nu-nu

lorenzennio:
Extend the b-to-vec-nu-nu differential decay width for the scalar and tensor sector, as in eq. 13 of https://arxiv.org/pdf/2111.04327.pdf

### DIFF
--- a/eos/rare-b-decays/b-to-vec-nu-nu.cc
+++ b/eos/rare-b-decays/b-to-vec-nu-nu.cc
@@ -2,6 +2,7 @@
 
 /*
  * Copyright (c) 2021 Danny van Dyk
+ * Copyright (c) 2023 Lorenz Gaertner
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -65,6 +66,7 @@ namespace eos
 
         std::function<complex<double> ()> lambda_t;
         std::function<WilsonCoefficients<wc::SBNuNu> ()> wc;
+        std::function<double ()> m_D;
 
         GSL::QAGS::Config int_config;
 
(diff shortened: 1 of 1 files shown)


eos/rare-b-decays/b-to-vec-nu-nu.cc
```diff
@@ -173,27 +173,43 @@ namespace eos
         {
             const double m_B = this->m_B(), m_B2 = m_B * m_B;
             const double m_V = this->m_V(), m_V2 = m_V * m_V;
+            const double m_b = model->m_b_msbar(mu);
+            const double m_s = model->m_s_msbar(mu);
```
dvandyk, line 177: Can you please create a new member of the Implementation class, ``std::function<double (const double &)> m_D`` and assign it a lambda returning ``model->m_s_msbar(mu);``, between lines 160 and 161. Makes it easier if we want to implement ``dbnunu`` decays in the future :smile:

dvandyk (changes requested):
OK except for one nitpicky thingy that ensures that we do not miss a strange-quark mass in a future modification.

dvandyk:
> OK except for one nitpicky thingy that ensures that we do not miss a strange-quark mass in a future modification.

Oh, also, can you fix the commit message please? See [the wiki](https://github.com/eos/eos/wiki/How-to-Contribute), Point 6. E.g.:
```
[rare-b-decays] Account for scalar and tensor contribution in B->Vnunu observables
```

eos/rare-b-decays/b-to-vec-nu-nu.cc
lorenzennio: Good idea. I will rename `m_s` to `m_D` everywhere in the equation implementation as well. I will add this in a new commit and we can squash later when we are happy.

lorenzennio:
@dvandyk let me know if you are happy. Then I will squash the commits.

dvandyk:
> @dvandyk let me know if you are happy. Then I will squash the commits.

Looks perfect, go ahead and squash, so that we can merge it asap. (Hopefully in time for the talk!)

PR #744, opened 2023-12-01
Fixes hyperlinks from constraints to observables/parameters

### LINKED ISSUE #708: Broken hyperlinks in list of constraints
Some of the hyperlinks in https://eos.github.io/doc/constraints.html are broken.

There are two different potential reasons:

1) The conversion from the EOS name to the html id is wrong
1) They should link to the list of parameters, not the list of observables

One example of the first case:
The fourth constraint in the table links to `B->D^*::A_2(q2)`, which is an observable but the generated link id is `bmmtod^*cocoaun2poq2pc` while the correct id (by which I mean the one on the list of observables page) is `bmmtod-cocoaun2poq2pc`.
It is not immediately obvious to me from either [doc/observables.rst.py](https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/doc/observables.rst.py) or [doc/constraints.rst.py 
](/https://github.com/eos/eos/blob/b639be8d351148c34f632c5877dfd89952a1283c/doc/constraints.rst.py) how/why the observable names get mangled differently. 

An example for the second case:
The sixth constraint in the table links to `B->D::alpha^fT_2@BSZ2015`, which is a parameter rather than an observable, so the hyperlink should point to `doc/parameters.html#bmmtodcocoalpha^ftun2atbsz2015` rather than `doc/observables.html#bmmtodcocoalpha^ftun2atbsz2015`. 
Although this also doesn't work, as the correct id should be `bmmtodcocoalpha-ftun2atbsz2015` (so this looks related to the first case, where an `^` is being converted to `-` in some places and not others.

MJKirk:
Adding mangling for asterisks and carets in EOS QualifiedNames, which means the hyperlinks generated in the constraints list match the HTML id's in the parameters and observables list.

Also now checks if the links in the constraint page are to observables or parameters, and makes the link accordingly

Fixes #708

### DIFF
--- a/doc/constraints.rst.py
+++ b/doc/constraints.rst.py
@@ -1,16 +1,10 @@
 import eos
 import re
 import yaml
-from jinja_util import print_template
+from jinja_util import print_template, qn_to_link_map
 
 def latex_to_rst(s):
     return(re.sub(r'\$([^\$]*)\$', r':math:`\1`', s))
-
-qn_to_link_map = {
-    ord(':'): 'co', ord('@'): 'at', ord('/'): 'sl', ord('_'): 'un',
-    ord('('): 'po', ord(')'): 'pc', ord('+'): 'pp', ord('-'): 'mm',
-    ord('>'): 'to', ord('^'): 'ca', ord('*'): 'as'
-}
 
 def make_constraints():
     # All observables (pseudo or otherwise), to be used to see if a constraint is an observable or a parameter
--- a/doc/constraints.rst.py
+++ b/doc/constraints.rst.py
@@ -13,6 +13,9 @@ qn_to_link_map = {
 }
 
 def make_constraints():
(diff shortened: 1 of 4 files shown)


dvandyk (changes requested):
Thank you @MJKirk, this looks really good! I have only one nitpicky request: can you move the name-mangling to the existing ``doc/jinja_util.py``? I want to avoid creating an adhoc python module in ``doc/`` :smile:

MJKirk:
> Thank you @MJKirk, this looks really good! I have only one nitpicky request: can you move the name-mangling to the existing `doc/jinja_util.py`? I want to avoid creating an adhoc python module in `doc/` 😄

Yeah that fine, I did think about putting it in that file but wasn't sure if it fit.

dvandyk (approved):

Issue #748, opened 2023-12-05
Confusing error when trying to create undefined `SignalPDF`


lorenzennio:
Steps to reproduce:
`eos.SignalPDF.make('blah::blah', eos.Parameters(), eos.Kinematics(), eos.Options())`

Error message:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/lorenz/venv_eos/lib/python3.10/site-packages/eos/signal_pdf.py", line 125, in make
    pdf.__class__ = SignalPDF
TypeError: __class__ assignment only supported for mutable types or ModuleType subclasses

```
It would be nice to catch this sooner and give an informative error message.

Issue #733, opened 2023-11-21
Broken link in `SignalPDF` documentation


lorenzennio:
The file ``signal-pdfs.html``, linked in the documentation, seems to be missing.
https://github.com/eos/eos/blob/868a426c067d213f05ee2cefb55f6a03e8701e75/python/eos/signal_pdf.py#L112

Issue #610, opened 2023-03-06
Refactoring of `signal_pdf`


mreboud:
- [ ] Move `signal_pdf.{cc,hh}` to a dedicated directory
- [ ] Move `make_signal_pdf_entries()` to a separate file and maybe split it into subfiles corresponding to the decay type
- [ ] Harmonize the kinematic names

@dvandyk what do you think?
@pluegh you wanted to look at this?

dvandyk:
I agree that this needs some tidying up. But I would like to structure it similar to what he have for the observables, i.e.:
  - self-documenting;
  - store close to the code (i.e., ``B->Dlnu::P(q2)`` should be in ``eos/b-decays/signal-pdf.cc``.
  
For that purpose, we have would need
 - [ ] ``SignalPDFSection``: contains one or more groups of signal PDFs, e.g., all signal PDFs within ``eos/b-decays``;
 - [ ] ``SignalPDFGroup``: contains one or more signal PDF entries, e.g., all signal PDFs corresponding to the process ``B->Dpilnu``;
 - [ ] ``SignalPDFEntry``: describes a single signal PDF;
 - [ ] ``SignalPDFs``: contains all sections, is accessible by the user, can iterate over all sections and all entries.
 
 A lot of this is copy&paste from ``eos/observable.{hh,cc}`` and find&replace.

PR #749, opened 2023-12-07
Fix three issues related to signal PDFs

dvandyk:
- refactors the ``SignalPDFs`` code to reflect what we have been doing for ``Observables`` (#610)
- add list of known signal PDFs to the documentation (#733)
- handle non-existing PDFs gracefully in the factory method (#748)

### DIFF
--- a/eos/Makefile.am
+++ b/eos/Makefile.am
@@ -26,7 +26,7 @@ libeos_la_SOURCES = \
 	constraint.cc constraint.hh \
 	observable.cc observable.hh observable-fwd.hh observable-impl.hh \
 	reference.cc reference.hh \
-	signal-pdf.cc signal-pdf.hh
+	signal-pdf.cc signal-pdf.hh signal-pdf-fwd.hh signal-pdf-impl.hh
 libeos_la_CXXFLAGS = $(AM_CXXFLAGS) \
 	-DEOS_DATADIR='"$(datadir)"' \
 	$(GSL_CXXFLAGS) \
@@ -57,7 +57,8 @@ AM_TESTS_ENVIRONMENT = \
 TESTS = \
 	constraint_TEST \
 	observable_TEST \
-	reference_TEST
+	reference_TEST \
+	signal-pdf_TEST
 
 LDADD = \
 	$(top_builddir)/test/libeostest.la \
@@ -73,7 +74,8 @@ LDADD = \
 check_PROGRAMS = \
 	constraint_TEST \
 	observable_TEST \
-	reference_TEST
+	reference_TEST \
+	signal-pdf_TEST
 
 constraint_TEST_SOURCES = constraint_TEST.cc
 constraint_TEST_CXXFLAGS = $(AM_CXXFLAGS) $(GSL_CXXFLAGS)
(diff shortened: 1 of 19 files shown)


eos/Makefile.am
```diff
@@ -73,7 +74,8 @@ LDADD = \
 check_PROGRAMS = \
 	constraint_TEST \
 	observable_TEST \
-	reference_TEST
+	reference_TEST \
+	signal-pdf_TEST
```
mreboud, line 78: Unrelated to this PR: Why does this list appear twice?

eos/b-decays/signal-pdfs.cc
```diff
@@ -0,0 +1,453 @@
...
+                make_signal_pdf("B->gammalnu::d^2Gamma/dEgamma/dcos(theta_l)",
+                    Options{ },
+                    &BToGammaLeptonNeutrino::fully_differential_decay_width,
+                    std::make_tuple(
+                        KinematicRange{ "Egamma", 0.1, 2.64, BToGammaLeptonNeutrino::kinematics_description_Egamma },
+                        KinematicRange{ "cos(theta_l)", -1.0, +1.0, BToGammaLeptonNeutrino::kinematics_description_c_theta_l}
+                    ),
+                    &BToGammaLeptonNeutrino::integrated_branching_ratio,
+                    std::make_tuple(
+                        "E_gamma_min"
```
mreboud, line 54: Can we use the same convention `E_gamma` or `Egamma` throughout EOS?

eos/signal-pdf.cc
```diff
@@ -19,22 +19,13 @@
...
-#include <eos/rare-b-decays/b-to-kstar-ll.hh>
-#include <eos/rare-b-decays/b-to-psd-nu-nu.hh>
-#include <eos/rare-b-decays/b-to-vec-nu-nu.hh>
+#include <eos/b-decays/signal-pdfs.hh>
+#include <eos/rare-b-decays/signal-pdfs.hh>
 #include <eos/utils/density.hh>
+#include <eos/utils/instantiation_policy-impl.hh>
+#include <eos/utils/log.hh>
 #include <eos/utils/private_implementation_pattern-impl.hh>
 #include <eos/utils/wrapped_forward_iterator-impl.hh>
```
mreboud, line 31: Can you remove lines 55 - 56 below? In the function `norm`, the `using std::pow` is not needed.

python/eos/signal_pdf.py
```diff
@@ -167,11 +167,88 @@ def filter_entry(self, qn):
...
+                group_entries = 0
+                group_result  = fr'''
+                    <tbody>
+                        <tr>
+                            <th style="text-align:left" colspan=1>
+                                <a style="text-decoration: none" onclick="toggle_group(this, 'grp{group_id}')">{group.name()}</a>
+                            </th>
+                        </tr>
+                    </tbody>
+                '''#.format(group=group.name(), id=group_id)
```
mreboud, line 231: This comment might be removed, maybe?

mreboud (approved):
I am happy with the changes, I haven't check that the documentation looked nice, but I guess you did. Thanks for the work!

eos/Makefile.am
dvandyk: Ah, a peculiarity of automake. It has two targets: ``check-PROGRAMS`` and ``check-TESTS``. The former build the list of all programs (``check_PROGRAMS``) that must be available to the tests. The latter run all the tests in ``TESTS``.

eos/signal-pdf.cc
dvandyk: Done.

python/eos/signal_pdf.py
dvandyk: Done.

eos/b-decays/signal-pdfs.cc
dvandyk: Yes. That should be done more systematically though. I have opened #750 and will resolve this conversation.

Issue #750, opened 2023-12-11
Check names of kinematic variables against a whitelist


dvandyk:
The test cases for ``Observable`` and ``SignalPDF`` should iterate over their entries' kinematic variables and flag any variable names that are not whitelisted.

Suggestion for the initial whitelist:
 - ``q2``, ``k2``, etc.
 - ``E_gamma``, ``E_pi``, ``E_K``, etc.
 - ``cos(theta_l)``, ``cos(theta_pi)``, ``cos(theta_K)``, etc.
 - ``phi``
 
+ names that start with any of the above and have a ``_min`` or ``_max`` appended.

Issue #747, opened 2023-12-03
Switch to using static polynomial objects where possible


dvandyk:
Adapt ``GegenbauerPolynomial``:
- [ ] check if the exponent ``alpha`` can be made ``constexpr``
- [ ] check if we can template the class with respect to the order parameter.
- [ ] adjust all usages of ``GegenbauerPolynomial`` in the code.

dvandyk:
> * [ ]  check if the exponent `alpha` can be made `constexpr`
> * [ ]  check if we can template the class with respect to the order parameter.

Both of these should be possible, since as off C++20 non-type template parameter types can be of floating point type!

Issue #693, opened 2023-09-13
Unify importing of optional Python in ``eos.tasks``


dvandyk:
Right now we depend on a few optional python packages, which are imported only when needed within a task.
If unavailable, an error message is raised. This is a bit untidy, and it would be nice if we could list the names of optional
python packages used by a task within its ``@task`` decorator. The latter should then handle the importing (via ``importlib``)
and raise errors uniformly as appropriate.

PR #756, opened 2024-01-01
Changes for master

dvandyk:
Merges the following changes:

- use ``analysis.yaml`` instead of the (hidden) ``.analysis.yaml`` as the default name for an analysis files in ``eos-analysis``
- run ``pyupgrade`` on ``python/eos/tasks.py``
- provide means to specify optional required modules for every task (#693)
- add a new subcommand ``eos-analysis init``, which install the contents of [eos/analysis-template](https://github.com/eos/analysis-template) into the current working directory
- check the names of kinematic variables in all observables against a whitelist (#750)
- improve the usage of ``GegenbauerPolynomial`` in classes ``PionLCDAs``, ``AntiKaonLCDAs``, and ``KaonLCDAs`` (in furtherance of addressing #747)

### DIFF
--- a/eos/observable_TEST.cc
+++ b/eos/observable_TEST.cc
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2021 Méril Reboud
+ * Copyright (c) 2021-2023 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -19,6 +20,9 @@
 #include <eos/observable.hh>
 #include <eos/utils/options.hh>
 #include <eos/utils/units.hh>
+
+#include <ranges>
+#include <iostream>
 
 using namespace test;
 using namespace eos;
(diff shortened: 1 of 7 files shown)


python/eos/tasks.py
```diff
@@ -48,10 +48,22 @@ def __exit__(self, type, value, traceback):
 
 _tasks = {}
 
-def task(name, output, mode=lambda **kwargs: 'w'):
+def task(name, output, mode=lambda **kwargs: 'w', modules=[]):
```
mreboud, line 51: This can be applied to pypmc, dynesty,... right?

src/scripts/eos-analysis
```diff
@@ -663,5 +679,34 @@ def cmd_list_posteriors(args):
         print(name)
 
 
+# Initialize analysis directory
+def cmd_init(args):
```
mreboud, line 683: I don't understand where `base_directory` is used

eos/observable_TEST.cc
```diff
@@ -127,6 +131,60 @@ class ObservableTest :
                     )")
                 );
             }
+
+            /* Test the names of the kinematic variables against a whitelist */
+            {
```
mreboud, line 136: Shouldn't this list be part of another file? It is a bit weird to have it here, no?

eos/observable_TEST.cc
```diff
@@ -127,6 +131,60 @@ class ObservableTest :
...
+                static const std::set<std::string> allowed_kinematic_variables
+                {
+                    "q2", "q2_min", "q2_max",
```
mreboud, line 139: Shouldn't we better write a list without min and max, and try to parse min and max in the test?

eos/observable_TEST.cc
```diff
@@ -127,6 +131,60 @@ class ObservableTest :
...
+                    "cos(theta_k)",
+                    "phi",
+                    "E", "E_min", "E_max",
+                    "Re{E}", "Im{E}",
+                    "E_gamma", "E_gamma_min",
+                    "E_pi",
+                    "w", "w_min", "w_max", "w_min_num", "w_max_num", "w_min_denom", "w_max_denom",
+                    "mu", "tau",
+                    "z", "z_min", "z_max", "z_min_num", "z_max_num", "z_min_denom", "z_max_denom",
+                    // needs to be unified with the notation of Re{E} and Im{E}
```
mreboud, line 158: Agreed! Can you open a new issue?

mreboud (changes requested):
Thanks @dvandyk for these changes! I left comments.

python/eos/tasks.py
dvandyk: Correct. ``modules`` can contain a list of string, where each string corresponds to the name of a potential python module. Each module will then be imported and a consistent error message thrown if it is unavailable.

src/scripts/eos-analysis
dvandyk: Nowhere. When I kept it as an argument, I was considering adding it to the analysis file itself as an override. However, I noticed that this fails when people share a common analysis files across different machines. Hence, I left it for later modification. I can take it out if you like.

eos/observable_TEST.cc
dvandyk: I followed the instruction in the corresponding issue title. The whitelisting should be done in the test case, since it is a "soft" target. We want users to be able to create their own kinematic variables with arbitrary names; however, the EOS ones should be rather uniform. Put differently, we had agreed not to use a new class like ``KinematicVariableName`` or similar.

eos/observable_TEST.cc
dvandyk: Could be done. Does it help? Currently, not all variables follow the scheme ``"common_name" + "_min"``, e.g., ``w_min_denom``, which probably should be ``w_denom_min``.

We can adapt the test case once all offending instances of kinematic variable names have been removed?

eos/observable_TEST.cc
dvandyk: Will do!

src/scripts/eos-analysis
dvandyk: It has been removed.

python/eos/tasks.py
mreboud: Then we should certainly do it (maybe in another PR).

eos/observable_TEST.cc
mreboud: I agree

eos/observable_TEST.cc
mreboud: Sounds good

python/eos/tasks.py
dvandyk: Yes. This requires some work though. Currently, those modules are globally imported; those imports now need to be moved into the tasks / classes. I'll open an issue.

dvandyk:
> Can we merge this one @dvandyk?

I think so!

Issue #670, opened 2023-08-02
EOS should check the order of the parameters when importing samples


mreboud:
The order should match the one in the analysis file

mreboud:
Closed by #762

PR #762, opened 2024-01-12
Check analysis file parameters match those of a loaded sample

MJKirk:
Fixes issue #670


mreboud (changes requested):
Thanks @MJKirk!
Can you:
- Exchange the order of your two commits and rename the pre-commit one?
- Apply your test to the `corner-plot` task as well?

dvandyk (approved):
Agreed on the contents. The ``pre-commit`` commit can probably be dropped entirely, since I had already included that change in PR #750. (Please check! :pleading_face:)

PR #765, opened 2024-01-15
Updates for B to gamma ell nu

pluegh:
This PR implements the following changes:
 - [x] Improve efficiency by updating the form factor implementation with an alternative to numerical integration.
 - [x] Fix an inconsistent expression for the form factors (related to the twist expansion).


eos/form-factors/analytic-b-to-gamma-qcdf.cc
```diff
@@ -241,15 +241,23 @@ namespace eos
...
-        // cf. [BBJW:2018A], Eq. (4.14)
-        const double omega_cut = s_0 / (2.0 * Egamma);
-        const double sigma = 2.0 * Egamma / M2;
-
-        const double term_soft_tw_3_4 = e_u * m_B * f_B / (4.0 * Egamma * Egamma) * (
-                2.0 * Egamma / (m_rho * m_rho) * std::exp(m_rho * m_rho / M2)
-                    * lapltr_incomplete_dsigma(omega_cut, sigma) / (-sigma)
-                - norm_incomplete(omega_cut)
-            );
+        // For consistency with the current implementation, the twist-3,4 contribution
```
dvandyk, line 244: This could have been achieved by introducing a further ``switch_soft_tw34`` or similar, could it not?

pluegh: This is a good option, switching it off per default. Do you think a comment here in the code, that the contribution is incomplete, is sufficient?

dvandyk (approved):
Thanks @pluegh!

Issue #771, opened 2024-01-18
EOS wrongly throws a warning that some parameters do not appear in the likelihood


smeiser:
I have a problem running an analysis. I defined the likelihood in likelihood.yaml and then I run: eos-analysis sample-nested -f likelihood.yaml no-fsfd -d 1.0 -s 1701 . This then gives me the following warnings:
```
WARNING:EOS:likelihood does not depend on parameter 'B^0->D^-pi^+::BR@Parametric'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'B^0->D^-K^+::BR@Parametric'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'hadronization::fs/fd@LHCb(7TeV)'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'hadronization::fs/fd@LHCb(13TeV)'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'B^0->D^*-pi^+::BR@Parametric'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'B_s^0->D_s^*-pi^+::BR@Parametric'; remove from prior or check options!
WARNING:EOS:likelihood does not depend on parameter 'B^0->D^*-K^+::BR@Parametric'; remove from prior or check options!

```
However, I checked that the parameters that are mentioned in the warning are being used in the constraints. I also checked for typos in the parameter names, but I didn't see any.

PR #773, opened 2024-01-19
Changes for ``master``

dvandyk:
- When plotting a constraint, we currently produce an ungraceful Python exception if the plot contains no data points whatsoever.
This problem, for example, happens if you plot ``process::observable1`` but the constraint does not contain that observable. Practically, this happened in one of my Jupyter notebooks, where I used the same code to create plots for all ``B->D^*`` form factors, but one constraint did not provide information on the tensor form factors.
The user will now be informed of this issue through the logfile and the plotting will continue w/o the use of the constraint at issue.
- When creating an expression observable, it should keep track of the parameters used by the expression and provided through the observable's ``ParameterUser`` interface. This is now done, using a dedicate ``ExpressionVisitor`` called ``ExpressionUsedParameterVisitor`` (#771)

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -989,6 +989,10 @@ class Plotter:
                 yvalues = np.array(yvalues)
                 yerrors = np.array(yerrors)
 
+                if len(xvalues) == 0:
+                    eos.info(f'   skipping plot for constraint {name} since it does not contain the requested observable')
+                    return
+
                 if self.plot_residues:
                     yvalues -= obs_values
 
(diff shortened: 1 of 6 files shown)


python/eos/plot/plotter.py
```diff
@@ -989,6 +989,10 @@ def plot(self):
                 yvalues = np.array(yvalues)
                 yerrors = np.array(yerrors)
 
+                if len(xvalues) == 0:
```
mreboud, line 992: Hmm... Not very clean, but I guess we will improve all of this in the new plotting framework :)

mreboud (approved):
Looks good, thanks a lot! (I though we did that already, but apparently not!)

PR #777, opened 2024-01-28
Restructure the documentation

dvandyk:
Changes include

- [x] correct and amend several entries in ``eos/references.yaml``;
- [x] turn the bibliography into a table and provide hyperlinked badges to the arXiv and to CDS whereever possible;
- [x] restructure the documentation.

The new structure is roughly:
    
```
  doc/
     |
     + user-guide/
     |
     + reference/
```

The existing example notesbooks and the page on common tasks have been
moved to the user guide.

The pages on the Python API, the CLI, the analysis description format,
the domain-specific language for defining new observables, the lists of
parameters, observables, constraints, and signal PDFs, and the
bibliography parameter have been moved to ``reference/``.


MJKirk:
I like the new structure!
One point: in at least two places there's a hyperlink to user-guide/analysis-organisation.html, which currently doesn't work - are you going to add a new guide or is this supposed to be a reference to something already existing?

MJKirk (approved):

Issue #418, opened 2021-11-10
:


dvandyk:

dvandyk:
This should also make use of the solution to #618.

PR #784, opened 2024-01-31
Add sbqq sector of the WET

MJKirk:
Implementing the sbcc sector as part of the wider sb sector, so as to partially implement the idea of issue #418 

Draft for now, as at least two more things need doing:
- [ ] Danny is going to add the RG running for the SM sbcc coefficients
- [ ] Deal with my comments below

Once we are happy, I can also squash commits together


eos/models/wet.cc
```diff
@@ -579,6 +579,74 @@ namespace eos
...
+                quark_flavour_string = "u";
+                break;
+            case QuarkFlavor::down:
+                quark_flavour_string = "d";
+                break;
+            case QuarkFlavor::strange:
+            case QuarkFlavor::bottom:
+            default:
+                throw InternalError("ClassVhadronic::full WCs only implemented for u,d,c");
+        }
```
MJKirk, line 602: I'm not convinced having a `QuarkFlavor` argument and then doing a switch/case is much nicer than just passing in the variable quark prefix as a string, and checking it is either `cc`, `uu`, or `dd`

eos/models/wilson-coefficients.hh
```diff
@@ -37,6 +37,11 @@ namespace eos
     {
         struct ClassII {};  // |Delta B| = 1 semileptonic operators, cf. [AFGV:2017A], eq. (2.5), p. 6.
         struct ClassIII {}; // |Delta B| = 1 = |Delta C| four-quark operators, cf. [AFGB:2017A], eq. (2.6), p. 7.
+        namespace ClassVhadronic
+        {
+            struct full {}; // |Delta B| = 1, |Delta C| = 0 four-quark operators sbqq (q=u,d,c), cf. [AFGB:2017A], eq. (2.10)-(2.11), p. 8.
+            struct restricted {}; // |Delta B| = 1, |Delta C| = 0 four-quark operators, sbqq (q=s,b) cf. [AFGB:2017A], eq. (2.10)-(2.12), p. 8.
```
MJKirk, line 43: Just now Danny and I couldn't come up with anything better than `full` and `restricted` but maybe there is something else

PR #790, opened 2024-02-03
Fix issue787

gubernari:


eos/form-factors/parametric-bgjvd2019-impl.hh
```diff
@@ -185,6 +186,13 @@ namespace eos
         _m_B(p[Process_::name_B], *static_cast<ParameterUser *>(this)),
         _m_P(p[Process_::name_P], *static_cast<ParameterUser *>(this))
     {
+        static const Log::OneTimeMessage message_HQET_FFs_PToP
+        (
+            std::string("HQETFormFactors<") + Process_::name_B + "->" + Process_::name_P + ",PToP>",
```
dvandyk, line 191: I just tested this in a notebook. I find that both the B->P and B->V message are displayed. And they are display only once, too!

Small issue: ``Process_::name_B`` refers to the parameter name for the B-meson mass. My bad. Can you replace
this with
```c++
"HQETFormFactors<" + Process_::label + ",PToP>"
```
instead? Of course, in all four occurrences!

dvandyk (changes requested):
Hi Nico,

my bad, I told something slightly incorrect. See the comment in line 191 please.
Also, could you adjust the commit message for the first commit? We refer to papers using the proper reference name. Here, it should be ``[BGJvD:2019A]`` rather than ``bgjvd2019``. This ensures that grepping for the reference name, we find all commits that use it!

Thanks,
Danny

eos/form-factors/parametric-bgjvd2019-impl.hh
gubernari: ok, I will implement your suggestions
In the command line it displays only PToP warning, but it is not a big issue

dvandyk (approved):

PR #795, opened 2024-02-06
B -> K^(*) nu nu fix

mreboud:
Set the differential branching fraction of B -> M nu nu to zero outside the physical range

### DIFF
--- a/eos/rare-b-decays/b-to-psd-nu-nu.cc
+++ b/eos/rare-b-decays/b-to-psd-nu-nu.cc
@@ -180,7 +180,12 @@ namespace eos
             const double m_P = this->m_P(), m_P2 = m_P * m_P;
             const double m_b = model->m_b_msbar(mu);
             const double m_s = model->m_s_msbar(mu);
-            const double lambda = eos::lambda(m_B2, m_P2, q2), sqrt_lambda = std::sqrt(lambda);
+            const double lambda = eos::lambda(m_B2, m_P2, q2);
+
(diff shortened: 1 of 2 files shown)


eos/rare-b-decays/b-to-psd-nu-nu.cc
```diff
@@ -180,7 +180,12 @@ namespace eos
             const double m_P = this->m_P(), m_P2 = m_P * m_P;
             const double m_b = model->m_b_msbar(mu);
             const double m_s = model->m_s_msbar(mu);
-            const double lambda = eos::lambda(m_B2, m_P2, q2), sqrt_lambda = std::sqrt(lambda);
+            double lambda = eos::lambda(m_B2, m_P2, q2);
```
dvandyk, line 183: I am not sure that I like this. My default approach would be to keep ``lambda`` const and run
```c++
if (lambda < 0)
{
    return 0;
}
```

mreboud: Ok! I was trying to follow what we have elsewhere in EOS, but I am happy to amend that :)

dvandyk: Maybe I am off. My reasoning would be that
1. the computation of the observable might not always be well defined for ``lambda = 0``;
2. the return of 0 immediately avoids unnecessary computations.

mreboud: I will change that

dvandyk (approved):

PR #808, opened 2024-02-27
Improve the install instructions by recommending virtual environments everywhere

MJKirk:


MJKirk:
Some self-review comments
1) This `\cite` 
https://github.com/eos/eos/blob/fdfa086ee0e37f1cb84331c318c4b3e183867882/doc/installation.rst?plain=1#L194
and `\package`
https://github.com/eos/eos/blob/fdfa086ee0e37f1cb84331c318c4b3e183867882/doc/installation.rst?plain=1#L242
are presumably left over from some old version? Should they be removed or altered?
2) I've written to specifically install ``dynesty==2.0.3``  - I lost track, is this version still specifically required or did we remove the code that broke with newer versions?

doc/installation.rst
```diff
@@ -202,18 +205,17 @@ by running the following commands:
...
 
-::
 
-  # for the 'pip3' command
-  apt-get install python3-pip
-  pip3 install --user dynesty pypmc wilson
-  pip3 install --user --upgrade matplotlib
+::
+  # Make sure to activate your virtual environment first
+  pip install dynesty===2.0.3 matplotlib networkx scipy pypmc pyyaml wilson
```
mreboud, line 218: Three equal signs?

doc/installation.rst
```diff
@@ -247,7 +251,7 @@ To install the remaining packages, run the following command in a shell
 
 ::
 
-  pip3 install matplotlib scipy PyYAML
+  pip3 install dynesty===2.0.3 matplotlib networkx scipy pypmc pyyaml wilson
```
mreboud, line 254: ditto

mreboud (changes requested):

doc/installation.rst
MJKirk: Good catch

mreboud: I am not 100% sure if this is correct or not. I think `===` has a specific meaning. To be checked :)

MJKirk: It does have some specific meaning (see https://packaging.python.org/en/latest/specifications/version-specifiers/#version-matching) but it's not what we want

dvandyk:
> Some self-review comments
> 
> 1. This `\cite`
>    https://github.com/eos/eos/blob/fdfa086ee0e37f1cb84331c318c4b3e183867882/doc/installation.rst?plain=1#L194
>    
>    and `\package`
>    https://github.com/eos/eos/blob/fdfa086ee0e37f1cb84331c318c4b3e183867882/doc/installation.rst?plain=1#L242
>    
>    are presumably left over from some old version? Should they be removed or altered?
Yes, please remove those old LaTeX macros.

> 2. I've written to specifically install `dynesty==2.0.3`  - I lost track, is this version still specifically required or did we remove the code that broke with newer versions?

I think we removed that issue, which was only present in a test case.

dvandyk:
> @dvandyk - what about my comments?

I am certain that I had commented on this. Either Github or the Marseille network seem to have eaten my comments.
 1. Please remove the old LaTeX macros. 
 2. Please remove the version restriction for ``dynesty`` altogether.

Issue #807, opened 2024-02-27
Create a Pyhf example 


mreboud:
It would be nice to have a MWE for how to use Pyhf in the `example` repository.

lorenzennio:
@mreboud I will create a simple example. I think the main reason I have not done it yet, is that with nested sampling this is not very useful for realistic models. It only works properly for small models. But I can just add a small not on that in the example.

mreboud:
> @mreboud I will create a simple example. I think the main reason I have not done it yet, is that with nested sampling this is not very useful for realistic models. It only works properly for small models. But I can just add a small not on that in the example.

Yes that would be perfect! thanks a lot!

dvandyk:
I think I can close this now, since PR #811 has been merged some time ago.

PR #811, opened 2024-02-29
[examples] add pyhf example notebook

lorenzennio:
Add a pyhf example (see #807).

TODOs:
- [x] fix renaming through `parameter_map`. This is not actually renaming the parameter, but only the qualified name.
- [ ] fix warning ```RuntimeWarning: invalid value encountered in subtract
  return xlogy(n, lam) - lam - gammaln(n + 1.0)  # type: ignore[no-any-return]```

### DIFF
--- a/python/eos/pyhf_likelihood.py
+++ b/python/eos/pyhf_likelihood.py
@@ -44,31 +44,61 @@ class PyhfLogLikelihood:
         self.model = self.workspace.model()
         self.data = np.array(self.workspace.data(self.model, include_auxdata=False))
 
-        self.varied_parameters = []
-        for pn, v, bounds in zip(self.model.config.par_names,
-                                 self.model.config.suggested_init(),
-                                 self.model.config.suggested_bounds()):
+        self._pyhf_parameters = []
+        for parameter_name, init, bounds in zip(
+            self.model.config.par_names,
+            self.model.config.suggested_init(),
+            self.model.config.suggested_bounds()
+            ):
 
-            if pn in parameter_map:
-                qn = eos.QualifiedName(parameter_map[pn])
+            kinematics = eos.Kinematics()
+            options = eos.Options()
+
+            if parameter_name in parameter_map:
+                if isinstance(parameter_map[parameter_name], str):
+                    qualified_name = eos.QualifiedName(
+                        parameter_map[parameter_name]
+                        )
+                elif isinstance(parameter_map[parameter_name], dict):
+                    qualified_name = eos.QualifiedName(
+                        parameter_map[parameter_name]['name']
+                        )
+                    kinematics = eos.Kinematics(parameter_map[parameter_name].get('kinematics', {}))
+                    options = eos.Options(parameter_map[parameter_name].get('options', {}))
+                else:
+                    raise ValueError('parameter_map values must be either strings or dictionaries.')
             else:
-                qn = eos.QualifiedName('pyhf::' + pn)
+                qualified_name = eos.QualifiedName('pyhf::' + parameter_name)
 
-            if self.parameters.has(qn):
-                p = self.parameters[qn]
+            # check if qualified name corresponds to an existing observable
+            observables = eos.Observables()
+            if qualified_name in observables:
+                pyhf_obs = eos.Observable.make(qualified_name, self.parameters, kinematics, options)
+                self._pyhf_parameters.append(pyhf_obs)
             else:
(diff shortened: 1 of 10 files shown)


lorenzennio:
@dvandyk any updates on how to assign a custom observable as a pyhf parameter here?

python/eos/pyhf_likelihood.py
```diff
@@ -44,31 +44,52 @@ def __init__(self, cache, workspace, parameter_map={}):
...
+                    options = eos.Options(parameter_map[parameter_name].get('options', {}))
+                else:
+                    raise ValueError('parameter_map values must be either strings or dictionaries.')
             else:
-                qn = eos.QualifiedName('pyhf::' + pn)
+                qualified_name = eos.QualifiedName('pyhf::' + parameter_name)
 
-            if self.parameters.has(qn):
-                p = self.parameters[qn]
+            if self.parameters.has(qualified_name):
```
dvandyk, line 73: The logic here is off. You have 3 cases:
1. ``qualified_name`` corresponds to an existing parameter;
2. ``qualified_name`` corresponds to an observable name;
3. ``qualified_name`` is unknown to EOS.

You are currently handling case 1 in line 74, and both cases 2 and 3 in lines 76...83. That is definitely buggy. I am unsure if that causes the warning you see, though.

lorenzennio: Ah I see. This indeed solves the problem. Before it created a new parameter with the same name as the observable, even if it existed. Hence, the warning. Should be fixed now.

python/eos/pyhf_likelihood.py
```diff
@@ -44,31 +44,60 @@ def __init__(self, cache, workspace, parameter_map={}):
...
+                else:
+                    raise ValueError('parameter_map values must be either strings or dictionaries.')
             else:
-                qn = eos.QualifiedName('pyhf::' + pn)
+                qualified_name = eos.QualifiedName('pyhf::' + parameter_name)
 
-            if self.parameters.has(qn):
-                p = self.parameters[qn]
+            # check if qualified name corresponds to an existing observable
+            if qualified_name in [o[0] for o in eos.Observables()]:
```
lorenzennio, line 74: Is there a better way to check if an observable exists?

python/eos/pyhf_likelihood.py
```diff
@@ -79,7 +108,12 @@ def factory(cache, workspace, parameter_map={}):
     def priors(self):
         priors = []
         for name, _ in self.model.config.modifiers:
-            param_name = self.parameter_map[name] if name in self.parameter_map else 'pyhf::' + name
+            param_name = 'pyhf::' + name
+            if name in self.parameter_map:
+                if isinstance(self.parameter_map[name], str):
+                    param_name = self.parameter_map[name]
+                elif isinstance(self.parameter_map[name], dict):
+                    continue
```
lorenzennio, line 116: I assume now that if a dictionary is passed, it is an observable and hence no prior is assigned. It is the desired behaviour, but maybe not the best way to implement it.

python/eos/pyhf_likelihood.py
dvandyk: Brilliant!

python/eos/pyhf_likelihood.py
lorenzennio: Done!

dvandyk:
I still would like to go over this with you at some point prior to merging. For example, I am puzzled by the fact that the distribution in C_VL^ublnu is flat. Shouldn't you be able to extract C_VL in the 2nd fit?

lorenzennio:
> I still would like to go over this with you at some point prior to merging. For example, I am puzzled by the fact that the distribution in C_VL^ublnu is flat. Shouldn't you be able to extract C_VL in the 2nd fit?

Ah, I am still working on the notebook and what to include. So far the pure pyhf posterior looks like this: 

image

lorenzennio:
@dvandyk now the notebook is updated too. Let me know what you think (:

lorenzennio:
@dvandyk  from my side we can mergre this

lorenzennio:
@dvandyk I would be happy to merge this, as I would like to show this in a bayesian pyhf workshop soon. 

There is still one warning, which comes from the fact that the Gaussian and Poisson priors in eos are unbounded, whereas pyhf parameters are bounded. But this is not too much of an issue if your model is sensible. We could create a separate issue for this and fix it later.

dvandyk (approved):
Thank you @lorenzennio, merging this ASAP.

PR #815, opened 2024-03-13
Take absolute values of B lifetimes

smeiser:


eos/b-decays/lifetime.cc
```diff
@@ -293,13 +293,13 @@ namespace eos
     double
     Lifetime::decay_width_dbcu() const
     {
-        return _imp->decay_width_dbcu();
+        return std::fabs(_imp->decay_width_dbcu());
```
dvandyk, line 296: I think this requires a bit of a comment. Something along the lines of:
```
/*
 * The partial decay width computed with ``decay_width_dbcu`` can be negative.
 * This issue arises, since
 *  a) the computation is only a partial one (i.e., further terms can arise);
 *  b) the computation is done to leading-order in alpha_s.
 * For the purpose of determining bounds on the size of the WET Wilson coefficients,
 * we would like to compare this partial decay width with the measured total decay
 * width. Returning the absolute value suffices for this comparison.
 */
 ```

dvandyk (changes requested):

PR #816, opened 2024-03-13
Add three-particle contributions to B -> D^(*) K and Bs -> Ds^(*) pi

smeiser:


dvandyk (approved):
Well done.
 - [x] Clean implementation.
 - [x] Tested for SM and BSM cases.
 - [x] Included in the accuracy selection.
 
 That makes me very happy! :smile:

PR #817, opened 2024-03-18
Rework AnalysisFile 

MJKirk:
Some comments of my own:

I have tested this new code against all the analysis file examples I have access to.

For the two examples using pyhf in #811, my `PyHFConstraintDescription` class is adequate, but it would good to have someone who knows that part of the code check if there are other attributes I have missed.
I left a note in the code about a corner case I found - if the analysis file using a pyhf likelihood and the corresponding workspace file are not in the directory you run EOS from (see https://github.com/MJKirk/eos/blob/3164502236ea72ca2d08053eef5b760208019227/python/eos/analysis_file.py#L119)

I have only one example using the steps part of the `AnalysisFile`, so again this is reconstructed from that one example plus the code. See also my `??????` in the schema I wrote for myself at the bottom of `analysis_file_description.py`. 

For the moment, the arguments when creating an `Analysis` object are converted back from my classes to dicts so the change is compatible with no changes needed there.


python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+                return Deserializable.make(PoissonPriorDescription, **_kwargs)
+
+
+        raise ValueError('Unknown type of prior description')
+
+@dataclass
+class PoissonPriorDescription(Deserializable):
+    parameter:str
+    k:float
+    type: str = field(repr=False, init=False, default="poisson")
```
dvandyk, line 62: Coding styles of lines 60+61 and 62 differ. I think elsewhere (in ``python/eos/tasks.py``) we use something like
```py
type:str=field(repr=False, init=False, default="poisson")
```

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+@dataclass
+class CurtailedGaussianDescription(Deserializable):
+    parameter:str
+    central:float
+    sigma:float
+    min:float
+    max:float
+    type: str = field(repr=False, init=False, default="curtailed gaussian")
```
dvandyk, line 71: Ditto.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+@dataclass
+class GaussianPriorDescription(Deserializable):
+    parameter: str
+    central: float
+    sigma: float
+    type: str = field(repr=False, init=False, default="gaussian")
```
dvandyk, line 78: Ditto.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+@dataclass
+class ScalePriorDescription(Deserializable):
+    parameter: str
+    min: float
+    max: float
+    mu_0: float
+    lambda_scale: float
+    type: str = field(repr=False, init=False, default="scale")
```
dvandyk, line 88: Ditto.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+@dataclass
+class UniformPriorDescription(Deserializable):
+    parameter: str
+    min: float
+    max: float
+    type: str = field(repr=False, init=False, default="uniform")
```
dvandyk, line 95: Ditto.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+
+@dataclass
+class EOSConstraintDescription(Deserializable):
+    constraint: str
+
+@dataclass
+class ManualConstraintDescription(Deserializable):
+    name: str
```
dvandyk, line 127: Does using ``eos.QualifiedName`` in lieu of ``str`` work here? Ditto for line 123.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+        if 'manual_constraints' in kwargs:
+            _kwargs['manual_constraints'] = [ManualConstraintDescription.from_dict(name=n, info=d) for n, d in kwargs['manual_constraints'].items()]
+        if 'pyhf' in kwargs:
+            _kwargs['pyhf'] = PyHFConstraintDescription.from_dict(**kwargs['pyhf'])
+        return Deserializable.make(cls, **_kwargs)
+
+
+
+@dataclass
+class PosteriorComponent(Deserializable):
```
dvandyk, line 158: I don't like the name. It suggests that this does not describe a complete posterior, although it does.

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+        if "parameters" in kwargs:
+            eos.warn(f'\'parameters\' is in the description of prior component \'{kwargs["name"]}\', use \'descriptions\' instead')
+            _kwargs.pop("parameters")
+            _kwargs['descriptions'] = [PriorDescription.from_dict(**d) for d in kwargs['parameters']]
+        return Deserializable.make(cls, **_kwargs)
+
+
+
+@dataclass
+class EOSConstraintDescription(Deserializable):
```
dvandyk, line 122: Hm. Do not like the ``EOS`` prefix here. Why not ``ConstraintLikelihoodDescription``?

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+
+@dataclass
+class PredictionObservableComponent(Deserializable):
+    name: str
+    kinematics: dict = field(default_factory=dict)
+
+@dataclass
+class PredictionComponent(Deserializable):
```
dvandyk, line 183: Hm, again, not a component. A full prediction set, similar to what happens for a posterior.

python/eos/analysis_file.py
```diff
@@ -85,9 +85,16 @@ def __init__(self, analysis_file):
         if 'parameters' in input_data:
             _params = [ParameterComponent.from_dict(name=n, **d) for n, d in input_data["parameters"].items()]
             for p in _params:
-                qn = eos.QualifiedName(p.name)
-                eos.Parameters.declare(qn, p.latex, eos.Unit(p.unit), p.central, p.min, p.max)
-                eos.info(f'Successfully declared parameter: {qn}')
+                try:
```
dvandyk, line 88: I think this commit should be squashed with the previous one.

python/eos/analysis_file_description.py
```diff
@@ -202,6 +202,7 @@ class ParameterComponent(Deserializable):
     central: float
     min: float
     max: float
+    alias_of: list = field(default_factory=list)
```
dvandyk, line 205: This should be squashed into the commit introducing ``analysis_file_description.py``.

dvandyk (changes requested):
Mostly nit picky comments and requests. Centred around naming of classes and splitting the last commit + squashing the results into commits no.2 and no.3.

python/eos/analysis_file_description.py
MJKirk: I chose this as these are pre-existing EOS constraints, so it is literally a description of an EOS constraint. But I didn't do that for the EOS priors so :shrug:. It would probably make sense for `EOSConstraintDescription` and `ConstraintPriorDescription` to be named consistently though. I would go for `EOSConstraintDescription` and `EOSConstraintPriorDescription`, but `ConstraintLikelihoodDescription` and `ConstraintPriorDescription` work fine as well.

python/eos/analysis_file_description.py
MJKirk: Seems to, so will change

python/eos/analysis_file_description.py
MJKirk: Fair - `PosteriorDescription` better?

python/eos/analysis_file_description.py
MJKirk: Again -  `PredictionDescription` better?

python/eos/__init__.py
MJKirk, line 1: Do we want to `import *` here? Or even import anything?

python/eos/analysis_file_description.py
dvandyk: I would remove ``EOS`` from the name altogether and prefix with ``Manual`` in cases where we add information not contained in the constraints database / parameter database.

python/eos/analysis_file_description.py
dvandyk: Yes.

python/eos/analysis_file_description.py
dvandyk: Yes!

python/eos/__init__.py
dvandyk: I would not import anything at the level of the full EOS module. I would only import within ``analysis_file.py``.

MJKirk: Removed

python/eos/analysis_file_description.py
MJKirk: Done

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,281 @@
...
+    fixed_parameters:dict=field(default_factory=dict)
+
+
+
+@dataclass
+class ObservableComponent(Deserializable):
+    name:str
+    latex:str
+    unit:str
+    options:dict
```
dvandyk, line 142: ``options`` should be optional

python/eos/analysis_file_description.py
```diff
@@ -0,0 +1,281 @@
...
+    expression:str
+
+
+
+@dataclass
+class PredictionObservableComponent(Deserializable):
+    name:str
+    kinematics:dict=field(default_factory=dict)
```
dvandyk, line 150: Add comment:
```
TODO (Python 3.10): this should be "dict | list[dict]" once we bump the minimal python version to 3.10.
```

dvandyk (approved):
Nitpicky stuff, see review comments.

python/eos/analysis_file_description.py
MJKirk: Just FYI, in the current code, `options` is mandatory (see below) 
https://github.com/eos/eos/blob/e128cf93434ddb0378f98b3fee615eceb4c2f8dd/python/eos/analysis_file.py#L68-L77
But it does seem like it could be optional, since observables don't need to take options

PR #819, opened 2024-03-21
Bug fixes in the running of fperp of vector LCDAs and in the KStarLCDAs Gegenbauer moments

smeiser:


mreboud:
That's ok on EOS' side. What about the physics? Were Beneke et al. incorrect or is this just a typo?

dvandyk:
> That's ok on EOS' side. What about the physics? Were Beneke et al. incorrect or is this just a typo?

We just clarified. The PR is OK as it is. So is the expression in Beneke et al, which only holds for nf=5. However, here we need to work agnostically, i.e., without any knowledge of the number of active quark flavours.

dvandyk (approved):

PR #826, opened 2024-03-27
Changes to the release procedure

dvandyk:
As discussed at a recent developer meeting, we will change our release procedure to decrease the number of development release to PyPI.

1. We will continue to push changes to ``master`` via pull requests. Pull request will need passing checks and approval. We will no longer produce (pre)release PyPI wheel or packagecloud .deb files from ``master``.

2. We will periodically push from ``master`` to ``release``. The new branch already exists and a branch protection has been setup. Every push to ``release`` will produce either a prerelease PyPI wheel + documentation; or produce a full release PyPI wheel + documentation + packagecloud .deb file.

- [x] ``pre-commit`` workflow only applied to ``master`` PRs. No need to clutter ``release`` with it.
- [x] ``PyPI`` workflow only deploys wheels when triggers for the ``release`` branch.
- [x] ``Ubuntu`` workflow only deploys documentation and .deb file for the ``release`` branch.

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -1,5 +1,7 @@
 on:
-    pull_request:
+  pull_request:
+    branches:
+      - master
 
 name: Check if pre-commit checks pass
 
--- a/.github/workflows/pypi-build+check+deploy.yaml
+++ b/.github/workflows/pypi-build+check+deploy.yaml
@@ -1,6 +1,6 @@
 on:
     push:
-        branches: [ master, testing ]
+        branches: [ release, master, testing ]
 
     pull_request:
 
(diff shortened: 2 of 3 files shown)


.github/workflows/pre-commit.yaml
```diff
@@ -1,5 +1,7 @@
 on:
-    pull_request:
+  pull_request:
+    branches:
+      - master
```
mreboud, line 4: Will the other branches like issueYYY etc... also have the pre-commit tests?

mreboud (commented):
I just have one question, see comment

.github/workflows/pre-commit.yaml
dvandyk: If I read [the documentatio](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches) correctly, then yes.

The list of ``branches`` are those **targeted** by a PR, not those that a PR originates from.

mreboud: Indeed!

MJKirk: From the documentation, it seems to me they won't. But there's a `branches-ignore` [option](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore), so you can do 
```
on:
  pull_request:
    branches-ignore:
      - release
```
instead.

mreboud: Oh, I merged too fast in this case :(
Let's see with the next PR and if it doesn't work I will implement your fix @MJKirk

PR #845, opened 2024-04-30
Improvements in the handling of eos DataSets

mreboud:


dvandyk:
Hmm. Why does the import fail? Do the ``scipy`` imports in the last commit depend on a newer ``scipy`` version than available in the ``manylinux_2_28`` images?

mreboud:
> > Hmm. Why does the import fail? Do the `scipy` imports in the last commit depend on a newer `scipy` version than available in the `manylinux_2_28` images?
> 
> Aha. It only fails for Python version 3.8!

I think it is the `list[str]` specification that you used for the authors of a a dataset. This kind of specifications seem to be accepted only for python versions > 3.8...

dvandyk:
> > > Hmm. Why does the import fail? Do the `scipy` imports in the last commit depend on a newer `scipy` version than available in the `manylinux_2_28` images?
> > 
> > 
> > Aha. It only fails for Python version 3.8!
> 
> I think it is the `list[str]` specification that you used for the authors of a a dataset. This kind of specifications seem to be accepted only for python versions > 3.8...

Thanks for noticing that. I rebased on eos/eos@b46e54b (handled in #847), which bumps the minimal Python version to 3.9.

python/eos/data/native.py
```diff
@@ -195,13 +202,22 @@ def density(self):
 
 
     @staticmethod
-    def create(path, density):
+    def create(path, density, varied_parameters=None, sigma_test_stat=None, samples=None, weights=None):
```
dvandyk, line 205: I created a new issue (#848) to remind us to make ``varied_parameters`` mandatory for creating any datafile.
We should merge this as is though.

python/eos/data/native.py
```diff
@@ -217,6 +233,25 @@ def create(path, density):
...
                 raise RuntimeError(f'Unsupported type of MixtureDensity component: {type(c)}')
         description['weights'] = density.weights.tolist()
+        if varied_parameters is not None:
+            description['varied_parameters'] = list(varied_parameters)
+
+        # The test statistics defaults to two empty lists
+        description['test statistics'] = { "sigma": [], "densities": [] }
+        if sigma_test_stat is not None and samples is not None and weights is not None:
+            sigma_test_stat = _np.array(sigma_test_stat)
+            samplesPDF = list(map(lambda x: -density.evaluate(x), samples))
```
dvandyk, line 243: Coding style: ``samplesPDF`` -> ``samples_pdf``. I will fix this is in a bit.

dvandyk (approved):

PR #849, opened 2024-05-10
Allow specifying options to a prediction observable in an AnalysisFile

### LINKED ISSUE #834: Enable specification of ``options`` in a ``PredictionObservableComponent``

MJKirk:
Resolves #834

### DIFF
--- a/python/eos/analysis_file.py
+++ b/python/eos/analysis_file.py
@@ -176,7 +176,6 @@ class AnalysisFile:
         global_options.update(prediction.global_options)
         fixed_parameters.update(prediction.fixed_parameters)
 
-        options = eos.Options(**global_options)
         for (p, v) in fixed_parameters.items():
             parameters.set(p, v)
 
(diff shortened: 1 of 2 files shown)


python/eos/analysis_file.py
```diff
@@ -185,9 +184,18 @@ def observables(self, _posterior, _prediction, parameters):
             for key, value in options_part:
                 if key in global_options and global_options[key] != value:
                     eos.error(f'Global option {key}={global_options[key]} overrides option part specification {key}={value} for observable {o.name} in prediction {_prediction}.')
+                if key in o.options and o.options[key] != value:
+                    eos.error(f'Local option {key}={o.options[key]} overrides option part specification {key}={value} for observable {o.name} in prediction {_prediction}.')
```
mreboud, line 188: I forgot to ask during the group meeting. Shouldn't one of these errors be a warning instead? @dvandyk?

MJKirk: I used `eos.error` as that was what was already being raised for the existing case of options overriding. Although they all would seem to make more sense as warnings

python/eos/analysis_file.py
```diff
@@ -185,9 +184,18 @@ def observables(self, _posterior, _prediction, parameters):
...
+            for key, global_value in global_options.items():
+                if key in o.options and o.options[key] != global_value:
+                    eos.error(f'Local option {key}={o.options[key]} overrides global option {key}={global_value} for observable {o.name} in prediction {_prediction}.')
```
dvandyk, line 191: Make this a warning.

python/eos/analysis_file.py
dvandyk: This one should probably stay an error, since it might be a bit surprising that an option part specification is overwritten by the local options.

dvandyk (changes requested):

dvandyk (approved):

PR #856, opened 2024-05-24
Fix inconsistent definition of kappa_4 parameter appearing in twist 4 Kaon LCDAs

smeiser:
Currently we have eq. 4.19 of https://arxiv.org/pdf/hep-ph/0603063 implemented, with kappa4(mu0) as a parameter. However, kappa_4(mu0) is fixed by eq. 4.18. Thus having kappa4(mu0) as a parameter can lead to inconsistent input, that would violate eq. 4.18. Therefore, I propose to remove kappa4_0 as a parameter and rather just plug in the expression given in 4.18 into 4.19. This ensures that there are no inconsistencies. However, this now breaks a number of test cases that have to be adjusted.

- [x] remove kappa_4(mu0) and update the expression for the running of kappa4.
- [x] fix k-lcda test cases
- [x] look for all other tests that might break

### DIFF
--- a/eos/b-decays/bq-to-dq-psd_TEST.cc
+++ b/eos/b-decays/bq-to-dq-psd_TEST.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2023 Stefan Meiser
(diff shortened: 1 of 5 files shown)


mreboud (approved):
Thank you @smeiser!

dvandyk (approved):

Issue #844, opened 2024-04-29
Create a new task ``sample-prior``


dvandyk:
Presently, pure prior predictions cannot be achieved using the tasks framework. Let's add a new tasks that produce prior samples and stores them in ``{POSTERIOR}/samples``. Since we have already provided a prior sampling interface to ``dynesty``, this should be rather simple. Will try to do this together with Ery as a learning exercise.

Issue #712, opened 2023-11-07
New example notebook showing calculation of simpler theory uncertainty 


MJKirk:
Suggested to Danny to add a new example notebook showing the calculation of the theory uncertainty for an observable without a kinematic argument, e.g. something like DeltaMs or Bs->mu mu.

For DeltaMs, this would need two parts:
- [ ] Add the appropriate constraints on the meson mixing bag parameters
- [ ] Write an example notebook to add to the other EOS examples

dvandyk:
I would highly suggest to use $\bar{B}_s\to \mu^+\mu^-$ instead of $\Delta m_s$, since it is easier from the point of view of the computing the hadronic uncertainties.

dvandyk:
@MJKirk I have just pushed a branch ``issue712``, adding
 - [x] a commit to export ``B_c->lnu::BR``, which has only a single hadronic matrix element, no kinematics, and is thematically connected to the b->clnu decays that populate all our examples;
 - [ ] a WIP commit that restructures the predictions notebook and add an analysis (predictions.yaml)

MJKirk:
Cheers Danny, the example outline looks good. The $B_s \to \mu \mu$ example I was looking at had an issue where I seemed to be getting much different uncertainties to `flavio`, but I never found time to follow up (one for the dev retreat perhaps?).

dvandyk:
Closing since changes to the notebook have been merged.

PR #858, opened 2024-05-28
Rewrite example notebook on theory predictions

dvandyk:
- [x] rewrite the notebook to use additionally a simpler example of an observable without kinematic variables (#712)
- [x] add such a simple observable within the ``cblnu`` sector, which we use as a red thread for all example notebooks
- [x] add a new task ``sample-prior`` to illustrate simple prior sampling using an analysis file (#844, thanks @ery-mcpartland !)
- [x] add a section to the notebook illustrating the use of an analysis file.

### DIFF
--- a/examples/predictions.ipynb
+++ b/examples/predictions.ipynb
@@ -11,7 +11,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "EOS can produce theory predictions for any of its built-in observables. The examples following in this section illustrate how to find a specific observable from the list of all built-in observables, construct an [eos.Observable](../reference/python.rst#eos.Observable) object and evaluate it, and estimate the theoretical uncertainties associated with it."
+    "EOS can produce theory predictions for any of its built-in observables. The examples following in this section illustrate\n",
+    "1. how to find a specific observable from the list of all built-in observables;\n",
+    "2. manually construct an [eos.Observable](../reference/python.rst#eos.Observable) object and evaluate it to predict the observables (basic);\n",
+    "3. use an [analysis file](../reference/analysis-file-format.rst) to estimate the theoretical uncertainties associated with one or more observables (advanced)."
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


examples/predictions.ipynb
```diff
@@ -106,10 +144,9 @@
    "metadata": {},
    "source": [
     "The default option ``l=mu`` select $\\ell=\\mu$ as the lepton flavour. The value of the observable is shown\n",
-    "to be about $2.4\\%$,\n",
-    "which is compatible with the current world average for the $\\bar{B}^-\\to D^0\\mu^-\\bar\\nu$ branching ratio.\n",
+    "to be about $9.8\\cdot 10^{-5}$; unfortunately, no experiment has been able to see this decay to datsa.\n",
```
MJKirk, line 147: "datsa" -> "data"

examples/predictions.ipynb
```diff
@@ -187,257 +269,339 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To estimate theoretical uncertainties of the observables, EOS uses Bayesian statistics.\n",
-    "The latter interprets the theory parameters as random variables and assigns *a priori*\n",
-    "probability density functions (prior PDFs) for each parameter.\n",
+    "To estimate theoretical uncertainties of an observable $O$, EOS uses Bayesian statistics.\n",
+    "The latter interprets the theory parameters as random variables $\\vec\\vartheta$ and assigns *a priori*\n",
```
MJKirk, line 273: What do you mean by "the latter" here? Did this previous sentence originally contain two things (like Bayesian not frequentist, or something) at some point?

MJKirk (changes requested):
Reads nicely Danny. Just two minor comments

examples/predictions.ipynb
dvandyk: No, this is a screw up by a non-native speaker. Will fix it!

examples/predictions.ipynb
dvandyk: Thanks!

MJKirk (approved):

PR #864, opened 2024-06-10
Include the most recent changes from the ``scnul`` branch

dvandyk:
- [x] add missing D->K form factors by FNAL/MILC (@cbolognani)
- [x] add a joint constraint obtained from a fit of D->K form factors to HPQCD + FNAL/MILC results (@cbolognani)


python/eos/analysis_file_description.py
```diff
@@ -38,7 +38,7 @@ class CurtailedGaussianDescription(Deserializable):
     sigma:float
     min:float
     max:float
-    type:str=field(repr=False, init=False, default="curtailed gaussian")
+    type:str=field(repr=False, init=False, default="gaussian")
```
dvandyk, line 41: This is weird. Why isn't this fix in ``master`` yet?

MJKirk: It is in master, your branch isn't properly up to date

MJKirk (approved):

PR #865, opened 2024-06-10
Miscellaneous changes

dvandyk:
Nothing to serious
- [x] Fix the docstring for ``eos.tasks.find_clusters``
- [x] Clarify who throws ``UnknownOptionError`` and ``UnspecifiedOptionError`` (all uses are correct as far as I can see; triggered by a recent PR review that was about to introduce an incorrect use)
- [x] Add a basic ``CODEOWNERS`` file: if we have Github Team support, we should leverage it!

### DIFF
--- a/eos/utils/options.hh
+++ b/eos/utils/options.hh
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2010, 2011, 2015, 2018 Danny van Dyk
+ * Copyright (c) 2010-2024 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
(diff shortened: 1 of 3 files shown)


eos/utils/options.hh
```diff
@@ -31,8 +31,7 @@
 namespace eos
 {
     /*!
-     * UnknownOptionError is thrown when no option of a given name could be
-     * found.
+     * UnknownOptionError is thrown when an Options object does contain a value for a given option key.
```
MJKirk, line 34: Pretty sure this should be "does not contain a value"

dvandyk: Silly me. Silly, silly me.

MJKirk (approved):

PR #869, opened 2024-06-17
Add LO lifetime contributions from sbcu and dbcu operators

smeiser:


eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+                {
+                    336.0 * z * (-1.0 - 9.0 * std::pow(z, 2) + 9.0 * std::pow(z, 4) + std::pow(z, 6) - 12.0 * (std::pow(z, 2) + std::pow(z, 4)) * log(z)),
+                    768.0 * z * (-1.0 - 9.0 * std::pow(z, 2) + 9.0 * std::pow(z, 4) + std::pow(z, 6) - 12.0 * (std::pow(z, 2) + std::pow(z, 4)) * log(z)),
+                    -60.0 * (-1.0 + 8.0 * std::pow(z, 2) - 8.0 * std::pow(z, 6) + std::pow(z, 8) + 24.0 * std::pow(z, 4) * log(z)),
+                    576.0 * (-1.0 + 8.0 * std::pow(z, 2) - 8.0 * std::pow(z, 6) + std::pow(z, 8) + 24.0 * std::pow(z, 4) * log(z)),
+                    -12480.0 * (-1.0 + 8.0 * std::pow(z, 2) - 8.0 * std::pow(z, 6) + std::pow(z, 8) + 24.0 * std::pow(z, 4) * log(z))
+                }
+            }};
+
+            const std::array<std::complex<double>, 5u> C_sing
```
dvandyk, line 324: I would order this as ``C_sing`` before ``G``. A few brief comment on what the various quantities are would be appreciated.

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+            {
+                wc.c1(),  wc.c3(),  wc.c5(),  wc.c7(),  wc.c9()
+            };
+
+            std::array<std::complex<double>, 5u> Cconj_sing;
+            complex<double> (*conj_sing)(const std::complex<double> &) = &std::conj<double>;
+            std::transform(C_sing.cbegin(), C_sing.cend(), Cconj_sing.begin(), conj_sing);
```
dvandyk, line 331: could you not replace ``conj_sing`` with ``&std::conj<double>`` in line 331 and remove line 330 entirely?

eos/b-decays/lifetime.cc
dvandyk: Also, can you elaborate why you append the suffix ``_sing``?

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+
+            const std::array<std::complex<double>, 5u> C_sing_p
+            {
+                wc.c1p(), wc.c3p(), wc.c5p(), wc.c7p(), wc.c9p()
+            };
+
+            std::array<std::complex<double>, 5u> Cconj_sing_p;
+            complex<double> (*conj_sing_p)(const std::complex<double> &) = &std::conj<double>;
+            std::transform(C_sing_p.cbegin(), C_sing_p.cend(), Cconj_sing_p.begin(), conj_sing_p);
```
dvandyk, line 340: See above. You could use the same function pointer!

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+
+            const std::array<std::complex<double>, 5u> C_oct
```
dvandyk, line 342: Ah, singlet vs octett!

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+            {
+                wc.c2(),  wc.c4(),  wc.c6(),  wc.c8(),  wc.c10()
+            };
+
+            std::array<std::complex<double>, 5u> Cconj_oct;
+            complex<double> (*conj_oct)(const std::complex<double> &) = &std::conj<double>;
+            std::transform(C_oct.cbegin(), C_oct.cend(), Cconj_oct.begin(), conj_oct);
```
dvandyk, line 349: Ditto.

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+            complex<double> (*conj_oct_p)(const std::complex<double> &) = &std::conj<double>;
+            std::transform(C_oct_p.cbegin(), C_oct_p.cend(), Cconj_oct_p.begin(), conj_oct_p);
+
+            const double Gamma_0      = g_fermi * g_fermi * ckm * ckm * power_of<5>(m_b) / (192.0 * power_of<3>(M_PI));
+            const double Gamma_sing   = real(dot(Cconj_sing, (G * C_sing)));
+            const double Gamma_sing_p = real(dot(Cconj_sing_p, (G * C_sing_p)));
+            const double Gamma_oct    = 2.0 / 9.0 * real(dot(Cconj_oct, (G * C_oct)));
+            const double Gamma_oct_p  = 2.0 / 9.0 * real(dot(Cconj_oct_p, (G * C_oct_p)));
+
+            return Gamma_0 * (Gamma_sing + Gamma_sing_p + Gamma_oct + Gamma_oct_p) / hbar * 1.0e-12;
```
dvandyk, line 366: Magic numbers are bad. What is ``1.0e-12`` here? Conversion from GeV to inverse ps? I see we used that also in the old functions. Declaring it as e.g.
```
static const double convert_GeV_to_inv_ps = 1.0e-12;
```
would help (if my hunch is correct).

eos/b-decays/lifetime.cc
```diff
@@ -272,6 +273,190 @@ namespace eos
...
+        }
+
+        double decay_width_lo_sbcu() const
```
dvandyk, line 369: Can you re-order and re-name the individual contributions? I would suggest:

```
decay_width_dbcu_dim3_lo
decay_width_dbcu_dim6_lo
decay_width_dbcu_dim3_lo
decay_width_dbcu_dim6_lo
```

dvandyk (changes requested):
Nice work @smeiser. Please see the detailed inline comments.

eos/b-decays/lifetime.cc
dvandyk: Conversion involves ``hbar``, which is a parameter. Changed this instead to an inline function:
```c++
inline double convert_GeV_to_inv_ps() const { return 1.0e-12 / hbar(); }
```

dvandyk (approved):

PR #870, opened 2024-06-18
Fix constraints related to class-I non-leptonic B-decays

smeiser:
- fixed the correlation matrix in the constraints related to BRs and FFs.


dvandyk (approved):
A bit embarrassed that I had not noticed this in the first place!

Good to go!

PR #871, opened 2024-06-21
Clarify LHCb's convention for B->K*ll angular observables

mreboud:


eos/rare-b-decays/observables.cc
```diff
@@ -1287,6 +1306,34 @@ namespace eos
                         Unit::None(),
                         R"( -1.0 * <<B->K^*ll::A_FB>> )"),
 
+                make_expression_observable("B->K^*ll::P_1_LHCb", R"(P_1^\mathrm{LHCb}(\bar{B}\to \bar{K}^*\ell^+\ell^-))",
```
dvandyk, line 1309: Should we use``B->K^*ll::P_1@LHCb``? That is kind of what the suffix part is made for :smile:

mreboud: Yes definitely, I will change that

dvandyk (approved):

PR #872, opened 2024-06-24
Merge remaining ``scnul`` changes

dvandyk:
- [x] Remove instability in Lc->Llnu decays if q2 is small
- [x] Add means to rescale a covariance matrix for an existing constraint
- [x] Permit floating point numbers a option values
- [x] Ensure conversion to ``float`` when reading kinematic variables from YAML.


eos/constraint.cc
```diff
@@ -1261,6 +1264,7 @@ namespace eos
             // If specified, these options allow to specify a subset of the measurements
             unsigned begin = destringify<unsigned>(options.get("begin", "0"));
             unsigned end = destringify<unsigned>(options.get("end", stringify(dim_meas)));
+            double rescale_factor = destringify<double>(options.get("rescale-factor", "1.0"));
```
mreboud, line 1267: This line might benefit from a comment :)

mreboud (approved):
I think it is all good!

eos/constraint.cc
dvandyk: Oh, yes. Fixed.

dvandyk:
> I think it is all good!

Thanks. Will merge as soon as the tests pass for the revised PR.

PR #877, opened 2024-07-02
Next attempt at fixing the ``weekly-builds`` workflow

dvandyk:
- [x] Use the ``checkout`` action to download the full history across all branches, rather than ``git remote update``.
- [x] Specify the use of ``bash`` as the shell.
- [x] Output the environment variables RELEASE, MASTER, and BASE as soon as they are set.
- [x] Use the bash built-in test utility over the POSIX one (``[[...]]`` rather than ``[...]``).
- [x] Fix potential bug in using ``=`` vs ``==`` operators.


MJKirk (approved):
Looks perfectly good (although admittedly this is not a useful statement given it was also true beforehand :sweat_smile: )

PR #880, opened 2024-07-15
Start using fine-grained log levels and pretty up ``eos-analysis`` output

dvandyk:
The output can now look as follows at verbosity level 4

image

while at verbosity level 3 we see

image

### DIFF
--- a/src/scripts/eos-analysis
+++ b/src/scripts/eos-analysis
@@ -26,6 +26,11 @@ import scipy
 import sys
 import yaml
 
+try:
+    from termcolor import colored
+except ImportError:
+    colored = lambda s, *args, **kwargs: s
+
 # return the value of the environment variable, or a default value if the variable is unset.
 def get_from_env(envvar, default):
     if not envvar in os.environ:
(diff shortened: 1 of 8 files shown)


src/scripts/eos-analysis
```diff
@@ -496,6 +501,30 @@ Initializes a new analysis directory based on the EOS analysis template.
...
 
+class CustomLogFormatter(logging.Formatter):
+    _MAP_LEVEL_TO_COLOR = {
+        logging.ERROR:      ('✖', 'red'),
+        logging.WARNING:    ('⚠', 'orange'),
+        logging.SUCCESS:    ('🗸', 'green'),
+        logging.COMPLETED:  ('🗸', 'green'),
+        logging.INPROGRESS: ('…', 'green'),
+        logging.INFO:       ('ℹ', 'blue'),
+        logging.DEBUG:      ('🤖', 'black'),
```
mreboud, line 512: Oh boy --"

mreboud (approved):

src/scripts/eos-analysis
dvandyk: @mreboud do you prefer "🐞"?

PR #882, opened 2024-07-20
Use ``argcomplete`` to provide bash completion for ``eos-analysis``

dvandyk:
Steps to enable bash completion:

```
pip install argcomplete # outside all virtual environments!
activate-global-python-argcomplete --user
```


mreboud (approved):
Thank you @dvandyk!

PR #886, opened 2024-07-22
Include <algorithm> where necessary

Herren:
Build from source fails on a Linux system with Python 3.12, Boost 1.83, g++ 14.1.1.
Reason was the use of algorithms from <algorithm> without explicitly including it,

### DIFF
--- a/eos/b-decays/lifetime.cc
+++ b/eos/b-decays/lifetime.cc
@@ -25,6 +25,7 @@
 #include <eos/utils/private_implementation_pattern-impl.hh>
 
 #include <complex>
+#include <algorithm>
 
 namespace eos
 {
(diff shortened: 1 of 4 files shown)


dvandyk (changes requested):
Thank you very much Florian @Herren, the contents look good! Could you please adjust the commit message to comply with the EOS guidelines? Please have a look [at the wiki, point 6](https://github.com/eos/eos/wiki/How-to-Contribute) for how we structure commit messages.

mreboud:
> Thank you very much Florian @Herren, the contents look good! Could you please adjust the commit message to comply with the EOS guidelines? Please have a look [at the wiki, point 6](https://github.com/eos/eos/wiki/How-to-Contribute) for how we structure commit messages.

There will also be a test failure in principle. I will fix it.

PR #888, opened 2024-07-25
Single-Channel Omnes factor

Herren:
This is an implementation of B. Moussallams algorithm (https://arxiv.org/abs/hep-ph/9909292) to solve the integral
equation for single-channel Omnes factors.
The idea is to use it in future to implement, e.g., the pi-pi p-Wave Omnes factor etc.

### DIFF
--- a/eos/maths/legendre-polynomial-vector.hh
+++ b/eos/maths/legendre-polynomial-vector.hh
@@ -1,0 +1,271 @@
+/*
+ * Copyright (c) 2024 Florian Herren
+ *
+ * This file is part of the EOS project. EOS is free software;
+ * you can redistribute it and/or modify it under the terms of the GNU General
+ * Public License version 2, as published by the Free Software Foundation.
+ *
+ * EOS is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 59 Temple
+ * Place, Suite 330, Boston, MA  02111-1307  USA
+ */
+
+#ifndef EOS_GUARD_EOS_MATHS_LEGENDRE_POLYNOMIAL_VECTOR_HH
+#define EOS_GUARD_EOS_MATHS_LEGENDRE_POLYNOMIAL_VECTOR_HH 1
+
+#include <array>
+#include <vector>
+#include <complex>
+
+#include <boost/math/special_functions/legendre.hpp>
+
+namespace eos
+{
+
+    /*
+     * Representation of a vector of Legendre polynomials for m = 0.
+     *
+     */
+    template <unsigned order_>
+    class LegendrePVector
+    {
+        public:
+            LegendrePVector()
+            {
+            }
+
+            LegendrePVector(const LegendrePVector &) = default;
+            LegendrePVector(LegendrePVector &&) = default;
+            ~LegendrePVector() = default;
+            LegendrePVector & operator= (LegendrePVector &&) = default;
+            LegendrePVector & operator= (const LegendrePVector &) = default;
+
+            // Evaluate the vector of Legendre polynomials
+            std::array<double, order_ + 1> constexpr operator() (const double & z) const
+            {
+                std::array<double, order_ + 1> ret_vec;
+
+                if (order_ == 0)
+                {
+                    ret_vec[0] = 1;
+                }
+                else if (order_ == 1)
+                {
+                    ret_vec[0] = 1;
+                    ret_vec[1] = z;
+                }
+                else
+                {
+                    ret_vec[0] = 1;
+                    ret_vec[1] = z;
+                    for (unsigned i = 2 ; i <= order_ ; i++)
+                    {
+                        ret_vec[i] = (z * ret_vec[i - 1] * (2 * i - 1) - ret_vec[i - 2] * (i - 1)) / i;
+                    }
+                }
+                return ret_vec;
+            }
+
+            // Return zeros and compute Gauss-Legendre weights
+            void gauss_legendre(std::array<double, order_> & zeros, std::array<double, order_> & weights)
+            {
+                std::vector<double> zerosp = boost::math::legendre_p_zeros<double>(order_);
+                std::array<double, order_ / 2 + order_ % 2> weightsp;
+                const unsigned len = order_ / 2 + order_ % 2;
+
+                for (unsigned i = 0 ; i < len ; i++)
+                {
+                    double lp = boost::math::legendre_p<double>(order_ + 1, zerosp[i]);
+                    weightsp[i] = 2 * (1 - zerosp[i] * zerosp[i]) / ((order_ + 1) * (order_ + 1) * lp * lp);
+                }
+
+                // Flip order for first half
+                for (unsigned i = 0 ; i < len ; i++)
+                {
+                    zeros[i] = -1 * zerosp[len - 1 - i];
+                    weights[i] = weightsp[len - 1 - i];
+                }
+
+                // Double them, take care in case of odd order_
+                if (order_ % 2 == 1)
+                {
+                    for (unsigned i = 1 ; i < len ; i++)
+                    {
+                        zeros[len + i - 1] = zerosp[i];
+                        weights[len + i - 1] = weightsp[i];
+                    }
+                }
+                else
+                {
+                    for (unsigned i = 0 ; i < len ; i++)
+                    {
+                        zeros[len + i] = zerosp[i];
+                        weights[len + i] = weightsp[i];
+                    }
+                }
+            }
+    };
+
+    /*
+     * Representation of a vector of the real part of the associated Legendre functions for m = 0.
+     * Implementation is based on Zhang, Shanjie and Jin, Jianming. “Computation of Special Functions”, John Wiley and Sons, 1996.
+     *
+     */
+    template <unsigned order_>
+    class LegendreReQVector
+    {
+        private:
+            const double _eps;
+            const unsigned _cut;
+
+        public:
+            // 10^-14 chosen to stay away from double precision noise
+            // 1000 iterations are enough to cover all practical cases,
+            // but keep us from going into an infinite loop if we really
+            // hit a point where things do not converge.
+            LegendreReQVector() :
+                _eps(1e-14),
+                _cut(1000)
+            {
+            }
+
+            LegendreReQVector(const double & eps, const unsigned & cut) :
+                _eps(eps),
+                _cut(cut)
+            {
+            }
+
+            LegendreReQVector(const LegendreReQVector &) = default;
+            LegendreReQVector(LegendreReQVector &&) = default;
+            ~LegendreReQVector() = default;
+            LegendreReQVector & operator= (LegendreReQVector &&) = default;
+            LegendreReQVector & operator= (const LegendreReQVector &) = default;
+
+            // Evaluate the vector of associated Legendre functions
+            // Throws exception if algorithm doesn't terminate
+            // Returns real part if abs(z) > 1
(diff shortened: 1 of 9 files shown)


eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,266 @@
...
+
+                for (unsigned i = 0; i < len; i++)
+                {
+                    weightsp[i] = 2*(1 - zerosp[i]*zerosp[i])/((order_ + 1)*(order_ + 1)*boost::math::legendre_p<double>(order_ + 1, zerosp[i])*boost::math::legendre_p<double>(order_ + 1, zerosp[i]));
+                }
+
+                // Flip order for first half
+                for (unsigned i = 0; i < len; i++)
+                {
+                    zeros[i] = -1*zerosp[len - 1 - i];
```
dvandyk, line 89: (coding style) binary operator

eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,266 @@
...
+
+            // Evaluate the vector of associated Legendre functions
+            // Throws exception if algorithm doesn't terminate
+            // Returns real part if abs(z) > 1
+            std::array<double, order_ + 1> constexpr operator() (const double & z) const
+            {
+                std::array<double, order_ + 1> retVec;
+
+                // Standard forward recursion
+                if (abs(z) < 1)
```
dvandyk, line 152: In general, we should use ``std::abs()``, ``std::log``, and similar. In a ``.cc`` file, this can be achieved at the top by ``using std::abs`` for the whole translation unit. However, this class template is in a header file, hence we need to use ``std::abs`` everywhere.

eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,266 @@
...
+                    else if (order_ == 1)
+                    {
+                        retVec[0] = log((1 + z)/(1 - z))/2;
+                        retVec[1] = z*retVec[0] - 1;
+                    }
+                    else
+                    {
+                        retVec[0] = log((1 + z)/(1 - z))/2;
+                        retVec[1] = z*retVec[0] - 1;
+                        for(unsigned i = 2; i <= order_; i++)
```
dvandyk, line 167: (coding style) missing whitespaces after ``for`` and around semicolons:
```
for (init ; condition ; increment)
{
    ...
}
```

also applies to other most of the other for loops in the file.

eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,266 @@
...
+                            {
+                                q1 = q2;
+                            }
+                        }
+
+                        double t1 = 1;
+                        double qr = 1;
+                        for (unsigned k = 1; k <= _cut; k++)
+                        {
+                            qr *= (order_/2. + k - 1)*(k + (order_ - 1)/2.)/(z*z*k*((order_ + k) - 1./2.));
```
dvandyk, line 202: (coding style) use ``2.0`` rather than ``2.0`` for floating point literals.

eos/maths/legendre-polynomial-vector_TEST.cc
```diff
@@ -0,0 +1,73 @@
...
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[0],    0.8047189562170503,     1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[1],    0.20707843432557507,    1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[2],    0.06356699912401897,    1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[3],    0.02086520825966291,    1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[4],    0.007095922338601302,   1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[5],    0.0024668237064868373,  1.0e-10);
+                    TEST_CHECK_RELATIVE_ERROR(resQ15[6],    0.0008704965773399678,  1.0e-10);
+
+                }
+
```
dvandyk, line 69: excessive empty lines

eos/utils/omnes-factor-impl.hh
```diff
@@ -0,0 +1,307 @@
...
+#include <array>
+
+namespace eos
+{
+
+    // Allocate memory and initialise zeros & weights
+    template <unsigned order_, unsigned nints_>
+    OmnesFactor<order_, nints_>::OmnesFactor(std::array<double, nints_> & intervals) :
+        _intervals(intervals),
+        _sys(gsl_matrix_calloc(nints_*order_ + 1,nints_*order_)),
```
dvandyk, line 39: (coding style) whitespaces, ``nints_ * order_ + 1, nints_ * order_``

eos/utils/omnes-factor-impl.hh
```diff
@@ -0,0 +1,307 @@
...
+        _V(gsl_matrix_calloc(nints_*order_,nints_*order_)),
+        _S(gsl_vector_calloc(nints_*order_)),
+        _work(gsl_vector_calloc(nints_*order_)),
+        _x(gsl_vector_calloc(nints_*order_)),
+        _b(gsl_vector_calloc(nints_*order_+1))
+    {
+        // Perform pointer check
+        if (_sys == nullptr)
+            throw std::bad_alloc();
+        if (_sys == nullptr)
```
dvandyk, line 50: 1. ``_sys`` checked twice.
2. (coding style) Short if statement can appear on the same line, but still require braces, e.g., ``if (...) { throw std::bad_alloc(); }``.
3. I think it would suffice throwing ``bad_alloc`` in case that any of the allocations failed, instead of each. So maybe ``if ((_sys == nullptr) || (_V == nullptr) || ...) { throw std::bad_alloc(); }``

eos/utils/omnes-factor-impl.hh
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    // Compute sums of Eq. 58 in [M:1999A]
+    template <unsigned order_, unsigned nints_>
+    std::array<double, order_> OmnesFactor<order_, nints_>::LQSum(const double & z)
+    {
+        std::array<double, order_> retVec;
+        LegendreReQVector<order_-1> LQV;
+        LegendrePVector<order_-1> LPV;
+        std::array<double, order_> Qjz = LQV(z);
```
dvandyk, line 187: (coding style) Try to leave an empty line between object declarations and code.

eos/utils/omnes-factor-impl.hh
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        _solved = true;
+        gsl_blas_dgemv(CblasNoTrans, 1., _sys, _x, -1., _b);
+        _err = gsl_blas_dnrm2(_b);
+    }
+
+    // Compute sums of Eq. 58 in [M:1999A]
+    template <unsigned order_, unsigned nints_>
+    std::array<double, order_> OmnesFactor<order_, nints_>::LQSum(const double & z)
```
dvandyk, line 182: (coding style) This member function is not a named constructor, correct? In that case, its name should be ``l_q_sum`` or similar.

dvandyk (changes requested):
Good work. I like the diligence with the allocations. Most of the requested changes are for coding style reasons.

eos/utils/omnes-factor-impl.hh
mreboud: I think that this one was directly taken from my kmatrix implementation, so I am the culprit!

Herren: Good catch, the second one should be _sys2...

eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,266 @@
...
+ */
+
+#ifndef EOS_GUARD_EOS_MATHS_LEGENDRE_POLYNOMIAL_VECTOR_HH
+#define EOS_GUARD_EOS_MATHS_LEGENDRE_POLYNOMIAL_VECTOR_HH 1
+
+#include <array>
+#include <vector>
+#include <complex>
+
+#include <boost/math/special_functions/legendre.hpp>
```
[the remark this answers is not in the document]
Herren: Added a new commit where I add it in doc/installation.rst. Does it need to go elsewhere?

eos/utils/omnes-factor-impl.hh
```diff
@@ -0,0 +1,291 @@
...
+        _work(gsl_vector_calloc(nints_ * order_)),
+        _x(gsl_vector_calloc(nints_ * order_)),
+        _b(gsl_vector_calloc(nints_ * order_ + 1))
+    {
+        // Perform pointer check
+        if ((_sys == nullptr) || (_sys2 == nullptr) || (_V == nullptr) || (_S == nullptr) || (_work == nullptr) || (_x == nullptr) || (_b == nullptr)) { throw std::bad_alloc(); }
+
+        // Obtain zeros of Legendre Polynomials and compute Gauss-Legendre weights
+        LegendrePVector<order_> lp;
+        lp.gauss_legendre(_zeros, _weights);
```
dvandyk, line 52: Both ``_zeros`` and ``_weights`` are of type ``std::array<double, ...>``. ``std::array`` default-initializes its elements if the default constructor is used. Default-initialization for a non-class type (like double) is undefined behaviour.

Please initialize ``_zeros`` and ``_weights``.

eos/utils/omnes-factor.hh
```diff
@@ -0,0 +1,127 @@
...
+            std::array<double, order_> _weights;
+
+            // Vector containing zeros of Legendre Polynomials
+            std::array<double, order_> _zeros;
+
+            // Vector containing solution weights
+            std::array<double, nints_ * order_> _sol;
+
+            // Error of approximation
+            double _err = -1.;
```
dvandyk, line 61: Sorry for not spotting this earlier. Please put this into the constructor.
(coding style) ``-1.`` -> ``-1.0``.

eos/utils/omnes-factor.hh
```diff
@@ -0,0 +1,127 @@
...
+
+            // Did we already compute a solution?
+            bool _solved = false;
```
dvandyk, line 64: Is this purely binary? Can the status change from ``_solved == true`` to ``_solved == false``? If not, let's solve this in the constructor. Being constructed then serves the same purpose :-)

eos/utils/omnes-factor.hh
```diff
@@ -0,0 +1,127 @@
...
+            // Prepare the system of equations, Omega(bc_pos) = 1
+            void prep_sys(const double & bc_pos);
+
+            // Solve the system of equations
+            void solve_sys();
+
+            // Return weights
+            std::array<double, nints_ * order_> get_weights()
+            {
+                if(!_solved) {
```
dvandyk, line 92: (coding style) ``if (! _solved)``.

eos/maths/legendre-polynomial-vector.hh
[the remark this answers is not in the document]
dvandyk: Since the test cases pass in our docker images: probably not. Nevertheless, it would be nice to include it expliciltly when building Boost. I have opened an issue: eos/docker.io/issues/19

eos/maths/legendre-polynomial-vector.hh
```diff
@@ -0,0 +1,267 @@
...
+    template <unsigned order_>
+    class LegendreReQVector
+    {
+        private:
+            const double _eps;
+            const unsigned _cut;
+
+        public:
+            LegendreReQVector() :
+                _eps(1e-14),
```
dvandyk, line 128: A comment on both values would be nice, to judge why they have been chosen like this.

dvandyk (changes requested):
A few more coding style things (apologies), and some undefined behaviour in the ``OmnesFactor`` constructor.

I am not so happy with the use of ``_solved``, which we might be able to avoid (see my question in the source).

eos/utils/omnes-factor.hh
Herren: Good point, after solve_sys or set_weights is called it will be true no matter what.

dvandyk (approved):

PR #895, opened 2024-08-13
[maths] Fix undefined behaviour in FlatMeasure.

Herren:
The warning caused in FlatMeasure in szego-polynomial.hh is caused by one faulty comparison.
Comparing with literature, e.g. the paragraph below Eq. B.6 in [https://arxiv.org/abs/2305.06301](2305.06301), shows that i (k in 2305.06301) is smaller than the order, not smaller or equal.


mreboud (approved):
Absolutely correct. Thank you @Herren!

Issue #868, opened 2024-06-15
Add new task ``report``


dvandyk:
Add a new task ``report`` to generate [jinja](https://jinja.palletsprojects.com/en/3.1.x/)-based reports of the ongoing or completed analysis.

- [x] add proof-of-concept report that generates a LaTeX file (to be compiled to PDF) summarizing the steps of the analysis.
- [ ] ensure that the report includes a list of references used in the course of the entire analysis.

PR #901, opened 2024-08-26
Add a new `report` task 

lorenzennio:
Solves #868, by
- extending `eos.Mode` properties to include `global_chi2` and `dof`, which are automatically exported by `eos.find_mode`
- adding `report` task, with `_get_modes` function to get all modes of posteriors
- adding a jinja template `template.jinja` to `examples/`


dvandyk (approved):
Nice work @cbolognani & @lorenzennio!

PR #907, opened 2024-08-28
Automatize updating the minimum autoconf and Python versions

dvandyk:


doc/Makefile
```diff
@@ -29,5 +29,13 @@ html: source Makefile
 %.rst: %.rst.py %.rst.jinja
 	PYTHONPATH="./:$$PYTHONPATH" python3 $< > $@
 
+%.rst: %.rst.in
+	MIN_AUTOCONF_VERSION=$$(sed -n -e 's/\s*AC_PREREQ(\[\(.*\)\])/\1/p' ../configure.ac) ; \
+	MIN_PYTHON_VERSION=$$(sed -n -e 's/\s*AM_PATH_PYTHON(\[\(.*\)\])/\1/p' ../configure.ac) ; \
```
MJKirk, line 34: Your regexp should be
`AC_PREREQ(\(\[.*\]\))` (i.e. square and round brackets swapped)
and equivalent for Python, as in the `configure.ac` the versions are 
`AC_PREREQ([2.69])` and `AM_PATH_PYTHON([3.9])`

MJKirk (changes requested):

doc/Makefile
dvandyk: @MJKirk's comment is correct for ``sed -E``. I am using ``sed -e``, in which the roles of ``\(\)`` and ``()`` are interchanged. :upside_down_face:

MJKirk: You win this round

MJKirk (approved):

Issue #905, opened 2024-08-28
Unclear documentation on eos-analysis 


MJKirk:
Looking at the documentation for [find-mode](https://eos.github.io/doc/reference/command-line-interface.html#find-mode), it is not clear that various options are mutually exclusive, only in the code:
https://github.com/eos/eos/blob/dd31f071c7c77572c5743c8e504ca14baff9549e/python/eos/tasks.py#L138-L148

The description of the `start-point` option for [sample-mcmc](https://eos.github.io/doc/reference/command-line-interface.html#sample-mcmc) does not make clear it expects a comma separated string of floats.

Similarly, from-point in [find-mode](https://eos.github.io/doc/reference/command-line-interface.html#find-mode) should say something similar.
Although I think there is actually also a bug here, in that the argparser doesn't split up the string given into a list before passing it to the `find_mode` function - ~~haven't checked this yet though~~ I am now sure this is a bug: compare
https://github.com/eos/eos/blob/dd31f071c7c77572c5743c8e504ca14baff9549e/src/scripts/eos-analysis#L330
in `find-mode` with
https://github.com/eos/eos/blob/dd31f071c7c77572c5743c8e504ca14baff9549e/src/scripts/eos-analysis#L148
in `sample-mcmc`

ery-mcpartland:
The bug is fixed but the documentation needs updating

PR #908, opened 2024-08-28
[script] Fix bug in find-mode arg parsing

ery-mcpartland:


MJKirk:
This fixes the bug I mention at the end of #905

PR #916, opened 2024-08-30
Add Transform prior to distinguish between local modes

cbolognani:


eos/statistics/log-prior.cc
```diff
@@ -753,6 +756,259 @@ namespace eos
...
+                // inputs
+                gsl_vector * const _shift;
+                gsl_matrix * const _transform;
+                gsl_vector * const _min;
+                gsl_vector * const _max;
+
+                // the normalization constant of the density
+                double _value;
+
+                // cholesky matrix of covariance, inverse of the cholesky matrix, and inverse of covariance
```
dvandyk, line 780: This comment is outdated. It lists 3 matrices, but only 2 matrix pointers are declared below.

eos/statistics/log-prior.cc
```diff
@@ -40,6 +40,8 @@
 
 #include <config.h>
 
+#include <iostream>
```
dvandyk, line 43: You had added this for debugging, right? Still need to remove it.

eos/statistics/log-prior.cc
```diff
@@ -753,6 +756,259 @@ namespace eos
...
+                        _parameters.push_back(param);
+                        _varied_parameters.push_back(param);
+                    }
+
+                    // cholesky decomposition (informally: the sqrt of the transform matrix)
+                    // the GSL matrix contains both the cholesky and its transpose, see GSL reference, ch. 14.5
+                    cholesky();
+                    invert_transform();
+                    _value -= log_determinant();
+                    for (size_t i = 0; i < _transform->size1; i++) {
```
dvandyk, line 832: This debug output snuck into the branch :smile:

eos/statistics/log-prior.cc
```diff
@@ -753,6 +756,259 @@ namespace eos
...
+                        gsl_vector_set(_tmp2, i, _parameters[i].evaluate());
+                    }
+
+                    // shift: _tmp2 <- _tmp2 - _shift
+                    gsl_vector_sub(_tmp2, _shift);
+
+                    // transform: _tmp <- _transform_inv * _tmp2
+                    gsl_blas_dgemv(CblasNoTrans, 1.0, _transform_inv, _tmp2, 0.0, _tmp);
+
+                    // compute cdfs for uniform distributions in _tmp
```
dvandyk, line 924: This comment does not reflect what we are doing in ``operator()``.

eos/statistics/log-prior.cc
```diff
@@ -753,6 +756,259 @@ namespace eos
...
+
+                    gsl_vector * max = gsl_vector_alloc(_max->size);
+                    gsl_vector_memcpy(max, _max);
+
+                    return LogPriorPtr(new priors::Transform(parameters, _names, shift, transform, min, max));
+                }
+
+                virtual void sample()
+                {
+                    // generate multivariate uniform sample in _tmp
```
dvandyk, line 956: This comment was taken from ``MultivariateGaussian`` and does not apply here.

eos/statistics/log-prior_TEST.cc
```diff
@@ -284,6 +284,24 @@ class LogPriorTest :
                 TEST_CHECK_NEARLY_EQUAL(inverse_cdf(poisson_prior, param, 0.95) * k, 16.9622192357219,   eps);
             }
 
+            // Transform prior (w/ infinite support)
```
dvandyk, line 287: Hm, this is not true, right? This prior has finite support. I would remove the ``(w/ infinite support)``.

eos/statistics/log-prior_TEST.cc
```diff
@@ -284,6 +284,24 @@ class LogPriorTest :
...
+                // use factory
+                std::vector<double> shift     = {0.0,0.0};
+                std::vector<std::vector<double>> transform = {{ 0.707106, 0.707106 }, {-0.707106, 0.707106}};
+                std::vector<double> min       = {-2.0,-2.0};
+                std::vector<double> max       = {2.0,2.0};
+                LogPriorPtr transform_prior = LogPrior::Transform(parameters, {"scnuee::Re{cVL}","scnuee::Re{cVR}"}, shift, transform, min, max);
+
+                parameters["scnuee::Re{cVL}"] = 0.0;
+                parameters["scnuee::Re{cVR}"] = 0.0;
+                TEST_CHECK_NEARLY_EQUAL((*transform_prior)(), -2.77259, 5.0e6*eps);
```
dvandyk, line 298: Can you specify the precision in absolute terms, rather than relative to ``eps``? I would simply use ``1.0e-5`` as the precision argument.

dvandyk (changes requested):
This looks nice, thank you @cbolognani! I have some nitpicky comments / requests for changes. Mostly adjusting comments, and removing debug output.

dvandyk (approved):

Issue #915, opened 2024-08-30
find_mode task does not throw an error when mutually exclusive options are given 


smeiser:
When running
`import eos`
`
bfp, _ = eos.find_mode('analysis-draft.yaml', "no-fsfd-WET1-Red", base_directory="data", optimizations=10, start_point=[-0.041469, -0.899772, 0.011126, 0.194904], importance_samples=True)
`
I get a type error:
![image](https://github.com/user-attachments/assets/275bb930-94de-462d-b122-dbb1aab6aec9)

However, eos should throw an error since the two options are mutually exclusive.

PR #918, opened 2024-08-30
[python] Fix errors for mutually incompatible options in find_modes

ery-mcpartland:
This solves issue #915

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -143,11 +143,17 @@ def find_mode(analysis_file:str, posterior:str, base_directory:str='./', optimiz
     if not seed is None and not chain is None:
         raise ValueError('The arguments seed and chain are mutually exclusive')
 
-    if not chain is None and not start_point is None:
+    if not seed is None and not importance_samples is None:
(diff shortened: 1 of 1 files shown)


python/eos/tasks.py
```diff
@@ -143,10 +143,16 @@ def find_mode(analysis_file:str, posterior:str, base_directory:str='./', optimiz
...
+
+    if not start_point is None and not chain is None:
         raise ValueError('The arguments chain and start_point are mutually exclusive')
 
-    if not importance_samples is None and not start_point is None:
+    if not start_point is None and not importance_samples is None:
+        raise ValueError('The arguments importance_samples and start_point are mutually exclusive')
+
+    if not chain is None and not importance_samples is None:
         raise ValueError('The arguments importance_samples and start_point are mutually exclusive')
```
MJKirk, line 156: You've copied and pasted the error message without changing it
`importance_samples and chain are mutually exclusive`
for the last one

MJKirk (changes requested):
I don't like the non PEP8 style `if not x is None` instead of `if x is not None` but I won't fight over it

MJKirk (approved):

PR #926, opened 2024-09-13
Add option for the used sampling procedure in the nested sampling task

smeiser:


dvandyk (approved):
Thank you @smeiser, this looks good!

PR #943, opened 2024-10-22
Create a ``clang-format``-based coding style, apply it to ``test``, and check in pre commit stage

dvandyk:


mreboud (approved):
That's really nice, I am really looking forward applying this to other directories!

PR #945, opened 2024-10-23
Generalize B-meson LCDAs to heavy-meson LCDAs and allow for multi-loop alpha_s running

dvandyk:


eos/form-factors/heavy-meson-lcdas-flvd2022.hh
```diff
@@ -18,38 +18,46 @@
...
+#include <eos/form-factors/heavy-meson-lcdas.hh>
+#include <eos/models/model.hh>
 #include <eos/utils/diagnostics.hh>
 #include <eos/utils/parameters.hh>
 #include <eos/utils/options.hh>
 #include <eos/utils/options-impl.hh>
 #include <eos/utils/reference-name.hh>
 
 #include <array>
+#include <map>
```
MJKirk, line 33: Was it intentional to include it here in the .hh vs in the .cc ? Since map is only used in the .cc as far as I can see

dvandyk: Cheers, good catch. Fixed!

MJKirk (approved):

PR #950, opened 2024-11-25
Add infrastructure to run reproducible analyses with Snakemake

dvandyk:
This is the EOS side of changes. An example of a snakemake file utilizing the new tasks has been posted on the Discord server and will be merged into the analysis template repository.

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -760,6 +760,17 @@ def validate(analysis_file:str):
     analysis_file.validate()
 
 
+@task('list-steps', '', logfile=False)
+def list_steps(analysis_file:str):
+    """
+    Lists the steps in the analysis file.
+
+    :param analysis_file: The name of the analysis file that shall be inspected`.
+    :type analysis_file: str or :class:`eos.AnalysisFile`
+    """
+    return [step.id for step in analysis_file._steps]
+
+
 class Executor:
     _factory_methods = {}
 
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -797,39 +797,3 @@ def list_step_dependencies(analysis_file:str, id:str):
         raise ValueError(f'Step with id \'{id}\' not found in analysis file')
 
     return steps[id].depends_on
-
-
-class Executor:
-    _factory_methods = {}
-
-    def __init__(self, steps, dry_run):
-        self._steps = steps
-        self._dry_run = dry_run
-
-    @staticmethod
-    def register(name, type):
-        Executor._factory_methods.update({name: type})
-
-    @staticmethod
-    def make(executor, **kwargs):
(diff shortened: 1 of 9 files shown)


python/eos/tasks.py
```diff
@@ -516,10 +516,10 @@ def predict_observables(analysis_file:str, posterior:str, prediction:str, base_d
 
 
 # Run analysis steps
-@task('run', '', modules=['networkx'])
-def run(analysis_file:str, base_directory:str='./', dry_run:bool=False, executor:str='serial'):
+@task('run', '')
+def run(analysis_file:str, id:str, base_directory:str='./', dry_run:bool=False):
     """
-    Runs a list of predefined steps recorded in the analysis file.
+    Runs one out of a list of predefined steps recorded in the analysis file.
```
mreboud, line 522: The comment should then be `#Run one analysis step` in line 518.
Should we also add a `run-all` task?

python/eos/tasks.py
```diff
@@ -760,37 +770,33 @@ def validate(analysis_file:str):
...
 
-    @staticmethod
-    def make(executor, **kwargs):
-        if executor not in Executor._factory_methods:
-            raise ValueError(f'Task "run" encountered invalid executor "{executor}"')
+    for step in analysis_file._steps:
+        result.append(step.id)
 
-        return Executor._factory_methods[executor](**kwargs)
+    return result
```
mreboud, line 786: Why not directly
```python
return [step.id for step in ...]
```
?

python/eos/tasks.py
```diff
@@ -760,37 +770,33 @@ def validate(analysis_file:str):
...
+@task('list-step-dependencies', '', logfile=False)
+def list_step_dependencies(analysis_file:str, id:str):
+    """
+    Lists all steps required to be completed before the given step can be executed.
 
-    def run(self):
-        pass
+    :param analysis_file: The name of the analysis file that shall be inspected`.
+    :type analysis_file: str or :class:`eos.AnalysisFile`
+    """
```
mreboud, line 796: the description of `id` is missing

python/eos/tasks.py
dvandyk: > The comment should then be `#Run one analysis step` in line 518.
Yes!

> Should we also add a `run-all` task?
No, since we might get the ordering (dependencies)  wrong!

dvandyk: Done.

python/eos/tasks.py
dvandyk: Yes, done!

python/eos/tasks.py
dvandyk: thx, fixed!

mreboud (approved):

Issue #903, opened 2024-08-27
Interpolation issue for `uncertainty` plotter


lorenzennio:
There is an interpolation issue when plotting uncertainties in regions of high gradient. @mreboud mentioned Runge's phenomenon.

One (extreme) example plot:
![image](https://github.com/user-attachments/assets/1f070f5f-3bc5-4ef8-aca7-d5fed5b99c91)

where the actual points are:
![image](https://github.com/user-attachments/assets/99943239-cad7-4b42-9f86-6d8cc9463075)

Potential solutions could be to disable continuous plotting in the `uncertainty` plotter or to warn the user about this.

mreboud:
I think this is not a problem we should fix on the EOS side. What we can do is:

1. Add a comment, maybe in the FAQ, saying that this is due to Runge's phenomenon and perhaps a couple of suggestions to eliminate the oscillations (reducing the number of points, using Chebyshev-compatible points, ...).
2. Add an option to turn the spline interpolation off

@dvandyk, any thoughts?
I am slightly surprised that such oscillations appear with CubicSplines though...

MJKirk:
My vote would be to simply replace the CubicSpline interpolation with a linear interpolation (or offer both, but defaulting to linear), and also offer an option to show the points being plotted on top. 
Then if the curve isn't smooth enough, the user has to produce more data points, rather than us trying to magic up a smooth curve out of potentially not smooth data.

mreboud:
> My vote would be to simply replace the CubicSpline interpolation with a linear interpolation (or offer both, but defaulting to linear), and also offer an option to show the points being plotted on top. Then if the curve isn't smooth enough, the user has to produce more data points, rather than us trying to magic up a smooth curve out of potentially not smooth data.

I like this suggestion.

Herren:
> My vote would be to simply replace the CubicSpline interpolation with a linear interpolation (or offer both, but defaulting to linear), and also offer an option to show the points being plotted on top. Then if the curve isn't smooth enough, the user has to produce more data points, rather than us trying to magic up a smooth curve out of potentially not smooth data.

I fully agree, having both options available would be great!

PR #952, opened 2024-11-26
Improves uncertainty band plotting

MJKirk:
A fix for #903 - I've tried it quickly, see below
image
image

### DIFF
--- a/python/eos/plot/plotter.py
+++ b/python/eos/plot/plotter.py
@@ -560,6 +560,8 @@ class Plotter:
            If ``'median'`` is provided, the band's median line is drawn.
            If ``'area'`` is provided, the band's areas are filled.
            Defaults to ``['area', 'outer', 'median']``.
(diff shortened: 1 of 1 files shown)


mreboud (approved):
I like it. I had a long discussion with myself about whether we should make `cubic` the default. But I understand that the user facing issue 903 wouldn't know where it came from. While seeing the bad linear interpolation makes it clear that cubic interpolation is needed. So in the end, I agree.

python/eos/plot/plotter.py
```diff
@@ -664,6 +672,9 @@ def plot(self):
             elif 'median' in self.band:
                 self.plotter.ax.plot(xvalues, ovalues_central,                       alpha=self.alpha, color=self.color, ls=self.style, label=self.label, lw=self.lw)
 
+            if self.plot_data:
+                self.plotter.ax.scatter(self.xvalues, _ovalues_central,              alpha=self.alpha, color=self.color, ls=self.style, lw=self.lw)
```
mreboud, line 676: Do we really need `ls` for a scatter plot?

MJKirk:
Good catch, we don't

Issue #956, opened 2024-11-28
Typo in installation doc 


mreboud:
In the installation instructions:
```bash
export PREFIX=$VIRTUAL_ENV
```
should read
```bash
export PREFIX="$VIRTUAL_ENV"
```

Issue #954, opened 2024-11-27
Improve documentation by mentioning EOS_VERBOSITY variable


MJKirk:
Might be worth mentioning early on in the docs, e.g. to suggest people set it to a higher number while learning how EOS works

Issue #953, opened 2024-11-27
Broken link in documentation


MJKirk:
https://eos.github.io/doc/reference/analysis-file-format.html#predictions
See `name` after "The observables accept two keys"
https://github.com/eos/eos/blob/0a20d42f4746ef04825813cf70c4b0245028f081/doc/reference/analysis-file-format.rst?plain=1#L213

PR #958, opened 2024-11-29
Improvement to the documentation

### LINKED ISSUE #938: Test the latex representation of units in `units.cc`
Solved it - I forgot to add the latex representation of fm^2 in `units.cc`. Should add a test for this

_Originally posted by @MJKirk in https://github.com/eos/eos/issues/936#issuecomment-2411097398_

mreboud:
This closes #938, #953, #954 and #956.

### DIFF
--- a/python/eos_TEST.py
+++ b/python/eos_TEST.py
@@ -138,7 +138,6 @@ class PythonTests:
         for obs in con.observables():
             print(obs.name())
 
-
     """
     Check if an instance of Model can be created, and if we can compute the
     running MSbar mass of the b quark.
@@ -162,6 +161,22 @@ class PythonTests:
                 raise TestFailedError('internal error')
         except:
             raise TestFailedError('cannot determine running b quark mass')
+
+    """
+    Check if the latex strings of the Units are valid.
+    """
+    def check_009_Units(self):
+        from eos import Unit
+        import matplotlib
+        from matplotlib import texmanager
+
+        for attr, value in Unit.__dict__.items():
+            if attr[0] != "_" and attr != "latex":
+                try:
+                    s = value.__func__().latex()
+                    matplotlib.texmanager.TexManager.get_text_width_height_descent(f"${s}$", 1)
+                except:
+                    raise TestFailedError('invalid latex string for unit ' + attr)
 
 
 class LoggingTests(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


python/eos_TEST.py
```diff
@@ -4,6 +4,9 @@
 import inspect
 import os
 
+import matplotlib.pyplot
+import matplotlib.texmanager
```
mreboud, line 8: hmm, I don't know why this is here. Probably due to some VScode stupid auto-include....

python/eos_TEST.py
```diff
@@ -163,6 +165,21 @@ def check_008_Model(self):
...
+    """
+    def check_009_Units(self):
+        from eos import Unit
+        import matplotlib
+        from matplotlib import texmanager
+
+        for attr, value in Unit.__dict__.items():
+            if attr[0] != "_" and attr != "latex":
+                try:
+                    matplotlib.texmanager.TexManager.get_text_width_height_descent(f"${value().latex()}$", 1)
```
MJKirk, line 179: Is there a reason you can't just call value().latex() and catch errors? Without having to involve matplotlib? (I can't check it myself as for some reason for me the whole thing fails with a `TypeError: 'staticmethod' object is not callable` error)

mreboud: `value().latex()` only returns the latex string. So no TeX involved. If you want to check that the latex string is valid, you need to call a TeX parser. I found this to be the simplest ^^

dvandyk: On Python 3.9, ``value()`` raises the type error mentioned by @MJKirk. The reason is that the method is a static method created by the equivalent of a ``@staticmethod`` decorat. E.g., the name ``Unit.Undefined`` refers to the decorator for the static method, not the method itself. On Python 3.9, we can circumvent the problem by using ``value.__func__()`` instead of ``value()``. Here's my proposed solution:
```diff
diff --git a/python/eos_TEST.py b/python/eos_TEST.py
index c7e939bc..e28107d9 100644
--- a/python/eos_TEST.py
+++ b/python/eos_TEST.py
@@ -176,7 +176,8 @@ class PythonTests:
         for attr, value in Unit.__dict__.items():
             if attr[0] != "_" and attr != "latex":
                 try:
-                    matplotlib.texmanager.TexManager.get_text_width_height_descent(f"${value().latex()}$", 1)
+                    s = value.__func__().latex()
+                    matplotlib.texmanager.TexManager.get_text_width_height_descent(f"${s}$", 1)
                 except:
                     raise TestFailedError('invalid latex string ' + value().latex())
 ```

MJKirk (approved):
LGTM

dvandyk (approved):

Issue #961, opened 2024-12-02
Various issues with new snakemake stuff


MJKirk:
So far trying to use snakemake for our pion FF project, I have come across the following issues:

- Any `base_directory` specified in the AnalysisFile gets overridden 
https://github.com/eos/eos/blob/3cea9ad265fecaf676b08f643488ab7d779abda5/python/eos/tasks.py#L546-L554
by the arguments of the `run` task, which has a default value of the current directory.
I think swapping the order of the dictionaries would make sense.

- task arguments in the analysis file have to be called their internal names not their CLI option name (e.g. `nlive` rather than `number-of-live-points` or `NLIVE`), which can only be found by looking at the source code.  Better documentation is probably enough to fix this.

- When allowing snakemake to run tasks in parallel, the outputs from dynesty appear simultaneously and overwrite each other (e.g. the progress bar from sample-nested).
It looks like the `print_progress` argument in dynesty can turn this ouput off, so maybe we want to use that when running things from snakemake (or at least when doing things in parallel)?

- Danny changed the `StepComponent` parser class in https://github.com/eos/eos/commit/fdeae95f65f974c75ca59dc4739adbbb3714a181, and now the `AnalysisFile.steps` function doesn't work - this function was introduced in PR #521 as a first attempt at allowing for reproducible analyses, so perhaps could be removed if we think snakemake is the better way going forward.

MJKirk:
From the dev meeting:
If you have multiple similar tasks in a `task`, the code gets repetitive:
```
steps:
  - title: 'Sample from CKM-all posterior'
    id: 'CKM-all.sample'
    depends_on: []
    tasks:
      - task: 'sample-nested'
        arguments:
          posterior: 'CKM-all'
          bound: 'multi'
          nlive: 100
          dlogz: 9.0
          maxiter: 4000
      - task: 'sample-nested'
        arguments:
          posterior: 'CKM-all2'
          bound: 'multi'
          nlive: 100
          dlogz: 9.0
          maxiter: 4000
```
Suggestion is to add `default_arguments` below tasks that are added to all tasks. e.g.
```
steps:
  - title: 'Sample from CKM-all posterior'
    id: 'CKM-all.sample'
    depends_on: []
    tasks:
      - default_arguments:
          bound: 'multi'
          nlive: 100
          dlogz: 9.0
          maxiter: 4000
      - task: 'sample-nested'
        arguments:
          posterior: 'CKM-all'
      - task: 'sample-nested'
        arguments:
[...]

PR #965, opened 2024-12-10
Some snakemake improvements

MJKirk:
Address some of the issues I raised in #961:
- [x] Allow `default_arguments` as suggested by Nico
- [x] Remove `AnalysisFile.steps` function

Still to be done:
- [x] Potentially allow using CLI argument names in the AnalysisFile. There's a function https://github.com/eos/eos/blob/103d61cd349b469baf4511caf075d2973f601ad7/python/eos/analysis_file.py#L232-L232
which isn't currently used, but looks like it was designed to do something similar for the OG `steps` section of the AnalysisFile, which I want to think about.
- [x] Turn off / alter the dynesty output when running from snakemake in parallel <-- this is most easily achieved by redirecting stderr in the snakemake file


python/eos/analysis_file_description.py
```diff
@@ -239,6 +240,9 @@ def __post_init__(self):
     @classmethod
     def from_dict(cls, **kwargs):
         _kwargs = _copy.deepcopy(kwargs)
+        if "default_arguments" in kwargs:
```
dvandyk, line 314: Hm, I would prefer to apply this logic in the execution, i.e., the ``run`` task, not the description.

src/scripts/eos-analysis_TEST
```diff
@@ -10,15 +10,15 @@ echo using PYTHON=${PYTHON} >&2
...
+## Run the Analysis's steps automatically ##
 ###########################################
 echo running analysis steps test \(CKM+FF.find-mode\)... >&2
 $PYTHON ${SOURCE_DIR}/eos-analysis \
     run \
     CKM+FF.find-mode \
     -f ${SOURCE_DIR}/eos-analysis_TEST.d/analysis.yaml
 echo ... success >&2
-echo running analysis steps test \(CKM+ff.sample\)... >&2
+echo running analysis steps test \(CKM+FF.sample\)... >&2
```
dvandyk, line 21: how did that happen?

src/scripts/eos-analysis
```diff
@@ -629,7 +629,7 @@ def main():
 
             args.cmd(args)
     except Exception as e:
-        print(colored('✖ Encountered an unrecoverable error', 'red', attrs=['bold']), f'{e}')
+        print(colored('✖ Encountered an unrecoverable error:', 'red', attrs=['bold']), f'{e}')
```
dvandyk, line 641: Should we add a line break here?

python/eos/analysis_file.py
```diff
@@ -108,6 +108,8 @@ def __init__(self, analysis_file):
         if 'steps' not in self.input_data:
             self._steps = {}
         else:
+            if len(self.input_data['steps']) != len({s['id'] for s in self.input_data['steps']}):
```
dvandyk, line 111: Yes!

python/eos/analysis_file.py
```diff
@@ -425,7 +427,12 @@ def validate(self):
                     known_params[param]
                 except RuntimeError:
                     messages.append(f"Error in prediction {p_name}: Fixed parameter '{param}' not known to EOS")
-
+        # Check all posteriors named in steps exist
```
dvandyk, line 279: Yes as well!

python/eos/analysis_file_description.py
```diff
@@ -201,6 +201,77 @@ def __post_init__(self):
             raise ValueError(f'Task \'{self.task}\' is not a valid task')
         task = eos.tasks._tasks[self.task]
 
+        # Mapping between CLI argument names and internal names for individual steps
```
dvandyk, line 204: steps -> tasks

dvandyk (commented):
Some comments.

src/scripts/eos-analysis_TEST
MJKirk: A certain DvD mistyped when he wrote this line a few weeks ago? It's just the echo message saying what's happening so no actual issue

src/scripts/eos-analysis
MJKirk: I think it's fine as is, but I don't have a strong opinion

python/eos/analysis_file_description.py
MJKirk: By doing it here, the `__post_init__` step of `TaskComponent` gives an error if you have `default_arguments` that don't apply to the tasks. If you want to do it in `run`, you have to do some checks there, else the task will fail in the case of wrong arguments.

dvandyk: At the meeting, agreed to move this logic into ``AnalysisFile.validate()``.

MJKirk:
Still to do: some documentation (at least in the Analysis File Description page)

python/eos/analysis_file_description.py
```diff
@@ -201,6 +201,77 @@ def __post_init__(self):
...
+            ('sample-pmc', 'l'): 'pmc_lookback', ('sample-pmc', 'pmc-lookback'): 'pmc_lookback', ('sample-pmc', 'PMC_LOOKBACK'): 'pmc_lookback',
+            ('sample-pmc', 'p'): 'initial_proposal', ('sample-pmc', 'initial-proposal'): 'initial_proposal', ('sample-pmc', 'INITIAL_PROPOSAL'): 'initial_proposal',
+            ('sample-pmc', 'S'): 'sigma_stat_test', ('sample-pmc', 'sigma-stat-test'): 'sigma_stat_test', ('sample-pmc', 'SIGMA_STAT_TEST'): 'sigma_stat_test',
+            # sample-nested
+            ('sample-nested', 'B'): 'bound', ('sample-nested', 'target-bound'): 'bound', ('sample-nested', 'BOUND'): 'bound',
+            ('sample-nested', 'n'): 'nlive', ('sample-nested', 'number-of-live-points'): 'nlive', ('sample-nested', 'NLIVE'): 'nlive',
+            ('sample-nested', 'd'): 'dlogz', ('sample-nested', 'evidence-tolerance'): 'dlogz', ('sample-nested', 'DLOGZ'): 'dlogz',
+            ('sample-nested', 'm'): 'maxiter', ('sample-nested', 'max-number-iterations'): 'maxiter', ('sample-nested', 'MAXITER'): 'maxiter',
+            ('sample-nested', 's'): 'seed', ('sample-nested', 'use-random-seed'): 'seed', ('sample-neested', 'SEED'): 'seed',
```
MJKirk, line 235: typo: "neested"

python/eos/analysis_file_description.py
```diff
@@ -201,6 +201,77 @@ def __post_init__(self):
...
+            ('find-clusters', 'c'): 'K_g', ('find-clusters', 'clusters-per-group'): 'cluster', ('find-clusters', 'CLUSTER'): 'cluster', ('find-clusters', 'K_G'): 'K_g',
+            # predict-observables
+            ('predict-observables', 'POSTERIOR'): 'posterior',
+            ('predict-obserables', 'PREDICTION'): 'prediction',
+            ('predict-observables', 'B'): 'begin', ('predict-observables', 'begin-index'): 'begin', ('predict-observables', 'BEGIN'): 'begin',
+            ('predict-observables', 'E'): 'end', ('predict-observables', 'end-index'): 'end', ('predict-observables', 'END'): 'end',
+            # corner-plot
+            ('corner-plot', 'B'): 'begin', ('corner-plot', 'begin-parameter'): 'begin', ('corner-plot', 'BEGIN'): 'begin',
+            ('corner-plot', 'E'): 'end', ('corner-plot', 'end-parameter'): 'end', ('corner-plot', 'END'): 'end',
+            ('corner-plot', 'F'): 'format', ('corner-plot', 'format'): 'format', ('corner-plot', 'FORMAT'): 'format',
```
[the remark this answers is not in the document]
MJKirk: Same for ('find-mode', 'optimizations') optimizations and ('find-clusters', 'threshold') threshold

python/eos/analysis_file_description.py
MJKirk: How do you feel about this new version? Default arguments are kept separate, but checked in `AnalysisFile.validate`, and applied in the `run` task.
Note I've also altered how `default_arguments` are specified - they are now per task, as this keeps the mapping between CLI and internal names clean, and allows you to have multiple different tasks specified in a single step.

MJKirk: Note I am coming around to thinking we should have a specific task component dataclasses, i.e. a `SampleNestedTaskConfiguration`, ...  (similar to what is done for Priors), and so enforce more of the structure via the AnalysisFileDescription code. To be discussed...

doc/reference/analysis-file-format.rst
```diff
@@ -242,3 +244,56 @@ The following code provides a valid example of predictions.
...
+
+
+Steps
+~~~~~~~~~~~
+
+The ``steps`` key contains a list of pre-defined tasks that allow an analysis to be fully reproduced. Each step has three mandatory keys, and two optional keys:
+
+  - ``title`` (**mandatory**) --- The title of this step.
+  - ``id`` (**mandatory**) --- The unique identifier of this step.
+  - ``depends_on`` (**optional**) --- A list of step ``id``'s that need to be executed prior to this step.
```
dvandyk, line 256: That apostrophe is incorrect, isn't it?

python/eos/analysis_file.py
```diff
@@ -425,7 +277,24 @@ def validate(self):
...
+            # Check that the default arguments correspond to real tasks
+            if (unknown_tasks := step.default_arguments.keys() - eos.tasks._tasks.keys()):
+                raise ValueError(f'Step \'{step_id}\' has default arguments for unknown tasks: {unknown_tasks}')
+            for tc in step.tasks:
+                # Check all posteriors named in steps exist
+                if 'posterior' in tc.arguments:
+                    if tc.arguments['posterior'] not in self._posteriors:
+                        messages.append(f"Error in step {step_id}: Posterior '{task.arguments['posterior']}' not known to EOS")
+                # Check that default arguments in steps can be applied to all specified tasks
+                task_func = eos.tasks._tasks[tc.task]
```
dvandyk, line 293: You are not yet checking if ``tc.task`` is a valid task.

python/eos/analysis_file.py
```diff
@@ -425,7 +277,24 @@ def validate(self):
                     known_params[param]
                 except RuntimeError:
                     messages.append(f"Error in prediction {p_name}: Fixed parameter '{param}' not known to EOS")
-
+        for step_id, step in self._steps.items():
+            # Check that all the dependencies of a step exist
+            if (unknown_dependencies := step.depends_on - self._steps.keys()):
+                raise ValueError(f'Step \'{step_id}\' depends on unknown steps: {unknown_dependencies}')
```
dvandyk, line 283: Hm. Shouldn't we append this to the message rather than raising an error?

python/eos/analysis_file.py
```diff
@@ -425,7 +277,24 @@ def validate(self):
...
+            # Check that the default arguments correspond to real tasks
+            if (unknown_tasks := step.default_arguments.keys() - eos.tasks._tasks.keys()):
+                raise ValueError(f'Step \'{step_id}\' has default arguments for unknown tasks: {unknown_tasks}')
```
dvandyk, line 286: Ditto.

python/eos/analysis_file_TEST.d/valid-analysis-file.yaml
```diff
@@ -129,8 +129,8 @@ predictions:
           q2_max: 25.0
 
 steps:
-  - title: 'Sample from CKM-all posterior'
-    id: 'CKM-all.sample'
+  - title: 'Sample from CKM-all and WET-all posteriors'
+    id: 'CKM-all,WET-all.sample'
```
dvandyk, line 133: I'm not sure that I like this short hand. I used the dot as a separator for personal preference only. Not sure that everybody will.

src/scripts/eos-analysis
```diff
@@ -590,9 +595,14 @@ class CustomLogFormatter(logging.Formatter):
         if record.levelno not in self._MAP_LEVEL_TO_COLOR:
             levelno = logging.ERROR
 
-        symbol, color = self._MAP_LEVEL_TO_COLOR[record.levelno]
-        record.levelname = colored(symbol, color, attrs=['bold'])
-        record.msg = colored(record.msg, color)
+        if levelno != logging.DEBUG:
```
dvandyk, line 598: Really don't like this if here. Can't we determine this based on the contents of colour? Maybe make it ``None`` if we don't want to change the front colour?

python/eos/analysis_file.py
```diff
@@ -106,9 +106,9 @@ def __init__(self, analysis_file):
             eos.completed(f'... finished declaring {len(self._params)} custom parameters')
 
         if 'steps' not in self.input_data:
-            self._steps = []
+            self._steps = {}
```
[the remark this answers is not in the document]
MJKirk: The order they are written in the analysis file doesn't matter (in the sense I can write them in any order without changing the functionality), and using a dict rather than a list makes the code nicer. And the same argument would apply to priors and likelihoods, which we also store as dicts internally.
Also the order of dicts is guaranteed to be the same as insertion (since python 3.7 or something), so actually the order is preserved anyway.

python/eos/tasks.py
```diff
@@ -778,7 +776,7 @@ def list_steps(analysis_file:str):
     :param analysis_file: The name of the analysis file that shall be inspected`.
     :type analysis_file: str or :class:`eos.AnalysisFile`
     """
-    return [step.id for step in analysis_file._steps]
+    return analysis_file._steps.keys()
```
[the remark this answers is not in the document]
MJKirk: See above - dict order is guaranteed.

python/eos/analysis_file_description.py
```diff
@@ -227,6 +228,7 @@ class StepComponent(Deserializable):
     id:str
     tasks:list
     depends_on:list=field(default_factory=list)
+    default_arguments:defaultdict=field(default_factory=lambda: defaultdict(dict))
```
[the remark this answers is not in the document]
MJKirk: Pretty sure I got the idea from stackexchange :laughing:

python/eos/analysis_file_TEST.d/valid-analysis-file.yaml
```diff
@@ -150,9 +150,24 @@ steps:
 
   - title: 'Create corner plot for CKM-all posterior'
     id: 'CKM-all.corner-plot'
-    depends_on: ['CKM-all.sample']
+    depends_on: ['CKM-all,WET-all.sample']
```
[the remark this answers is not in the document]
MJKirk: No, because 'CKM-all,WET-all.sample' is the id for a single step which does two samplings (I was just showing off / testing the ability to use both long and short names)

doc/reference/analysis-file-format.rst
MJKirk: Yeah that's wrong :disappointed:

python/eos/analysis_file_TEST.d/valid-analysis-file.yaml
MJKirk: I'm open to different conventions. I guess the main place it matters is not here, but the example in the documentation, there we should use some sensible convention

python/eos/analysis_file.py
MJKirk: That is enforced by the `__post_init__` of `TaskComponent`, so must be true once we have loaded the analysis file successfully.

python/eos/analysis_file.py
MJKirk: Yes

python/eos/analysis_file.py
MJKirk: Ditto my response

src/scripts/eos-analysis
MJKirk: I couldn't think of a nice way to do this - the problem was the the `colored` function doesn't accept anything like `None ` as an argument, as far as I can see the only way to make it print in the default colour is just to call it with two arguments not three. 
I suppose you could do:
```python
symbol, color = self._MAP_LEVEL_TO_COLOR[record.levelno]
if color:
    record.levelname = colored(symbol, color, attrs=['bold'])
    record.msg = colored(record.msg, color)
else:
    record.levelname = colored(symbol, attrs=['bold'])
    record.msg = colored(record.msg)
 ```
 Do you prefer that?

MJKirk: You know what, it looks like `colored` does accept `None`, maybe there was some other issue. Let me double check this...

MJKirk: Seems to work fine with `None`, so I don't know why I didn't do that before. Now changed

python/eos/analysis_file.py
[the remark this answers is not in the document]
MJKirk: Yep, dict insertion order is preserved at of some version, so all good

PR #969, opened 2024-12-19
Rewrite ``prior::Transform`` to use LU decomposition

dvandyk:
- [x] Use LU decomposition rather than Cholesky decomposition (which only works for hermitian positive definite matrices)
- [x] Rather than inverting the transformation matrix, keep the LU decomposition and solve the system. This is fast and numerically more stable.

### DIFF
--- a/eos/statistics/log-prior.cc
+++ b/eos/statistics/log-prior.cc
@@ -775,9 +775,9 @@ namespace eos
                 // the normalization constant of the density
                 double _value;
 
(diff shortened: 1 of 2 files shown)


mreboud:
This seems to work, I would just like to test the prior volume and see if I can recover the log evidence of the base priors from the one of the transformed priors.

mreboud (approved):
Thanks @dvandyk!

PR #978, opened 2025-01-29
Clean up models library

dvandyk:
- [x] enforce common ordering to the WET-related parts of the model library
- [x] enable running the ``clang-format`` pre-commit hook on the model library
- [x] adapt the ``clang-format``-based coding style on the model library


eos/models/ckm.cc
```diff
@@ -111,15 +111,21 @@ namespace eos
...
+        // Charged-current semileptonic sectors (Delta C = 1)
+        SMComponent<components::WET::DCNuL>(parameters, *this),
+        SMComponent<components::WET::SCNuL>(parameters, *this),
+        // Charged-current semileptonic sectors (Delta B = 1)
         SMComponent<components::WET::UBLNu>(parameters, *this),
         SMComponent<components::WET::CBLNu>(parameters, *this),
+        // Neutral-current semileptonic sectors (Delta B = 1)
         SMComponent<components::WET::SBNuNu>(parameters, *this),
-        SMComponent<components::WET::SBCU>(parameters, *this),
+        // Hadronic sectors (Delta B = 1)
```
mreboud, line 122: Delta C = 1 as well in these two sectors :)

eos/models/model.cc
```diff
@@ -26,16 +26,12 @@
...
-    }
+    Model::~Model() {}
 
-    const std::map<Model::KeyType, Model::ValueType>
-    Model::models
-    {
-        std::make_pair("CKM",       &CKMScanModel::make),
-        std::make_pair("SM",        &StandardModel::make),
-        std::make_pair("WET",       &WilsonScanModel::make),
+    const std::map<Model::KeyType, Model::ValueType> Model::models{
```
mreboud, line 31: Do you know why the final `{` is not on a new line?

eos/models/standard-model.cc
```diff
@@ -814,22 +898,16 @@ namespace implementation
 
         // only unprimed WCs are non-zero in the SM
         // leading order in alpha_s
-        const std::array<double, 10u> lo_unprimed = {
-            -1.0 / 9.0, -2.0 / 3.0, +1.0 / 36.0, +1.0 / 6.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0
-        };
+        const std::array<double, 10u> lo_unprimed  = { -1.0 / 9.0, -2.0 / 3.0, +1.0 / 36.0, +1.0 / 6.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0 };
         // next-to-leading order in alpha_s
-        const double L = 2.0 * std::log(_mu_0__sbcu / _m_W__sbcu);
+        const double                  L            = 2.0 * std::log(_mu_0__sbcu / _m_W__sbcu);
```
mreboud, line 903: hmm --"

mreboud (changes requested):
Shouldn't the two last commits be swapped?

eos/models/ckm.cc
dvandyk: I intend to only list the "major" change of quantum number. Otherwise, we also have to account for Delta S, Delta I, etc...

eos/models/model.cc
dvandyk: clang-format see this as a plain initializer, which have their braces / parentheses put on the same line as the object name.

eos/models/standard-model.cc
dvandyk: That's a Github display problem, see the screenshot:
image

dvandyk: It is perfectly aligned:

image

dvandyk: Sorry, this is ``dbcu``, but ``sbcu`` looks the same.

mreboud: Yes, it is ugly but I understand

dvandyk:
> Shouldn't the two last commits be swapped?

I agree. Just fixed it, pushing now.

mreboud (approved):

PR #980, opened 2025-01-31
Un-hardcoded susceptibilities and resonance masses in BGL parametrization

### LINKED ISSUE #974: Un-hardcode susceptibilities and resonance masses in BGL parametrization
https://github.com/eos/eos/blob/544f0e0061ddf6f9594f054bd0651e2a58219eac/eos/form-factors/parametric-bgl1997-impl.hh#L39
https://github.com/eos/eos/blob/544f0e0061ddf6f9594f054bd0651e2a58219eac/eos/form-factors/parametric-bgl1997-impl.hh#L141

These values should obviously not be hard coded.

Maximilian0901:
Fixes #974: It is now possible to dynamically set the resonance masses and susceptibilities of the BGL parametrization to a specific value.

### DIFF
--- a/eos/form-factors/parametric-bgl1997-impl.hh
+++ b/eos/form-factors/parametric-bgl1997-impl.hh
@@ -4,6 +4,7 @@
  * Copyright (c) 2020 Danny van Dyk
  * Copyright (c) 2020 Nico Gubernari
  * Copyright (c) 2020 Christoph Bobeth
+ * Copyright (c) 2025 Maximilian Hoverath
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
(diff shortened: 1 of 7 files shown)


eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -137,6 +138,25 @@ groups:
...
+          latex:   '$m_{B_c}[1]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c[2]@BSZ2015' :
+          central: +7.250
+          min:     +7.250
+          max:     +7.250
+          latex:   '$m_{B_c}[2]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c[3]@BSZ2015' :
+          central: +0.0
```
dvandyk, line 155: Why is the default value zero for this state?

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -191,6 +211,25 @@ groups:
...
+          latex:   '$m_{B_c^*}[1]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c^*[2]@BSZ2015' :
+          central: +7.020
+          min:     +7.020
+          max:     +7.020
+          latex:   '$m_{B_c^*}[2]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c^*[3]@BSZ2015' :
+          central: +0.0
```
dvandyk, line 228: See above.

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -221,6 +260,25 @@ groups:
...
+          latex:   '$m_{B_{c,0}}[1]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c,0[2]@BSZ2015' :
+          central: +0.0
+          min:     +0.0
+          max:     +0.0
+          latex:   '$m_{B_{c,0}}[2]^\mathrm{BSZ}$'
+          unit:    'GeV'
+      'mass::B_c,0[3]@BSZ2015' :
+          central: +0.0
```
dvandyk, line 277: One more.

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -269,7 +346,7 @@ groups:
           max:     +5.61951
           latex:   '$m_{\Lambda_b}^\mathrm{BMRvD}$'
           unit:     'GeV'
-  # }}}
+  # }}}
```
dvandyk, line 349: Are these added whitespaces? If yes, please remove.

eos/form-factors/parametric-bgl1997-impl.hh
```diff
@@ -124,7 +124,7 @@ namespace eos
             _traits.m_R3_1m
         };
         for (int i = 0; i < 4; ++i) {
-            if (resonances[i] <= _traits.tp()) { // only account for resonances that are smaller than tp or larger than tm.
+            if (resonances[i] * resonances[i] <= _traits.tp()) { // only account for resonances that are smaller than tp or larger than tm.
```
dvandyk, line 127: Please refer to the states as "bound states" (which they are), not resonances (which they are not with respect to the channel considered).

I do not like the logic. Your code changes the model, depending on the value of the parameter ``tp``. This leads to discontinuites in the posterior PDF as functions of ``tp``. Instead, I would suggest to fix the number of resonances that we know or assume to know contribute. We can make this number of bound states (per $J^P$ channel) an option. This logic has the following benefits:
1. We can still account for different numbers of resonances across different models, and we can compare them through their Bayes factors.
2. For each model, the posterior is free of discontinuities.
3. For each model, lowering ``tp`` below the mass of the heaviest bound state yields ``NaN``, which naturally confines the parameter space.

dvandyk (commented):
I left some comments. Probably best to discuss this.

eos/form-factors/parametric-bgl1997-impl.hh
```diff
@@ -33,139 +34,165 @@
...
-        _mV2(power_of<2>(_mV)),
-        // default t_0 = sqrt(t_p) (sqrt(m_B) - sqrt(m_M))^2 (optimal value)
-        _t_0(p["B->D^*::t_0@BGL1997"], *this)
+                 UsedParameter(p[_par_name("T23_3")], *this)
+        }},
+        _traits(BGL1997FormFactorTraits<Process_, PToV>(p)),
+        _mB(_traits.m_B),
+        _mV(_traits.m_V),
+        t_0(_traits.t_0),
+        n_bound_states_1m(o, _options, "n-bound-states-1m"),
```
dvandyk, line 84: As discussed in the meeting this morning: it is preferable to move this logic into the FormFactorTraits class and let it provide four Blaschke factor functions, one for each possible number of resonances. Then one ``std::function`` object per channel can be constructed here to take of each of the four channels' Blaschke factors.

eos/form-factors/parametric-bgl1997.hh
```diff
@@ -27,30 +28,243 @@
...
+            std::array<UsedParameter, 4> masses_1m;
+            std::array<UsedParameter, 4> masses_1p;
+            std::array<UsedParameter, 3> masses_0m;
+            std::array<UsedParameter, 2> masses_0p;
+            UsedParameter chi_1m, chi_0p;
+            UsedParameter chi_1p, chi_0m;
+            UsedParameter chi_T_1m, chi_T_1p;
+            UsedParameter t_0;
+
+            SpecifiedOption n_bound_states_1m;
```
mreboud, line 61: I will implement an `IntegerOption` in another PR.

mreboud (approved):

dvandyk (approved):
Approved. Thank you @Maximilian0901

PR #982, opened 2025-02-11
Improvement to the ``sample-nested`` user interface

dvandyk:
- [x] add custom print function
- [x] redirect timed output (1 every 15s) to the EOS logger
- [x] support demanding minimum number of samples

### DIFF
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -726,9 +726,12 @@ class Analysis:
         .. note::
            This method requires the dynesty python module, which can be installed from PyPI.
         """
-        import dynesty
+        import dynesty, tqdm
+        from functools import partial
+
         if print_function is None:
-            print_function = dynesty.results.print_fn
+            print_function = partial(dynesty.results.print_fn, pbar=tqdm.tqdm())
+
         sampler = dynesty.DynamicNestedSampler(self.log_likelihood, self._prior_transform, len(self.varied_parameters), bound=bound, nlive=nlive, rstate = np.random.Generator(np.random.MT19937(seed)), sample=sample)
         sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress, print_func=print_function)
         return sampler.results
--- a/python/eos/analysis.py
+++ b/python/eos/analysis.py
@@ -702,7 +702,7 @@ class Analysis:
         return self._u_to_par(u)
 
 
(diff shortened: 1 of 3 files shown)


python/eos/analysis.py
```diff
@@ -702,7 +702,7 @@ def _prior_transform(self, u):
         return self._u_to_par(u)
 
 
-    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=10, print_progress=True, sample='auto'):
+    def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, miniter=0, print_progress=True, print_function=None, seed=10, sample='auto'):
```
mreboud, line 705: Can you add a docstring for `print_function`?

python/eos/analysis.py
```diff
@@ -724,9 +726,16 @@ def sample_nested(self, bound='multi', nlive=250, dlogz=1.0, maxiter=None, seed=
...
+        from functools import partial
+
+        if print_function is None:
+            print_function = partial(dynesty.results.print_fn, pbar=tqdm.tqdm())
+
         sampler = dynesty.DynamicNestedSampler(self.log_likelihood, self._prior_transform, len(self.varied_parameters), bound=bound, nlive=nlive, rstate = np.random.Generator(np.random.MT19937(seed)), sample=sample)
-        sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress)
+        sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress, print_func=print_function)
+        while sampler.results['niter'] < miniter:
+            sampler.run_nested(dlogz_init=dlogz, maxiter=maxiter, print_progress=print_progress, print_func=print_function)
```
mreboud, line 738: Uff, you are running the sampler from scratch! Why don't you use `sampler.add_batch(mode='full')` instead?

mreboud (changes requested):
Looks good, but I am not sure about the implementation of the minimum number of iterations.

python/eos/analysis.py
dvandyk: Done on the ``Analysis`` side.

python/eos/analysis.py
dvandyk: Ah, I had not seen the ``full`` mode documentation in the dynesty API. Switched!

mreboud (approved):

PR #986, opened 2025-02-13
Omnes factor improvements

Herren:
Here we improve the evaluation time of Omnès factors significantly by caching everything that can be cached easily.
Furthermore, we move all the Omnès factor logic from [utils] to [maths].


eos/utils/omnes-factor.hh
```diff
@@ -64,6 +64,11 @@ namespace eos
             // Phase input
             std::function<double(const double &)> _scattering_phase;
 
+            // Cached values necessary when evaluating the Omnes factor
+            std::array<std::array<double, order_>, nints_> _tanvals;
```
mreboud, line 68: All these array entries could be made `const double` as they are evaluated once when constructing the Omnès class, right? It would probably make the implementation more complicated for basically nothing though...

mreboud (approved):
Looks good, I just have one comment.

eos/utils/omnes-factor.hh
Herren: I dabbled a bit with this, but it doesn't really get nice... I'll revisit that when I come back to this class next time around.

mreboud: Fair enough!

dvandyk (approved):
Looks good!

PR #987, opened 2025-02-13
Integer option

### LINKED ISSUE #985: Implement an `IntegerOption`
I will implement an `IntegerOption` in another PR.

_Originally posted by @mreboud in https://github.com/eos/eos/pull/980#discussion_r1952684837_

mreboud:
Closes #985.

I checked the use cases, and it seems that a more sophisticated option with min/max is not needed.

### DIFF
--- a/eos/form-factors/parametric-bgjvd2019.cc
+++ b/eos/form-factors/parametric-bgjvd2019.cc
@@ -28,7 +28,7 @@ namespace eos
         if ("B(*)->D(*)" == prefix)
             return prefix;
 
-        if ("1" == _opt_sslp_limit.value())
+        if (_opt_sslp_limit.value())
             return "B(*)->D(*)";
 
         return prefix;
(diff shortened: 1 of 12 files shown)


eos/form-factors/parametric-bgjvd2019.cc
```diff
@@ -39,16 +39,16 @@ namespace eos
...
-        _opt_lp_zorder(o, "z-order-lp", { "2", "3", "4", "5" }, "3"),
-        _enable_lp_z3(1.0 ? _opt_lp_zorder.value() >= "3" : 0.0),
-        _enable_lp_z4(1.0 ? _opt_lp_zorder.value() >= "4" : 0.0),
-        _enable_lp_z5(1.0 ? _opt_lp_zorder.value() >= "5" : 0.0),
-        _opt_slp_zorder(o, "z-order-slp", { "1", "2" }, "2"),
-        _enable_slp_z2(1.0 ? _opt_slp_zorder.value() >= "2" : 0.0),
-        _opt_sslp_zorder(o, "z-order-sslp", { "0", "1", "2" }, "1"),
-        _enable_sslp_z1(1.0 ? _opt_sslp_zorder.value() >= "1" : 0.0),
-        _enable_sslp_z2(1.0 ? _opt_sslp_zorder.value() >= "2" : 0.0),
-        _opt_sslp_limit(o, "SU3F-limit-sslp", { "0", "1" }, "0"),
```
dvandyk, line 51: @mreboud Careful, this change to using ``BooleanOption`` modified the behaviour of existing analyses.
The reason is that ``destringify<bool>`` interprets only ``true`` as a true boolean value, and anything else is interpreted as ``false``, including ``1``. But the latter was used in lieu of ``true`` here. We need to simultaneously adapt the (private) analysis 2019-04 and any ongoing analysis.

@gubernari Could you please have a look at this as soon as possible please and adapt your ongoing analysis file?

eos/form-factors/parametric-kkvdz2022.cc
```diff
@@ -25,7 +25,7 @@ using namespace std::literals::complex_literals;
 namespace eos
 {
     KKvDZ2022FormFactors::KKvDZ2022FormFactors(const Parameters & p, const Options & o) :
-        opt_subtracted(o, "subtracted", { "off", "on" }, "off"),
+        opt_subtracted(o, options, "subtracted"),
```
dvandyk, line 28: @mreboud There is no analysis file that used this.

dvandyk (commented):
Not quite happy with the changes to using ``BooleanOption`` yet.

@mreboud @gubernari Would it make sense to expand ``destringify<bool>`` to also accept other values than ``true``? If yes, then we should probably issue a warning that the user should change their inputs from ``true`` / ``false``?

eos/form-factors/parametric-bgjvd2019.cc
mreboud: @dvandyk thanks for spotting this; I thought `1.0` would be interpreted as `true`.
@gubernari Understood, although this should be completely safe.

eos/form-factors/parametric-kkvdz2022.cc
mreboud: Ok, but I will not clean KKvDZ in this PR. If this parametrization is still useful, it can be cleaned, otherwise we can remove it

eos/form-factors/parametric-bgjvd2019.cc
mreboud: Concerning your question, I would say no. Let's change the analysis file instead and make sure we use BooleanOption in the future.

MJKirk:
If you are feeling generous Meril, you could clean up the KSvD2025 $0 \to K \pi$ form factor code with this new option as well :smile:

mreboud:
This PR will be merged once Nico's analysis is published

dvandyk:
From the meeting today:
1. Rebase on top of master.
2. Modify ``destringify<bool>`` to only accept ``true`` or ``false`` and throw and error when encountering anything else.

PR #988, opened 2025-02-18
Clean up nonleptonic amplitudes

mreboud:
I had to exclude the matrix definition in eos/nonleptonic-amplitudes/nonleptonic-amplitudes.cc as the clang format made no sense.


eos/nonleptonic-amplitudes/nonleptonic-amplitudes.cc
```diff
@@ -47,6 +48,7 @@ namespace eos
         }
 
         // Note that this matrix is transposed w.r.t [HTX:2021A] to follow the convention M^i_j = M[i][j]
+        // clang-format off
```
dvandyk, line 51: That is very useful!

dvandyk (approved):

PR #990, opened 2025-02-20
Scattering phases for [HKvT:2025A]

Herren:
In this pull request we do two things:

1. Update the [GMKPRDEY:2011A] implementation of the pi-pi scattering phases to allow for the inclusion of the rho-omega interference according to [CHS:2018A]
2. Implement a modification of the [GMKPRDEY:2011A] implementation with the S-wave Omnès factor based on [DHK:2015A] and [RHK:2018A], as well as the P-wave scattering phase of [CHS:2018A]. Both of these come in numerical form and are interpolated in the implementation, thus the large files. This cocktail of scattering phases is used in our upcoming publication [HKvT:2025A].

Please let me know what you think!


eos/scattering/Makefile.am
```diff
@@ -8,9 +8,12 @@ libeosscattering_la_SOURCES = \
     ee-to-ccbar.cc ee-to-ccbar.hh \
     observables.cc observables.hh \
     parametric-gmkprdey2011.hh parametric-gmkprdey2011.cc \
+    parametric-hkvt2025.hh parametric-hkvt2025.cc \
     scattering-amplitudes.cc scattering-amplitudes.hh scattering-amplitudes-fwd.hh \
     single-channel.cc single-channel.hh \
-    single-channel-processes.hh
+    single-channel-processes.hh \
+    svalues-CHS.hh delvalues-CHS.hh \
```
dvandyk, line 15: We try to use lower-case characters in file name, except for very specific cases (``*_TEST.cc``).

eos/scattering/Makefile.am
```diff
@@ -8,9 +8,12 @@ libeosscattering_la_SOURCES = \
...
+    svalues-om-S.hh reom11-S.hh reom12-S.hh imom11-S.hh imom12-S.hh
```
dvandyk, line 16: It's a pitty that we do not yet have ``#embed`` in C++.

dvandyk (approved):
Looks good.

eos/scattering/Makefile.am
Herren: Alright! Will change that.

eos/scattering/Makefile.am
Herren: Maybe for C++-2050 or so ;)

PR #992, opened 2025-02-24
Silence warnings

Herren:
This PR silences a couple of warnings, including another one concerning a boost header that will be moved in future releases.


dvandyk (approved):
Perfect. Thank you very much @Herren!

PR #995, opened 2025-03-05
[constraints] Run @HPQCD:2023A tensor form factors down to sqrt(mb*mc) scale

gubernari:


dvandyk (commented):
Mean values look good, but covariance change are puzzling. Compare lines 420, last element of the matrix. It decreases, even though you increased the value of the form factor. I would assume that, if you multiply the mean values with ``1.0x`` that you mulitply the covariance entries with ``1.0x^2``!? This should increase the last element of the covariance matrix in line 420, shouldn't it?

gubernari:
> Mean values look good, but covariance change are puzzling. Compare lines 420, last element of the matrix. It decreases, even though you increased the value of the form factor. I would assume that, if you multiply the mean values with `1.0x` that you mulitply the covariance entries with `1.0x^2`!? This should increase the last element of the covariance matrix in line 420, shouldn't it?

I can't see your point. For instance the last element of line 420 goes from 0.001007  to 0.001068, so it increases, do you agree?

dvandyk:
> > Mean values look good, but covariance change are puzzling. Compare lines 420, last element of the matrix. It decreases, even though you increased the value of the form factor. I would assume that, if you multiply the mean values with `1.0x` that you mulitply the covariance entries with `1.0x^2`!? This should increase the last element of the covariance matrix in line 420, shouldn't it?
> 
> I can't see your point. For instance the last element of line 420 goes from 0.001007 to 0.001068, so it increases, do you agree?

You are absolutely right. I skipped a zero when reading that entry!

PR #998, opened 2025-03-12
[nonlocal-form-factors] Increase tolerance in test

MJKirk:
This test seems to fail when compiling on ws16, so slightly increase the tolerance


mreboud (approved):
Thanks!

PR #999, opened 2025-03-12
Changes to PToPP form factors

Herren:
Simple changes to make existing PToPP form factors compatible with partial-wave based parameterizations.
Nothing speaks against keeping the (q2,k2,costheta)-dependent implementation, as going from partial waves to
full form factors is trivial. The partial waves are normalized as in FFKMvD:2013A.

Furthermore, we introduce an IsospinOption, allowing to specify combinations of isospin-configurations.

### DIFF
--- a/eos/b-decays/b-to-pi-pi-l-nu_TEST.cc
+++ b/eos/b-decays/b-to-pi-pi-l-nu_TEST.cc
@@ -57,6 +57,8 @@ class BToPiPiLeptonNeutrinoTest :
                 Options oo;
                 oo.declare("model"_ok, "CKM");
                 oo.declare("form-factors"_ok, "BFvD2016");
+                oo.declare("I"_ok, "0|1");
+                oo.declare("L"_ok, "S|P|D|F");
 
                 BToPiPiLeptonNeutrino d(p, oo);
 
--- a/eos/form-factors/analytic-b-to-pi-pi.cc
+++ b/eos/form-factors/analytic-b-to-pi-pi.cc
@@ -3,6 +3,7 @@
 /*
  * Copyright (c) 2016-2025 Danny van Dyk
  * Copyright (c) 2018      Keri Vos
+ * Copyright (c) 2025      Florian Herren
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -56,6 +57,10 @@ namespace eos
         // further hadronic inputs
         PionLCDAs pi;
 
+        // isospin combination
+        PartialWaveOption opt_L;
+        double _S_switch, _P_switch, _D_switch, _F_switch;
+
         // routine to determine renormlization scale
         std::function<double (const double &)> mu;
 
@@ -67,7 +72,12 @@ namespace eos
             m_B(p["mass::B_d"], u),
             f_pi(p["decay-constant::pi"], u),
             _mu(p["B->pipi::mu@BFvD2016"], u),
-            pi(p, o)
+            pi(p, o),
+            opt_L(o, options, "L"_ok),
+            _S_switch(opt_L.value() && PartialWave::S),
+            _P_switch(opt_L.value() && PartialWave::P),
+            _D_switch(opt_L.value() && PartialWave::D),
+            _F_switch(opt_L.value() && PartialWave::F)
         {
             std::string scale = o.get("scale"_ok, "fixed");
 
@@ -83,6 +93,7 @@ namespace eos
             {
                 throw InvalidOptionValueError("scale"_ok, scale, "fixed, variable");
             }
+
 #if 0
             static const double q2 = 0.6, k2 = 18.6;
             std::cout << "q2 = " << q2 << std::endl;
@@ -463,6 +474,7 @@ namespace eos
     const std::vector<OptionSpecification>
     Implementation<AnalyticFormFactorBToPiPiBFvD2016>::options
     {
+        { "L"_ok, { "S|P|D|F" }, "S|P|D|F"}
     };
 
     AnalyticFormFactorBToPiPiBFvD2016::AnalyticFormFactorBToPiPiBFvD2016(const Parameters & p, const Options & o) :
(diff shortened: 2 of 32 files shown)


eos/form-factors/analytic-b-to-pi-pi.cc
```diff
@@ -528,6 +529,145 @@ namespace eos
...
+        {
+            return 0.5 / std::sqrt(3.0) * this->f_perp(q2, k2, x);
+        };
+
+        std::function<complex<double>(const double &)> integrandD = [&] (const double & x)
+        {
+            return 0.5 / std::sqrt(5.0) * x * this->f_perp(q2, k2, x);
+        };
+
+        unsigned npoints = 1000;
```
dvandyk, line 556: I am pretty sure that ``integrate1D`` requires a power of 2 as the number of integration points.
Also, ``integrate1D`` is deprecated. Can you allocate a suitable workspace upon construction of the observable and use e.g. ``integrate<gsl::QNG>``?

eos/utils/options.cc
```diff
@@ -386,4 +387,75 @@ namespace eos
...
+    IsospinOption::~IsospinOption() = default;
+
+    uint8_t
+    IsospinOption::value() const
+    {
+        static const std::map<std::string, uint8_t> map
+        {
+            { "0",          1     },
+            { "1",          2     },
+            { "0 & 1",      3     },
```
dvandyk, line 405: Hm, if we want to use this as part of an option's value, we need to avoid the space here. Can we use ``0&1`` instead? Or better (using the above idea of an isospin enum class) ``0|1``?

Also, we might want to specialise ``destringify<Isospin>``.

eos/utils/options.cc
```diff
@@ -386,4 +387,75 @@ namespace eos
...
+    PartialWaveOption::~PartialWaveOption() = default;
+
+    uint8_t
+    PartialWaveOption::value() const
+    {
+        static const std::map<std::string, uint8_t> map
+        {
+            { "S",          1     },
+            { "P",          2     },
+            { "S & P",      3     },
```
dvandyk, line 442: Ditto here. ``enum class PartialWave : uint8_t``, with similar operators as suggested for ``Isospin``?

eos/utils/options.cc
```diff
@@ -386,4 +387,75 @@ namespace eos
...
     }
+
+    IsospinOption::IsospinOption(const Options & options, const std::vector<OptionSpecification> & specifications, const std::string & key) :
+        SpecifiedOption(options, specifications, key)
+    {
+    }
+
+    IsospinOption::~IsospinOption() = default;
+
+    uint8_t
```
dvandyk, line 398: I think I would love to have an ``Isospin`` enum class for this. I would suggest something like the following to be put wherever we had already put ``QuarkFlavor`` and ``LeptonFlavor``:
```c++
#include <type_traits>

enum class Isospin : uint16_t
{
    empty = 0,
    zero = 1 << 0,
    onehalf = 1 << 1,
    one = 1 << 2,
    threehalf = 1 << 3,
    two = 1 << 4,
    fivehalf = 1 << 5
};

inline Isospin operator | (Isospin lhs, Isospin rhs)
{
    using T = std::underlying_type_t <Isospin>;
    return static_cast<Isospin>(static_cast<T>(lhs) | static_cast<T>(rhs));
}
    
inline Isospin & operator |= (Isospin & lhs, Isospin rhs)
{
    lhs = lhs | rhs;
    return lhs;
}

inline Isospin operator & (Isospin lhs, Isospin rhs)
{
    using T = std::underlying_type_t <Isospin>;
    return static_cast<Isospin>(static_cast<T>(lhs) & static_cast<T>(rhs));
}
```

This way, you can check if a given isospin should be included by testing
```c++
if (isospin & Isospin::two)
{
    // account for isospin I=2 contribution
    // ...
}
```

[not compile-tested!]

eos/b-decays/b-to-pi-pi-l-nu_TEST.cc
```diff
@@ -57,6 +57,8 @@ class BToPiPiLeptonNeutrinoTest :
                 Options oo;
                 oo.declare("model", "CKM");
                 oo.declare("form-factors", "BFvD2016");
+                oo.declare("I", "0 & 1");
```
dvandyk, line 60: This reminds me that ``Options::declare`` should really only accept a ``qnp::OptionKey`` and ``qnp::OptionValue`` rather than two strings.

dvandyk (commented):
Sorry @Herren that it took me so long to review your PR. Let me know what you think of my suggestions.

eos/utils/options.cc
mreboud: Cf #1001

eos/utils/options.cc
mreboud: cf #1001

eos/b-decays/b-to-pi-pi-l-nu_TEST.cc
Herren: Should we make a new issue to implement that globally?

eos/utils/destringify.hh
```diff
@@ -91,7 +92,7 @@ namespace eos
                 std::string::size_type i = 0, j = input.find('|');
                 do
                 {
-                    const auto value = input.substr(i, j);
+                    const auto value = input.substr(i, j - i);
```
dvandyk, line 95: Argh. I had planned to use ``std::views::split``. Unfortunately in C++20 (rather than C++23), using ``substr`` is less of a nuisance than the ranges library. Proven wrong...

eos/form-factors/analytic-b-to-pi-pi.cc
```diff
@@ -463,6 +471,8 @@ namespace eos
     const std::vector<OptionSpecification>
     Implementation<AnalyticFormFactorBToPiPiBFvD2016>::options
     {
+        { "I", { "0|1" }, "0|1" },
```
dvandyk, line 474: Does this work as intended? In particular, will this raise an error if the option value ``0`` is provided?

dvandyk (approved):
Conditionally approved, subject to my question on the isospin and partial wave options being answered positively.

eos/utils/destringify.hh
Herren: The `std::views` business is something I'll have to look into. No experience with it so far...

eos/form-factors/analytic-b-to-pi-pi.cc
Herren: Yes, it will. I will add one test case to make sure that is handled correctly and does not change accidentally in future.

eos/form-factors/analytic-b-to-pi-pi.cc
```diff
@@ -528,6 +538,128 @@ namespace eos
...
+        {
+            return 0.5 / std::sqrt(5.0) * x * this->f_perp(q2, k2, x);
+        };
+
+        std::function<complex<double>(const double &)> integrandF = [&] (const double & x)
+        {
+            return 0.125 / std::sqrt(7.0) * (5.0 * x * x - 1.0) * this->f_perp(q2, k2, x);
+        };
+
+        res[0] = 0.0;
```
dvandyk, line 561: Shouldn't this check against what partial waves are enabled?

dvandyk (approved):

PR #1001, opened 2025-04-06
Add new bitfieldable enums ``Isospin`` and ``PartialWave``

dvandyk:
@Herren can you have a look if these are helpful for your PR #999?


Herren:
> @Herren can you have a look if these are helpful for your PR #999?

@dvandyk Yes, I do believe this is the better way to go! I am happy as it is and would update PR #999 based on this.

Herren (approved):

PR #1008, opened 2025-05-05
Switch to using ``qnp::OptionKey`` in everything related to options

dvandyk:
- eed2114d8f3679601a1f4b4db7a9273b81dc2d77 Split the qualified name parts into a separate header file. This avoids circular dependencies between ``QualifiedName`` and ``Options``.
- bf7e196fd906de171df770ffa92a4648b40b00c6 Add a user-defined literal operator for ``qnp::OptionKey``. Can be used in form of
```c++
qnp::OptionKey key_OK  = "foo"_ok; // OK
qnp::OptionKey key_ERR = "alpha_s"_ok; // throws an exception, since '_' is not a permitted character in an option key
```
- 70f1d2cbf013f5d4ea7785b8ed96856c59def29c Remove an unused option specification in ``eos/form-factors/analytic-b-to-pi-pi.cc``.
- 9a6e42c409f831d64426d04accdcc7715b13f056 Add default comparison operator for ``qnp::OptionKey``. This allows us to use e.g. as they key type in ``std::map<>``.
- f5bd2fb105dd14d8eb9d31873ab47c33ddc05c14 Add short cuts to ``stringify`` any qualified name part.
- 9712d08a614adca21ba2341d21272798fb2ad8a8 Export ``qnp::OptionKey`` and ``qnp::OptionValue``.
- c9e65b737a25a167ea8a0b978f064f765f990e3d Mark ``eos.OptionSpecification`` as ``no_init``. While this is no longer needed after I had added implicit conversion from ``str`` to ``eos.qnpOptionKey`` in 9712d08a614adca21ba2341d21272798fb2ad8a8, I think it is useful. No Python user should ever need to create a new instance of the class ``eos.OptionSpecification``.
- ec132905f6e343cfc64504b33fc69e5b48ad1fde Remove options from three parametric constraints. These options were originally meant to record the values of certain additional parameters in the analysis that produced these constraints.
- 3f1a6c2926888462fe8a6fdf71b2992e5ee4f47a Make the switch and use ``qnp::OptionKey`` in lieu of ``std::string`` in everything related to ``Options``. I will handle the switch to ``qnp::OptionValue`` in a future PR. (Now free of the spurious ``analysis.yaml`` file!)
- ec4e5c861bb01c471d792cf1c30b746f83e3b83d Instrument more of ``eos/constraint.cc`` with contexts for debugging purpose. This was very helpful to find the culprit behind the test case failures fixed by ec132905f6e343cfc64504b33fc69e5b48ad1fde.


dvandyk:
TODO:
- [ ] remove spurious file ``analysis.yaml`` from this PR.

eos/constraint.cc
mreboud, line 1: Github made a mess with the file comparison, so this change is a bit hard to review...

mreboud (approved):
I only have minor comments!

eos/utils/qualified-name-parts.hh
```diff
@@ -79,6 +79,10 @@ namespace eos
                 OptionKey(const std::string &);
 
                 const std::string & str() const { return _key; };
+
+                inline auto operator<=> (const OptionKey & rhs) const = default;
+                inline bool operator== (const OptionKey & rhs) const = default;
+                inline bool operator< (const OptionKey & rhs) const = default;
```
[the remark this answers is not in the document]
dvandyk: I had trouble with Boost Python unless ``operator<`` is declared.

eos/utils/qualified-name-parts.hh
```diff
@@ -106,6 +108,54 @@ namespace eos
     {
         return qnp::OptionValue(std::string(str, len));
     }
+
+    namespace implementation
+    {
+        template <>
```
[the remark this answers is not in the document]
dvandyk: Through a base class possibly. But for this small number of elements (which I do not expect to enlarge at all), I think this is the least amount of code possible.

eos/constraint.cc
```diff
@@ -1652,176 +1654,176 @@ namespace eos
...
                 {
-                    std::vector<QualifiedName> observables;
                     for (auto && o : n["observables"])
                     {
                         observables.push_back(QualifiedName(o.as<std::string>()));
                     }
+                }
+                catch (QualifiedNameSyntaxError & e)
+                {
+                    throw ConstraintDeserializationError(name, "'" + observables.back().full() + "' is not a valid observable name (" + e.what() + ")");
```
mreboud, line 1667: `back` is incorrect since the observable has not been pushed to the list yet.

dvandyk: Indeed! Fixed.

PR #1018, opened 2025-05-24
Add arguments for minimal number of iterations to ``sample-nested`` tasks and CLI script

dvandyk:
- a299fba963a293aa6739413838646dfe83b3daf5 Add ``miniter`` argument to the ``sample-nested`` task. This simply pipe the argument through to ``eos.Analysis.sample_nested``.
- 2e667f6e77eb31743846b0b4d0904b14e43d3121 Add ``--min-number-iterations`` argument to the ``eos-analysis`` script. Since both ``-m`` and ``-M`` are already used as short cuts, I decided to use ``-l``.

### DIFF
--- a/python/eos/tasks.py
+++ b/python/eos/tasks.py
@@ -593,7 +593,7 @@ class DynestyResultLogger:
 
 # Nested sampling
 @task('sample-nested', '{posterior}/nested')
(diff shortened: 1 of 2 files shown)


mreboud (approved):
Thanks @dvandyk !

mreboud:
Do you understand why the tests are failing?

PR #1025, opened 2025-06-18
Fixes needed for issue 1024

### LINKED ISSUE #1024: Constraint test was commented out
In 712adb3, one of the tests in  `eos/constraint_TEST.cc` was commented out (see the very last diff)

MJKirk:
Commit f51d322 fixes #1024.
That test still fails in three cases that need further work:
- @mreboud's CMD ee->pi pi constraints (eos/constraints/ee-to-pipi.yaml) have non positive definite covariance matrices.
- For various constraints in "eos/constraints/B-to-psd-psd.yaml", `representation` must be specified as either `QCDF`, `SU3F`, or `topological`. Given these 
- In Nico's addition of `B_(s)->D_(s)^*::FormFactors[TraditionalBasis]@HPQCD:2023A` and `B_s->D_s^*::FormFactors@HPQCD:2023A`, get the error `No form factor found for process 'B_s->D_s^*' and tag 'BGL1997'`. Currently only `BSZ2015` and `BGJvD2019` are implemented

Danny already fixed missing FF options in the 0->pipi FF constraints (2ad9e57f), and I have made changes where it seems obvious / correct:
- Changing B->pipi FF observable names
- Removing constraints on Bd->ll, since this observable was removed in #930

### DIFF
--- a/eos/constraint_TEST.cc
+++ b/eos/constraint_TEST.cc
@@ -1294,9 +1294,9 @@ class ConstraintTest : public TestCase
             {
                 std::cout << "# Constraints :" << std::endl;
 
-                Options o;
-                auto constraints = Constraints();
-                unsigned n = 0;
+                Options  o;
+                auto     constraints = Constraints();
+                unsigned n           = 0;
 
                 for (auto cf = constraints.begin(); cf != constraints.end(); ++cf, ++n)
                 {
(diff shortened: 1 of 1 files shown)


mreboud:
> > Ok, my mistake. Since it was never used so far, I did not notice. It comes directly from the experimental paper. This will be fixed in a coming PR by Fatemeh.
> 
> Can she add her commit to this PR (whenever it is ready)? Since all we cannot merge without fixing all the constraint issues anyway.

I have no clue when it will be ready, maybe the simplest is to comment the problematic constraint out with a comment for the time being.

 
> > Yes, I am a bit surprised that I did not get a warning for those. Sorry about that. Can you put 'topological' as the default one?
> 
> Well, the warnings would come from the constraint test which this PR uncomments, so not so surprising... Can do re putting `topological`

> Thanks :)

dvandyk:
Thanks for taking care of this @MJKirk !

dvandyk (approved):
Thank you very much @MJKirk!

PR #1044, opened 2025-07-30
szego-polynomial_TEST fixed, check only the real part of the parameters

flavioarchilli:
szego-polynomial TEST was failing in checking the relative error of some parameters. fixed asking just to check the real part of the parameters


dvandyk:
@flavioarchilli Could you please rebase your branch on top the ``master`` branch? We do not allow merge commits / non-linear history.

mreboud (changes requested):
Can you please also test that the imaginary parts are compatible with 0?

mreboud:
Closing this PR and opening a clean one for that

PR #1051, opened 2025-08-21
Add time-dependent $B_s^0 \rightarrow \phi \mu^+\mu^-$ observables

schmitse:
This PR adds angular observables measurable in a time-dependent angular analysis of $B_s^0 \rightarrow \phi \mu^+\mu^-$ decays and updates the observables defined for the time-integrated case. 
In general the implementation follows Virto et al in arXiv 1502.05509, where the methodology of $B \rightarrow V \ell\ell$ are explored when the decay products are $CP$-eigenstates. 

The main differences concern the normalisation of the angular coefficients, the integrated decay rate in a given $q^2$ region and the definition of the observables themselves. 
In general this PR introduces six kind of angular coefficients: 
  - $K$, $W$: which are equivalent to $S$ and $A$ in the $B^0 \rightarrow K^{\ast 0}\mu^+\mu^-$ decays
  -  $\mathcal{H}$, $\mathcal{Z}$: which are added through the time-dependent $CP$-V
  - $S$, $A$: which are measured in the time-integrated case for $B_s^0 \rightarrow \phi\mu^+\mu^-$ (The name difference with respect to $K^{\ast 0}$ is unfortunate but induced by the fact that the time-integrated LHCb measured used this nomenclature) 

Key changes comprise: 
  - `BsToPhiDilepton`: 
    - Add two new public functions to retrieve the $J$ coefficients both differential and integrated. 
    - Change the prefactors for the $J_i$ to be aligned with Virto et al in arXiv 1502.05509, this results in a different prefactor of the decay rate as well.  
  - `BsToPhiDileptonAndConjugate`: 
    - Add new struct for the $s_i$ coefficients `AngularsCoefficients`
    - Add calculation of the $s_i$ following Virto et al in arXiv 1502.05509
    - Add functions to calculate the angular coefficients following Virto et al in arXiv arXiv 1502.05509
    - Add `decay_rate` function following Virto et al in arXiv 1502.05509
  - `obervables.cc`: Add definitions for all of the angular coefficients in their "theory" and "LHCb" conventions. I prepared some macros to reduce manual writing in the future but they are currently not used. 
  - `hadron-properties`: Update the phase of $q/p$ for the $B_s^0$ system to use the most recent value from HFAG. 

I will prepare a small accompanying pdf document that details the changes in the code, why i think they are necessary and adds comparisons to the implementation in flavio.


schmitse:
Hi Méril, 

sorry i did not see your first response before. 
I agree discussing tomorrow in person will be more efficient. 
The reason i was switching to use this definition is because then i can use all numerical factors from 1502.05509 for the $\phi$ implementation. 
In general i agree that having a streamlined version is desirable but i did not have the time to make sure i am not producing more typos than correct results. 
Using the parameters for $x$, $y$, $\Gamma$ makes sense. I was trying to get a working version to compare against flavio and did not think about best practices for the details.

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -91,7 +91,7 @@ namespace eos
         {
             Context ctx("When constructing Bs->Phill observables");
 
-            std::string tag = o.get("tag"_ok, "");
+            std::string tag = o.get("tag"_ok, "GvDV2020");
```
mreboud, line 94: Please keep this default value empty. We don't want to provide a default value here.

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -118,7 +118,7 @@ namespace eos
         {
         }
 
-        inline std::array<double, 12> angular_coefficients_array(const BsToPhiDilepton::Amplitudes & A, const double & s) const
+        std::array<double, 12> angular_coefficients_array(const BsToPhiDilepton::Amplitudes & A, const double & s) const
```
mreboud, line 121: Can you put the `inline` statement back here?

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -130,7 +130,7 @@ namespace eos
             double beta = std::sqrt(beta2);
 
             // j1s
-            result[0] = 3.0 / 4.0 * (
+            result[0] = (
```
mreboud, line 133: As discussed, we would like to keep the current definition for the j's observables. I can take care of rearranging the factors of 3/4 once the rest is ready, if you want.

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -241,7 +241,7 @@ namespace eos
         inline double decay_width(const BsToPhiDilepton::AngularCoefficients & a_c)
         {
             // cf. [BHvD2010], p. 6, eq. (2.7)
-            return 2.0 * a_c.j1s + a_c.j1c - 1.0 / 3.0 * (2.0 * a_c.j2s + a_c.j2c);
+            return 2.0 * a_c.j1s + a_c.j1c - 1.0 / 4.0 * (2.0 * a_c.j2s + a_c.j2c);
```
mreboud, line 244: A factor of  3/4 is missing in the first two terms

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -913,8 +924,9 @@ The azimuthal angle between the Kbar-K plane and the l^+l^- plane.";
 
 
     BsToPhiDileptonAndConjugate::BsToPhiDileptonAndConjugate(const Parameters & parameters, const Options & options):
-        bstophidilepton(parameters, options + Options{{"cp-conjugate"_ok, "true"}}),
-        bstophidilepton_conjugate(parameters, options + Options{{"cp-conjugate"_ok, "false"}})
+        bstophidilepton(parameters, options + Options{{"cp-conjugate"_ok, "false"}}),
+        bstophidilepton_conjugate(parameters, options + Options{{"cp-conjugate"_ok, "true"}}),
+        m_y(0.129), m_x(26.79), m_gamma_s(0.6578428049823233)
```
mreboud, line 929: Can you use parameters instead of class variables?

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -970,10 +981,10 @@ The azimuthal angle between the Kbar-K plane and the l^+l^- plane.";
         double beta2 = 1.0 - z;
         double beta = std::sqrt(beta2);
 
-        const complex<double> expiphi(cos(phiBs), sin(phiBs));
+        const complex<double> expiphi(cos(phiBs), -1.0 * sin(phiBs)); // e^iphi is the phase of -q/p
```
mreboud, line 984: I am not so sure, can you check that?

eos/rare-b-decays/bs-to-phi-ll.cc
```diff
@@ -1095,35 +1106,562 @@ The azimuthal angle between the Kbar-K plane and the l^+l^- plane.";
...
     }
+    BsToPhiDileptonAndConjugate::AngularhCoefficients
+    BsToPhiDileptonAndConjugate::differential_angular_h_coefficients(const double & s) const
+    {
+        return BsToPhiDileptonAndConjugate::AngularhCoefficients(differential_angular_h_coefficients_array(s));
+    }
 
     double
-    BsToPhiDileptonAndConjugate::integrated_H_1c(const double & s_min, const double & s_max) const
+    BsToPhiDileptonAndConjugate::integrated_H(const double & s_min, const double & s_max, const std::string & name) const
```
mreboud, line 1116: We need to discuss this in more details :)

mreboud (changes requested):
Thanks a lot for the PR @DominikSue. Here is a first round of comments. I think we need to discuss with @dvandyk how we want to implement some of the functions you suggested.

eos/rare-b-decays/bs-to-phi-ll.cc
schmitse: In the most recent version i have one suggestion using an enum instead of a string to curry the function instead. (See the integrated and differential S functions).

schmitse:
I should have fixed the commit messages, split the commit where the typo is fixed from the others adding new things and reverted the inline and tag change i had made locally.

schmitse:
Also i re-added the factor $3/4$ in the $J_i$ definitions and accordingly in the angular observable functions. 
For the $h_i$ and $s_i$ i don't have any different factors compared to Virto in 1502.05509.

schmitse:
Hi Méril, 
we have not had a meeting yet, i will follow up with Danny offline.

PR #1052, opened 2025-08-22
Allow the same variable to be fixed to several values in the same ExpressionObservable

### LINKED ISSUE #1048: Creating ExpressionObservables with kinematics using `=` operator not working properly
Creating new observables via expressions gives wrong results for evaluation when fixing kinematic variables with the same name via the `=`  operator in multiple observables. 

Example `test.py`:
    
    import eos 
    eos.Observables().insert('B->pi::f_+0TEST',R'test', eos.Unit.Unity(), eos.Options(), 
                         '<<B->pi::f_+(q2)>>[q2=1.0] +  <<B->pi::f_0(q2)>>[q2=2.0]')

    left = eos.Observable.make('B->pi::f_+(q2)', eos.Parameters.Defaults(), eos.Kinematics(q2=1.0), eos.Options(l='mu', **{'form-factors':    'BCL2008'}))
    right = eos.Observable.make('B->pi::f_0(q2)', eos.Parameters.Defaults(), eos.Kinematics(q2=2.0), eos.Options(l='mu', **{'form-factors': 'BCL2008'}))
    right_wrong = eos.Observable.make('B->pi::f_0(q2)', eos.Parameters.Defaults(), eos.Kinematics(q2=1.0), eos.Options(l='mu', **{'form-factors': 'BCL2008'}))
    total = eos.Observable.make('B->pi::f_+0TEST', eos.Parameters.Defaults(), eos.Kinematics(), eos.Options(l='mu', **{'form-factors': 'BCL2008'}))

    print()
    print(f'direct eval:   f_+0TEST = {total.evaluate()}')
    print(f'indirect eval: f_+(q2=1.0) + f_0(q2=2.0) = {left.evaluate() + right.evaluate()}')
    print(f'WRONG eval:    f_+(q2=1.0) + f_0(q2=1.0) = {left.evaluate() + right_wrong.evaluate()}')

Code is run in release EOS Version 1.10.16 and gives:

direct eval:   f_+0TEST = 0.6402822683329001
indirect eval: f_+(q2=1.0) + f_0(q2=2.0) = 0.6513426352930902
WRONG eval:    f_+(q2=1.0) + f_0(q2=1.0) = 0.6402822683329001

Both "direct eval" and "indirect eval" should give the same results, but do not.
The "WRONG eval" gives the same result as "direct eval", although it should not,  because it evaluates f_0  at q2=1.0 instead of the supposed q2=2.0. 

Using aliases via the `=>` operator instead of fixed kinematics does not give wrong results.

mreboud:
@dvandyk, can you please check that my fix makes sense? I don't understand why a fixed kinematic value was added to the set of aliased kinematics.

Github: Fixes #1048

### DIFF
--- a/eos/utils/expression-visitors.cc
+++ b/eos/utils/expression-visitors.cc
@@ -105,7 +105,7 @@ namespace eos::exp
     void
     ExpressionPrinter::operator() (ObservableExpression & e)
     {
-        _os << "ObservableExpression(" << e.observable->name() << ")";
+        _os << "ObservableExpression(" << e.observable->name();
         if (! e.kinematics_specification.aliases.empty())
         {
             auto a = e.kinematics_specification.aliases.cbegin();
--- a/eos/utils/expression-visitors.cc
+++ b/eos/utils/expression-visitors.cc
@@ -546,7 +546,6 @@ namespace eos::exp
             for (const auto & value : kinematics_values)
             {
                 kinematic_set.erase(value.first);
-                alias_set.insert(value.first);
             }
             for (const auto & alias : kinematics_aliases)
             {
--- a/eos/utils/expression-parser_TEST.cc
+++ b/eos/utils/expression-parser_TEST.cc
@@ -414,5 +414,33 @@ class ExpressionParserTest : public TestCase
                     TEST_CHECK_NEARLY_EQUAL(std::visit(evaluator, e), 1.0, 1e-10);
                 }
             }
+
+            // testing that a kinematic variable can be fixed twice via aliasing
+            {
+                // test::obs1 is a test observable that requires two kinematic specifications, q2_min and q2_max
+                // it returns p[mass::c] * multiplier * (q2_max - q2_min)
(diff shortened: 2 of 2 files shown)


eos/utils/expression-visitors.cc
```diff
@@ -246,6 +246,14 @@ namespace eos::exp
     {
     }
 
+    std::string
+    ExpressionCloner::next_hidden_alias()
+    {
+        this->_hidden_alias_index++;
+
+        return stringify(_hidden_alias_index);
```
dvandyk, line 254: This produces a number as the name of the hidden kinematic variable. Does it make sense to name the variables, e.g., ``internal_variable_{idx}``?

dvandyk (approved):
Approved, however, see my nitpicky question regarding the variable names.

mreboud:
@DominikSue, your wishes are orders: I added the test. Can you have a look? If Danny and you agree, this can be merged I think, the KinematicReader behaves as expected.

PR #1059, opened 2025-08-26
Issue1017

cbolognani:
Ensured the tasks `sample_mcmc`, `sample_prior`,  `mixture_product`, `sample_pmc`, `sample_nested` include correct log levels. For sampling tasks, they indicate a `eos.inprogress` to state the beginning of the computations, a `eos.completed` to indicate the end and a `eos.info` whenever information can be provided.


python/eos/tasks.py
```diff
@@ -239,7 +239,7 @@ def find_mode(analysis_file:str, posterior:str, base_directory:str='./', optimiz
             bfp = _bfp
             min_chi2 = _chi2
 
-    eos.completed(f'... optimization finished')
+    eos.ted(f'... optimization finished')
```
dvandyk: What happened here? :confused:  And why is this problem not found by tests?

dvandyk (changes requested):
Thank you for this PR @cbolognani !

Please see my comments. There is one unintended change in the first commit that were undone in the 2nd.

PR #1063, opened 2025-08-28
Improve the documentation of units and fill the missing ones in the parameters

### LINKED ISSUE #1054: Better checks on units
Our tests should pick up things like a756363c7c758e673115a4e7869bbf25f92454e7

Also there's a reasonable number of parameters which don't have units defined (see all the "undefined" in the [parameter page](https://eoshep.org/doc/reference/parameters.html)) - I would suggest we also make this an error

mreboud:
Github: closes #1054

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -19,6 +19,7 @@ reference/constraints.rst
 reference/observables.rst
 reference/parameters.rst
 reference/python.rst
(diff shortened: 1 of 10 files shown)


dvandyk (changes requested):
Merci @mreboud! Some nitpicky comments/questions and a typo.

PR #1069, opened 2025-08-29
Support writing observables in Python and plotting observables in the complex plane

dvandyk:
- d381d758490f353f99805eecde3867a0fddaa918 Fix a memory leek by adding the virtual destructor to the ``Observable`` base class.
- d2372ec2c4f6e4af8fbe5b44ff857efc28c401b2 Create a new class ``ExternalObservable`` (and companion ``ExternalObservableEntry``) that are used internally by the new Python function ``eos.register_python_observable``.
- d826a2828b9d8bfbb5dd091cabf38ca26b253ee2 Add a new item type to ``eos.figure`` to support plotting observables in the complex plan.

### DIFF
--- a/python/eos/figure/item.py
+++ b/python/eos/figure/item.py
@@ -1433,6 +1433,111 @@ class SignalPDFItem(Item):
 
         ax.plot(xvalues, _np.exp(pvalues), alpha=self.alpha, color=self.color, label=self.label, ls=self.linestyle)
 
+@dataclass(kw_only=True)
+class ComplexPlaneItem(Item):
+    """Plots a single observable as a function on the complex plan
+
+    :param fixed_kinematics: A dictionary of additional kinematic variables and the values to which they shall be fixed.
+    :type fixed_kinematics: dict[str, float] | None
+    :param fixed_parameters: A dictionary of EOS parameters and the values to which they shall be fixed.
+    :type fixed_parameters: dict[eos.QualifiedName, float] | None
+    :param fixed_parameters_from_file: Path to a file containing EOS parameters to be used for the observable.
+    :type fixed_parameters_from_file: str | None
+    :param observable: The name of the observable to be plotted.
+    :type observable: :class:`eos.QualifiedName`
+    :param options: A dictionary of options to be passed to the observable.
+    :type options: dict[str, str] | None
+    :param ranges: A tuple of two float values (min, max) representing the range of the kinematic variable to be plotted on the x- and y-axes.
+    :type ranges: tuple[tuple[float, float], tuple[float, float]]
+    :param resolution: The number of points to use for the plot.
+    :type resolution: int
+    :param variables: The names of the variables to be used as x and y coordinates.
+    :type variables: tuple[str, str]
+    """
+    fixed_kinematics:dict[str,float]|None=None
+    fixed_parameters:dict[eos.QualifiedName,float]|None=None
+    fixed_parameters_from_file:str|None=None
+    observable:eos.QualifiedName
+    options:dict[str, str]|None=None
+    variables:tuple[str,str]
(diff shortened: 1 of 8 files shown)


python/eos/figure/item.py
```diff
@@ -1433,6 +1433,101 @@ def draw(self, ax):
...
         ax.plot(xvalues, _np.exp(pvalues), alpha=self.alpha, color=self.color, label=self.label, ls=self.linestyle)
 
+@dataclass(kw_only=True)
+class ComplexPlaneItem(Item):
+    """Plots a single observable as a function on the complex plan
+
+    :param fixed_kinematics: A dictionary of additional kinematic variables and the values to which they shall be fixed.
+    :type fixed_kinematics: dict[str, float] | None
+    :param fixed_parameters: A dictionary of EOS parameters and the values to which they shall be fixed.
+    :type fixed_parameters: dict[eos.QualifiedName, float] | None
```
mreboud, line 1443: Why not provide an eos.Parameter entry directly?

python/eos/figure/item.py
```diff
@@ -1433,6 +1433,101 @@ def draw(self, ax):
...
+    :param fixed_parameters_from_file: Path to a file containing EOS parameters to be used for the observable.
+    :type fixed_parameters_from_file: str | None
+    :param observable: The name of the observable to be plotted.
+    :type observable: :class:`eos.QualifiedName`
+    :param options: A dictionary of options to be passed to the observable.
+    :type options: dict[str, str] | None
+    :param range: A tuple of two float values (min, max) representing the range of the kinematic variable to be plotted on the x-axis.
```
mreboud, line 1450: Shouldn't it be the x and the y axes?

python/eos/figure/item.py
```diff
@@ -1433,6 +1433,101 @@ def draw(self, ax):
...
+            if self.fixed_parameters_from_file:
+                eos.warn('Overriding values read from \'parameters-from-file\' with explicit values in \'parameters\'')
+
+            for key, value in self.fixed_parameters.items():
+                self._parameters.set(key, value)
+
+        # Declare variable that is either parameter or kinematic
+        self._variable = None
+
+        # Do the variables correspond to one of the kinematic variables?
```
mreboud, line 1494: It might be a good idea to make an external routine ourt of this, since it is used in many plotting functions

python/eos/figure/item.py
```diff
@@ -1433,6 +1433,101 @@ def draw(self, ax):
...
+
+        # Create options
+        self._options = eos.Options()
+        if self.options:
+            for key, value in self.options.items():
+                self._options.declare(key, value)
+
+        self._observable = eos.Observable.make(self.observable, self._parameters, self._kinematics, self._options)
+
+        self._xvalues = _np.linspace(-2.0, +2.0, self.resolution)
```
mreboud, line 1513: This should be the `range` provided by the user, no?

mreboud (commented):
Thank you, @dvandyk! The first commit looks good, the second will require an entry in the documentation, the third one is a good starting point, but we might have to add a couple of things (e.g. to plot in the z(q^2) plane).

python/eos/figure/item.py
mreboud: Ah yes good point. Can I easily convert an eos.Parameter to a dict like this?

python/eos/figure/item.py
dvandyk: Fixed the range. The shape can be nicely done as part of the transformation inside the observable.

dvandyk:
@mreboud ready for a check-up if you like!

mreboud (approved):
Nice, thanks @dvandyk!

PR #1070, opened 2025-08-29
Change references to new bibliography format and add missing references to bibliography

### LINKED ISSUE #1068: Small changes to references in comments
Many of the older code comments are written like ` cf. [BEKU2002], Eq. (3.6)` but in our bibliography this is listed as BEKU:2002A.
Adding the colon to do a `ctrl-f` when I want to look up the source paper wastes many milliseconds of my time, so I propose to update the comments to match the current bibliography style.

MJKirk:
Closes #1068

### DIFF
--- a/eos/b-decays/b-to-d-l-x-nu.cc
+++ b/eos/b-decays/b-to-d-l-x-nu.cc
@@ -201,7 +201,7 @@ namespace eos
     BToDLeptonInclusiveNeutrinos::description = "\
 The neutrino-inclusive decay B->D l X_nu, where l=e,mu is a light lepton, and \
 X_nu = {nu, nu nubar nu} is either a one or three neutrino final state. The \
-numerical implementations for this process follow [BIvD2016].";
+numerical implementations for this process follow [BIvD:2016].";
 
     const std::string
     BToDLeptonInclusiveNeutrinos::kinematics_description_s = "\
--- a/eos/b-decays/b-to-d-pi-l-nu.cc
+++ b/eos/b-decays/b-to-d-pi-l-nu.cc
@@ -97,7 +97,7 @@ namespace eos
             result -= lambda / (m_B + m_Dstar) * ff->a_2(q2);
             result /= (2.0 * m_Dstar * sqrt(q2));
 
-            // cf. [DDS2014], eq. (22), p. 17
+            // cf. [DDS:2014], eq. (22), p. 17
             return result;
         }
 
@@ -106,7 +106,7 @@ namespace eos
             const double m_B     = this->m_B();
             const double m_Dstar = this->m_Dstar();
 
-            // cf. [DDS2014], eq. (22), p. 17
+            // cf. [DDS:2014], eq. (22), p. 17
             return sqrt(2.0) * (m_B + m_Dstar) * ff->a_1(q2);
         }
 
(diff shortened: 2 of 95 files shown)


MJKirk:
Currently missing references:
AFGB:2017A
AGV:2019A
BBJW:1018A
BCvDV:2017A
BL:1998A
BMRvD:2021A
BMRvD:2022A
BRY:2006A
DKKMO:2008A
DSD:2014A
DV:2015A
FJV:2016A
GHMS:2012A
GMKPREYD:2011A
GRvDV:2022A
GRvDV:2022B
GvDV:2022B
HKVT:2025A
HTX:2021A
KKvDZ:2022A

MJKirk:
Ambiguous references (original comment or code has no letter suffix, while the current reference file has multiple references in that year)
BaBar:2007
Belle:2008
LHCb:2021
S:2004

MJKirk:
Unrelated, but here is a good a place as any: the KLOE:2006A paper title is wrong

dvandyk:
> The final error is from various `@FNAL+MILC:2013A` constraints. After some archaeology and help from Danny I established that while the underlying results are indeed based on 1310.3722, the actual numbers were produced by Danny and Frederik Beaujean. For the B->K*, the numbers in EOS can be found (approximately) in 1310.2478 Table VIII, while the Bs->K* numbers are (exactly) found in 1503.09063, Table V.
> 
> Do we rename the constraint to match the paper they can actually be found in, or leave as FNAL+MILC (but I would definitely add a comment saying the numbers can be seen in the corresponding paper by Danny)?

Thanks @MJKirk!

1.) It is all based on 1310.3722, Horgan et al. and we received updated results from a subsequent proceeding I think. I don't want to suggest that this is something either I or Fred did.
2.) I think this is actually an HPQCD paper using MILC ensembles, so the name should reflect this.

I suggest ``HPQCD:2013X`` as the new reference name, where ``X`` is the first free letter.

MJKirk:
> I suggest `HPQCD:2013X` as the new reference name, where `X` is the first free letter.

I actually just noticed that 1310.3722 is in references.yaml as HLMW:2013A, but it doesn't appear to be used anywhere. So I will rename the constraints to HPQCD:2013B and remove HLMW:2013A

eos/constraints/B-to-D-pi-l-nu.yaml
```diff
@@ -621,7 +621,7 @@ B^0->(Dpi)^+l^-nu::KinematicalDistribution[w]@Belle:2018A:
 # 2021
 #
 # {{{
-B^0->(Dpi)^+l^-nubar::KinematicDistributionsAngularObservables@BBGJvD:2021A:
+B^0->(Dpi)^+l^-nubar::KinematicDistributionsAngularObservables@BBGJvD:2020A:
```
dvandyk: I think the paper is from 2021. Can you change the reference name instead of the constraint?

MJKirk: Done

eos/references.yaml
```diff
@@ -1446,20 +1651,57 @@ GR:2004A:
   authors: Goldberger, J. and Roweis, S.
   url: http://eprints.pascal-network.org/archive/00001565/
   title: Hierarchical clustering of a mixture model
+GRvDV:2021A:
```
dvandyk: Why is this ``GRvDV:2021A`` here? Should all occurences of that that reference be replaced by the correct one, which is ``GRvDV:2022A``?

PR #1079, opened 2025-09-21
Changes applied after review of EOS-2024-04

dvandyk:
- 6393d1979c5b7a04e3a18fb08f7004dc754b74b6 When evolving the Wilson coefficients from the high scale to the low scale in the ``StandardModel`` class, the RGE objects are declared as ``static``. However, this leaves us open to a race condition: when the coefficients are computed for the first time concurrently in two or more separate threads, the static initialization can occur simultaneously for the same object. This commit fixes the problem by declaring the RGE as ``thread_local static``.
- 2ca2ebbc0e5eead7ff2990c4a3ef20b473e846f2 Add two new parameters to represent unknown power corrections in class-I non-leptonic B decays.
- 3a732245927429a21fc8142feec915966f4477ae Use the two new parameters in B->DP and B->D^*P decay observables.

### DIFF
--- a/eos/b-decays/bq-to-dq-psd.cc
+++ b/eos/b-decays/bq-to-dq-psd.cc
@@ -75,6 +75,8 @@ namespace eos
 
         UsedParameter mu;
 
+        UsedParameter DeltaP;
+
         RestrictedOption opt_accuracy;
         double switch_lo;
         double switch_nlo;
@@ -98,6 +100,7 @@ namespace eos
             f_P(p["decay-constant::" + stringify(opt_q.value() == QuarkFlavor::down ? "K_u" : "pi")], u),
             opt_cp_conjugate(o, options, "cp-conjugate"_ok),
             mu(p[stringify(opt_q.value() == QuarkFlavor::down ? "s" : "d") + "bcu::mu"], u),
+            DeltaP(p["B->DP::DeltaP"], u),
             opt_accuracy(o, options, "accuracy"_ok)
         {
             Context ctx("When constructing B_q->D_q P observable");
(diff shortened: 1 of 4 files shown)


eos/b-decays/bq-to-dq-psd.cc
```diff
@@ -75,7 +75,9 @@ namespace eos
 
         UsedParameter mu;
 
-        RestrictedOption opt_accuracy;
+        UsedParameter DeltaP;
+
+        SpecifiedOption opt_accuracy;
```
mreboud, line 80: Why was this converted back to a SpecifiedOption? I think a RestrictedOption made more sense?

eos/b-decays/bq-to-dstarq-psd.cc
```diff
@@ -76,7 +76,9 @@ namespace eos
 
         UsedParameter mu;
 
-        RestrictedOption opt_accuracy;
+        UsedParameter DeltaV;
+
+        SpecifiedOption opt_accuracy;
```
mreboud, line 81: Same as above

eos/nonlocal-form-factors/charm-loops-analytic_TEST.cc
```diff
@@ -221,8 +221,8 @@ class TwoLoopTest :
                     TEST_CHECK_NEARLY_EQUAL(f29c_ref[i].real(), f29c_res.real(), 150*eps);
                     TEST_CHECK_NEARLY_EQUAL(f29c_ref[i].imag(), f29c_res.imag(), 150*eps);
 
-                    TEST_CHECK_NEARLY_EQUAL(f27d_ref[i].real(), f27d_res.real(), 250*eps);
-                    TEST_CHECK_NEARLY_EQUAL(f27d_ref[i].imag(), f27d_res.imag(), 200*eps);
+                    TEST_CHECK_NEARLY_EQUAL(f27d_ref[i].real(), f27d_res.real(), 2000*eps);
+                    TEST_CHECK_NEARLY_EQUAL(f27d_ref[i].imag(), f27d_res.imag(), 2000*eps);
```
mreboud, line 225: I am 2000% happy with this, since we already enlarged the uncertainties. I think I will have to investigate a bit. Does @smeiser know on which architecture the test failed?

mreboud (changes requested):
Mostly ok :)

mreboud (approved):

PR #1081, opened 2025-09-23
Remove dependency on (empty) ``libboost_system``

### LINKED ISSUE #1078: configure Fails with Boost 1.89.0 Due to Removed libboost_system
The configure script fails with Boost `1.89.0` on macOS (ARM64, Homebrew) because it checks for `libboost_system`, which was removed in Boost `1.89.0` (now header-only; it used to be an empty library provided for backward compatibility until `1.89.0`). 
Downgrading to Boost `1.85.0` works, but `configure.ac` should skip the `libboost_system` check for modern Boost versions. 
Running the configure script with boost >= 1.89.0 leads to: 
library 'boost_system' not found. 
On my system. 

Please update the build system to support Boost >=1.89.0. 
Alternatively an explicit mention of using 
`brew install boost@1.85 boost-python; brew link --force boost@1.85` 
and subsequently pinning the packages would be appreciated to avoid other people having to squabble with this issue.

dvandyk:
As of Boost version 1.89, this library is not even available anymore.

GitHub: resolves #1078

@schmitse Can you please confirm if this fixes your issue?

### DIFF
--- a/doc/installation.rst.in
+++ b/doc/installation.rst.in
@@ -147,7 +147,7 @@ pkg-config
   the freedesktop.org library helper, in version 0.27.1 or higher;
 
 BOOST
-  the BOOST C++ libraries (sub-libraries ``boost-filesystem``, ``boost-system`` and ``boost-math``);
+  the BOOST C++ libraries (sub-libraries ``boost-filesystem`` and ``boost-math``);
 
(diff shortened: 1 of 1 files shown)


mreboud (approved):
Looks good! Thanks!

PR #1085, opened 2025-09-28
Change expected directory hierarchy for tasks

dvandyk:
Prior to this commit, tasks expect samples and other data produced by tasks to reside within ``EOS_BASE_DIRECTORY/POSTERIOR``. As of this commit, they will reside within ``EOS_BASE_DIRECTORY/data/POSTERIOR``, to mirror what is done for other outputs, such as reports and figures.

@mreboud We had discussed this at one point. Thought I'd PR it.

### DIFF
--- a/examples/analysis-organisation.ipynb
+++ b/examples/analysis-organisation.ipynb
@@ -319,7 +319,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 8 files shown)


mreboud (approved):
Looks also good and needed, thanks!

MJKirk (approved):

Issue #1080, opened 2025-09-23
[form-factors] Remove the old LCSR namespace


mreboud:
Currently, the LCSR based on B-meson LCDAs rely on process-dependent structures `analytic-[decay]-lcsr.cc`. We should update these structures (using `QuarkFlavor`, ...) and probably merge them with the ones in `mesonic-processes.hh`.

PR #1096, opened 2025-10-11
Convert to B-meson LCSR code to use form factor traits

dvandyk:
- 0b6c8f0489471fc89300499bdbbde4b4335bea06 Deals with B to pseudoscalar form factor code.
- 0fe07e31e81009d5a1283d234dc5e165fbc4de59 Deals with B to vector form factor code.

### DIFF
--- a/eos/form-factors/analytic-b-to-dstar-lcsr.cc
+++ b/eos/form-factors/analytic-b-to-dstar-lcsr.cc
@@ -1,8 +1,8 @@
 /* vim: set sw=4 sts=4 et foldmethod=syntax : */
 
 /*
- * Copyright (c) 2018 Danny van Dyk
- * Copyright (c) 2018 Ahmet Kokulu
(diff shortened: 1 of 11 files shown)


dvandyk:
> This is related to #1080 I guess? But the point was that the code is duplicated between `mesonic-processes.hh` and the individual `XX_lcsr.cc` files. Do we really want to keep the LCSR files separated? They really contain the same information as in `mesonic-processes.hh`: name of the particles, masses, resonances...

@mreboud Sorry for the delay. The structures do not carry the same information, unfortunately. I tried simply using the structs living in ``mesonic-processes.hh``, and the test cases failed.
For example, the LCSRs predict ``B->pi`` for ``b->u`` transitions. The ``mesonic-processes.hh`` assumes ``B->pi`` is a ``b->d`` transition.

mreboud (approved):
I don't like it, but who am I to complain?

PR #1097, opened 2025-10-11
Rework logging callback to fix race condition

dvandyk:
Matthew Kirk uncovered an issue in the logging callback code related to multi- threaded computations. This issue materialised only after switching to Python version 3.12 on our development machine.

The usual setup is a main thread (started by a Python interpreter), with many worker threads being spawned by the thread pool (typically on account of the ObservableCache). When a worked thread logs a message and Log subsequently forwards this log message via logging_callback to the Python side, we run into a race condition: the main thread still holds the Global Interpreter Lock (GIL), which prohibits us from creating new Python-side objects from within the logging_callback function. We cannot acquire the GIL either, since the main thread is not likely to give it up. Luckily, Python has a solution for this conundrum. Using Py_AddPending call, we schedule a function call to be scheduled whenever convenient to the main thread.

### DIFF
--- a/python/_eos/log.cc
+++ b/python/_eos/log.cc
@@ -1,7 +1,7 @@
 /* vim: set sw=4 sts=4 et foldmethod=marker : */
 
 /*
- * Copyright (c) 2023 Danny van Dyk
+ * Copyright (c) 2023-2025 Danny van Dyk
  *
  * This file is part of the EOS project. EOS is free software;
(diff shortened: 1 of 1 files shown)


MJKirk (approved):
Worked for me

MJKirk:
@mreboud - did this also fix the segfault you found?

PR #1104, opened 2025-10-27
Start building wheels for python 3.14

### LINKED ISSUE #1094: Build for python 3.14
3.14 was released yesterday, so a new line needs adding to .github/workflows/pypi-build+check+deploy.yaml at some stage

MJKirk:
Closes #1094 

Before merging this, we should make a new release of pypmc, with my new ABI forward compatible wheels, so that for python 3.14 (and 3.13) people don't have to build that wheel from source when installing EOS


MJKirk:
Ah, the docker images need updating too

PR #1105, opened 2025-10-30
uslnu extras

MJKirk:
- In c9decbe19a36d48a717ac7ca5afa393839eb6f3d I just remove a comment that I accidentally left in in PR #1082 
- In d4286c75acd6b419f7890272475c388620d24c01 I add $f_+(0)$, $\lambda_{+,0}^{(\prime)}$, (the slope in the linear / quadratic expansion approximation), and $\lambda_{+,0}^{''}$ (the quadratic term)

Leaving as draft in case anything else comes up that needs to be merged for this paper

### DIFF
--- a/python/eos/figure/plot.py
+++ b/python/eos/figure/plot.py
@@ -132,12 +132,15 @@ class XAxis(Deserializable):
     :type ticks: :class:`XTicks <eos.figure.XTicks>`
     :param unit: The unit of the x axis, e.g., 'GeV'. Defaults to None.
     :type unit: str | None
+    :param scale: The scale of the x axis, e.g., 'linear' or 'log'. Defaults to 'linear'.
+    :type scale: str
     """
 
     label:str=field(default=None)
     range:tuple[float, float]=field(default=None)
     ticks:XTicks=field(default_factory=XTicks)
     unit:str=None
+    scale:str='linear'
 
     def __post_init__(self):
         if self.range is not None and len(self.range) != 2:
@@ -155,6 +158,7 @@ class XAxis(Deserializable):
         if self.range is not None:
             ax.set_xlim(self.range)
         self.ticks.draw(ax)
+        ax.set_xscale(self.scale)
 
     @classmethod
     def from_dict(cls, **kwargs):
@@ -211,12 +215,15 @@ class YAxis(Deserializable):
     :type ticks: :class:`YTicks <eos.figure.YTicks>`
     :param unit: The unit of the y axis, e.g., 'GeV'. Defaults to None.
     :type unit: str | None
+    :param scale: The scale of the y axis, e.g., 'linear' or 'log'. Defaults to 'linear'.
+    :type scale: str
     """
 
     label:str=None
     range:tuple[float, float]=None
     ticks:YTicks=field(default_factory=YTicks)
     unit:str=None
+    scale:str='linear'
 
     def __post_init__(self):
         if self.range is not None and len(self.range) != 2:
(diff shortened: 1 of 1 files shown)


MJKirk:
3a1fd6122b898f0f5c3d774fbb4ef4bc123a0fb2 will close #1108

eos/form-factors/observables.cc
```diff
@@ -2886,6 +2886,15 @@ namespace eos
                 make_form_factor_adapter("0->Kpi::Im{f_+}(Re{q2},Im{q2})", R"(\textrm{Im}(f_+^{0\to K\pi}(q^2)))",
                         &FormFactors<VacuumToPP>::im_f_p, std::make_tuple("Re{q2}", "Im{q2}")),
 
+                make_observable("0->Kpi::f_+(0)", R"(f_+^{0 \to K\pi}(0))", Unit::None(),
+                        &KSvD2025FormFactors<VacuumToKPi>::fp_at_0),
+
+                make_observable("0->Kpi::lambda_+", R"(\lambda_+^{('), K \to \pi})", Unit::None(),
```
dvandyk: Inconsistent. The label contains a prime, and the next pseudoobservable is ``doubleprime``, but this has no ``prime`` in its name.

dvandyk: Same goes for the name of the providing method.

eos/form-factors/parametric-ksvd2025.cc
```diff
@@ -455,6 +456,46 @@ namespace eos
         return integrate<GSL::QAGS>(f, -M_PI, M_PI) / (2.0 * M_PI);
     }
 
+    double KSvD2025FormFactors<VacuumToKPi>::fp_at_0() const
+    {
+        return std::real(this->f_p(0.0));
+    }
+
+    double KSvD2025FormFactors<VacuumToKPi>::lambda_plus() const
```
dvandyk: Rename to ``lambda_prime_plus``.

python/eos/figure/data.py
```diff
@@ -95,7 +95,7 @@ def labels(self, variables):
             return [p[dist].latex() for dist in variables]
         elif self._type == 'prediction':
             o = eos.Observables()
-            return ["$" + o[dist].latex() + "$" for dist in variables]
+            return ["$" + o[dist.split(";")[0]].latex() + "$" for dist in variables]
```
dvandyk: Hm. Is this a qualified name? If so, we should probably use
```py
dist = eos.QualifiedName(dist)
...
return ["$" + o[str(dist)].latex() + "$" for dist in variables]
```

dvandyk (changes requested):
Thanks @MJKirk, looks mostly good.

Few comments:
- I think it makes sense to use ``lambda_prime_plus`` rather than ``lambda_plus``.
- Not so sure about ``lambda_prime_zero`` anymore, if we do not have the quadratic term.
- Please use ``eos.QualifiedName`` where possible.

Otherwise good to merge!

python/eos/figure/data.py
dvandyk: I see. That's a bug in whatever produces that particular file. It should read:
```
CKM::Vus[...];form-factors=KSvD2025,model=CKM,n-resonances-0p=2,n-resonances-1m=2
```
i.e., the kinematics should be added to the name part.

MJKirk: Yes let's do this for now, since even with the square bracket in the right place `str(eos.QualifiedName(dist))` contains the square brackets and so it still doesn't work

PR #1107, opened 2025-11-03
Add missing include for std::find_if

mesmith75:
This was need for my compilation: `gcc15`


mreboud:
@mesmith75 can you rename your commit "[eos] Add missing include for std::find_if" to follow our naming scheme?

mesmith75:
Completely mangled the git history. Start again.

PR #1140, opened 2026-03-10
Export Wilson polynomial coefficients to python

mreboud:
This allows the user to access the coefficients of the Wilson polynomials directly, e.g., to copy them into a paper.


dvandyk (changes requested):
Looks all good, except for the first commit. That one's dangerous.

dvandyk (approved):

PR #1142, opened 2026-03-17
WET Wilson coefficients for c->ull transitions

DominikSue:
I implement the WET Wilson coefficients for c -> u l l transitions by making the following changes: 

- Add basis for WCs with Delta c = Delta u = 1 transitions
- Add method `evolve_intermediate` to `MultiplicativeRenormalizationGroupEquation`  template that returns for NLL the same result as evolve as a series in alpha_s
- Add SM matching and RGEs at NLL with matching performed at mu_W and mu_b
- Add checks comparing results with own mathematica code
- Add general NNLL template for `MultiplicativeRenormalizationGroupEquation` with an `evolve_intermediate`  method
- Update UC WCs to NNLL including checks again

### DIFF
--- a/eos/models/model.hh
+++ b/eos/models/model.hh
@@ -4,6 +4,7 @@
  * Copyright (c) 2010-2025 Danny van Dyk
  * Copyright (c) 2018 Ahmet Kokulu
  * Copyright (c) 2018 Christoph Bobeth
+ * Copyright (c) 2026 Dominik Suelmann
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -66,6 +67,9 @@ namespace eos
 
             // Hadronic sectors (Delta B = 2)
             struct SBSB;
+
+            // Neutral-current semileptonic sectors (Delta C = 1)
+            struct UC;
         } // namespace WET
 
         // Old-style WET sectors
@@ -202,6 +206,16 @@ namespace eos
             virtual WilsonCoefficients<wc::SBSB> wet_sbsb() const = 0;
     };
 
+    /*!
+     * Base class for the Delta C = 1 = -Delta U FCNC component of models.
+     */
+    template <> class ModelComponent<components::WET::UC>
+    {
+        public:
+            /* c->u Wilson coefficients */
+            virtual WilsonCoefficients<wc::UC> wilson_coefficients_uc(const LeptonFlavor & lepton_flavor, const bool & cp_conjugate = false) const = 0;
+    };
+
     /* Old-style WET sectors */
 
     /*!
@@ -236,6 +250,8 @@ namespace eos
         public virtual ModelComponent<components::WET::SBCU>,
         // Hadronic sectors (Delta B = 2)
         public virtual ModelComponent<components::WET::SBSB>,
+        // Neutral-current semileptonic sectors (Delta C = 1)
+        public virtual ModelComponent<components::WET::UC>,
         // Old-style WET sectors
         public virtual ModelComponent<components::DeltaBS1>
     {
--- a/eos/models/wilson-coefficients.cc
+++ b/eos/models/wilson-coefficients.cc
@@ -3,6 +3,7 @@
 /*
  * Copyright (c) 2010, 2011 Danny van Dyk
  * Copyright (c) 2010 Christoph Bobeth
+ * Copyright (c) 2026 Dominik Suelmann
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -402,5 +403,13 @@ namespace eos
     WilsonCoefficients<wc::SBNuNu>::WilsonCoefficients()
     {
         _coefficients.fill(0.0);
+    }
+
+    WilsonCoefficients<wc::UC>::WilsonCoefficients(const double & alpha_s_mu) :
+        _alpha_s(alpha_s_mu)
+    {
+        _sm_like_coefficients.fill(0.0);
+        _primed_coefficients.fill(0.0);
+        _scalar_tensor_coefficients.fill(0.0);
     }
 } // namespace eos
--- a/eos/models/wilson-coefficients.hh
+++ b/eos/models/wilson-coefficients.hh
@@ -4,6 +4,7 @@
  * Copyright (c) 2010-2023 Danny van Dyk
  * Copyright (c) 2014 Frederik Beaujean
  * Copyright (c) 2014 Christoph Bobeth
+ * Copyright (c) 2026 Dominik Suelmann
  *
  * This file is part of the EOS project. EOS is free software;
  * you can redistribute it and/or modify it under the terms of the GNU General
@@ -421,6 +422,9 @@ namespace eos
 
         struct SBCU
         {};
+
+        struct UC
+        {};
     } // namespace wc
 
     /* Wilson coefficients for |Delta B| = |Delta S| = 2 operators */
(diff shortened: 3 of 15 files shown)


eos/parameters/sm-and-eft.yaml
```diff
@@ -468,6 +490,395 @@ groups:
...
+      'uc::mu' :
+          central: +1.5
+          min:     +1.5
+          max:     +1.5
+          unit:     'GeV'
+          latex:    '$\mu^{c \to u}$'
+          comments: ''
+
+      # Wilson coefficients C1 - C6 at mu = 1.5 GeV to NNLL accuary as calculated in the SM by EOS's StandardModel class.
+      # For the calculations, cf. [xxx].
```
dvandyk: I think we should cite [this paper](https://inspirehep.net/literature/1471106) for it, which would be abbreviated as ``dBMS:2016A`` and which would need an entry in ``eos/references.yaml``. The entry would need to be done via an additional, earlier commit.

eos/parameters/sm-and-eft.yaml
```diff
@@ -468,6 +490,395 @@ groups:
...
+      "c->u::c1" :
+          central: -0.29063621
+          min:     -0.29063621
+          max:     -0.29063621
+          unit:     '1'
+          latex:    $\mathcal{C}_1$
+          comment:  'cf. @REF[BMU:1999A]@, eqs. (1) and (2)'
```
dvandyk: I think I would refer here to table 1 of the deBoer/Müller/Seidel paper. Also, at NNLL the values seem a bit off.

eos/parameters/sm-and-eft.yaml
```diff
@@ -468,6 +490,395 @@ groups:
...
+          comment:  'cf. @REF[BHvD:2012A]@, eqs. (2) and (3)'
+      "c->umumu::Im{c9}" :
+          central: +0.0
+          min:     -15.0
+          max:     +15.0
+          unit:     '1'
+          latex:    $\mathrm{Im}\, \mathcal{C}_9^{(\mu)}$
+          comment:  'cf. @REF[BHvD:2012A]@, eqs. (2) and (3)'
+      "c->umumu::Re{c10}" :
+          central: -4.16611761
```
dvandyk: This looks like a copy and paste of the ``b->smumu`` coefficients.

eos/parameters/sm-and-eft.yaml
```diff
@@ -468,6 +490,395 @@ groups:
...
+          central: +1.5
+          min:     +1.5
+          max:     +1.5
+          unit:     'GeV'
+          latex:    '$\mu^{c \to u}$'
+          comments: ''
+
+      # Wilson coefficients C1 - C6 at mu = 1.5 GeV to NNLL accuary as calculated in the SM by EOS's StandardModel class.
+      # For the calculations, cf. [xxx].
+      "c->u::c1" :
```
dvandyk: The qualified name prefix should be ``uc``, not ``c->u``. The latter style is currently only used for ``b->s{gamma,ll}`` and needs to be retired.

eos/references.yaml
```diff
@@ -1254,13 +1254,25 @@ D0:2011A:
   inspire-id: Abazov:2011ry
   title: Measurement of the CP-violating phase $\phi_s^{J/\psi \phi}$ using the flavor-tagged
     decay $B_s^0 \rightarrow J/\psi \phi$ in 8 fb$^{-1}$ of $p \bar p$ collisions
+dB:2017A:
```
dvandyk, line 1257: This should be two separate commits, which come first in the ordering.

eos/models/wilson-coefficients.hh
```diff
@@ -421,6 +421,9 @@ namespace eos
 
         struct SBCU
         {};
+
+        struct DeltaCU1
+        {};
```
dvandyk, line 427: This should be ``UC``

eos/models/model.hh
```diff
@@ -202,6 +205,16 @@ namespace eos
...
     };
 
+    /*!
+     * Base class for the Delta C = 1 = -Delta U FCNC component of models.
+     */
+    template <> class ModelComponent<components::WET::DeltaCU1>
+    {
+        public:
+            /* c->u Wilson coefficients */
+            virtual WilsonCoefficients<wc::DeltaCU1> wilson_coefficients_c_to_u(const double & mu, const LeptonFlavor & lepton_flavor, const bool & cp_conjugate = false) const = 0;
```
dvandyk: I would prefer ``wilson_coefficients_uc``.

eos/utils/rge-impl.hh
```diff
@@ -242,6 +243,111 @@ namespace eos
 
         return result;
     }
+
+    template <unsigned nf_, unsigned dim_>
+    std::tuple<std::array<double, dim_>, std::array<double, dim_>, std::array<double, dim_>>
```
dvandyk: I think this should return only a tuple of 2 vectors (for LL and NLL), and not NNLL. The latter should only show up for the ``NNLL`` specialization of the class template. Similarly, the ``LL`` specialization should only return an array at LL accuracy.

eos/utils/rge_TEST.cc
```diff
@@ -230,5 +230,48 @@ class MultiplicativeRGENLLTest : public TestCase
                 TEST_CHECK_NEARLY_EQUAL(result[0], +0.229589, eps);
                 TEST_CHECK_NEARLY_EQUAL(result[1], +1.826340, eps);
             }
+
```
dvandyk, line 234: I think this commit should be folded into the commit introducing the new method.

dvandyk (changes requested):
Thank you @DominikSue for a solid first draft of the PR! I have left some comments inline. Happy to discuss my comments further via Zoom or similarly!

eos/parameters/sm-and-eft.yaml
DominikSue: yes this is indeed a copy of b->smumu, I forgot to adapt that first commit after doing the rest. Will cite dBMS:2016A here.

eos/references.yaml
DominikSue: I am not sure what you mean. Which should be separate commits? The two refs? One of the refs is needed now for the first commit, so it should be first in the ordering.

eos/utils/rge-impl.hh
DominikSue: I return this mostly to have a test which compares with the `evolve` method. I agree it makes sense to do it as you say. Maybe there is another way for a test.

eos/references.yaml
dvandyk: Yes, sorry for being ambiguous here. The two refs should be in two separate commits, and both commits should be early on in the ordering of commits in this PR.

dvandyk: Resolved, thanks!

eos/parameters/sm-and-eft.yaml
DominikSue: I added the correct values from the implementation with some small but expected deviations.
Except for C9. Here I am a bit confused, the values agree with both [https://arxiv.org/pdf/1510.00311](url) Tab. 1 and de Boers thesis [https://eldorado.tu-dortmund.de/server/api/core/bitstreams/09d1e6c5-535b-4586-9143-02baea45536a/content](url) Tab. 2.2 (both using a different basis e.g. alpha_s/4pi factor), but deviations are a bit large with [https://arxiv.org/pdf/1606.05521](url) Tab. 1

eos/models/wilson-coefficients.cc
```diff
@@ -403,4 +403,12 @@ namespace eos
     {
         _coefficients.fill(0.0);
     }
+
+    WilsonCoefficients<wc::UC>::WilsonCoefficients() :
+        _alpha_s(1.3)
```
dvandyk: Is this meant to be $\alpha_s(\mu)$ for $\mu = 1.3$ GeV?

eos/models/model.hh
```diff
@@ -66,6 +66,9 @@ namespace eos
 
             // Hadronic sectors (Delta B = 2)
             struct SBSB;
+
+            // (Delta C = 1)
```
dvandyk: Just to match the other comments, I would add here ``Neutral current sectors (Delta C = 1)``.

eos/models/wilson-coefficients.cc
DominikSue: yes, but it shouldn't it be used I think. Other methods will change the default setting. better set it to zero so we know something is wrong?

dvandyk: I think I would make it a mandatory argument of the constructor. No classes except ``Model`` and its descendants are supposed to create a new instance of ``WilsonCoefficients<...>`` anyway. So that would be fine.

eos/utils/rge.hh
```diff
@@ -140,6 +140,22 @@ namespace eos
              */
             std::array<double, dim_> evolve(const double & alpha_s_mu, const double & alpha_s_0, const std::array<double, dim_> & c_0_0,
                                             const std::array<double, dim_> & c_0_1) const;
+
+            /*!
+             * Specialized version of evolve(...) with identical arguments.
+             * Returns the evolved Wilson coefficients using evolve() expanded as a series in powers of alpha(mu) / (4 pi), to
```
dvandyk: What about:
```
Returns the evolved Wilson coefficients using evolve() expanded as a series in powers of alpha(mu) / (4 pi), for further processes. This is needed, for example when running the c->ull Wilson coefficients from the electroweak matching scale down to the charm-quark scale, with intermediate matching at the bottom-quark scale.
```
?

DominikSue: good point

eos/parameters/sm-and-eft.yaml
DominikSue: There was a bug with the alpha_s factor for C7,C8,C9 that is now fixed. Result are now in good agreement with 1606.05521 with only small and expected deviations.

eos/references.yaml
```diff
@@ -1254,6 +1254,11 @@ D0:2011A:
...
+  authors: de Boer, Stefan
+  inspire-id: deBoer:2017dgn
+  url: https://doi.org/10.17877/DE290R-18060
```
dvandyk, line 1260: Sorry, didn't notice this before.

@DominikSue Please keep it as is.

@mreboud For future entries, how should we handle publications (e.g., a PhD thesis like here) that have a DOI but no arXiv preprint?

eos/parameters/sm-and-eft.yaml
```diff
@@ -468,6 +483,395 @@ groups:
...
+          central: +1.3
+          min:     +1.3
+          max:     +1.3
+          unit:     'GeV'
+          latex:    '$\mu^{uc}$'
+          comments: ''
+
+      # Wilson coefficients C1 - C6 at mu = 1.3 GeV to NLL accuracy as calculated in the SM by EOS's StandardModel class.
+      # For the calculations, cf. [dBMS:2016A].
+      "uc::c1" :
```
dvandyk: Can you expand this into two parameters please, ``uc::Re{c1}`` and ``uc::Im{c1}``, and so on also for c2 through c6 + c8? I know we do not have the presently for the ``b->sll`` coefficients, but as I mentioned earlier: there are a lot of things that need changing / retiring...

DominikSue: I expanded them into two parameters, let me know if you find anything else that should be updated compared to `b->sll`

dvandyk (approved):
Excellent, I approve. Thank you @DominikSue!

PR #1146, opened 2026-04-21
Fixes and improvements for the BGL parameterization

mreboud:
This PR:
- Implement the saturations of the dispersive bound within BGL and remove the old and wrong bounds
- Fixes some convention mismatches to align with BSZ
- Fix the optimised t0 in the default parameters
- Implement some observables related to b -> c decays


eos/form-factors/parametric-bgl1997-impl.hh
```diff
@@ -479,6 +533,46 @@ namespace eos
...
+        coefficients[0] = a_0_0();
+        std::copy(_a_f_0.begin(), _a_f_0.end(), coefficients.begin() + 1);
+
+        return std::inner_product(coefficients.begin(), coefficients.end(), coefficients.begin(), 0.0);
+    }
+
+    template<typename Process_>
+    double BGL1997FormFactors<Process_, PToP>::saturation_0m_a() const
+    {
+        return 0.;
```
dvandyk: Coding style says: ``0.0``.

eos/form-factors/observables.cc
```diff
@@ -455,6 +455,24 @@ namespace eos
                         Unit::None(),
                         &HQETFormFactors<BToD, PToP>::h_T,
                         std::make_tuple("q2")),
+
+                make_observable("B->D::Saturation[0^+_V]", R"(\textrm{Saturation}[0^+_V])", Unit::None(),
```
dvandyk, line 459: I think this observable should have a suffix indicating which form factor parametrisation was used. This will make it easier to merge @gubernari's ongoing branch in the future. I suggest adding ``@BGL`` here.

eos/form-factors/unitarity-bounds.cc
```diff
@@ -3105,189 +3105,4 @@ namespace eos
...
-    BGLUnitarityBounds::begin_options()
-    {
-        return Implementation<BGLUnitarityBounds>::options.cbegin();
-    }
-
-    std::vector<OptionSpecification>::const_iterator
-    BGLUnitarityBounds::end_options()
-    {
-        return Implementation<BGLUnitarityBounds>::options.cend();
-    }
```
dvandyk, line 3292: Good riddance, thank you @mreboud!

eos/form-factors/observables.cc
```diff
@@ -2665,140 +2665,44 @@ namespace eos
                         Unit::None(),
```
dvandyk, line 2665: Commit message is not in the imperative mood. (``Removes`` -> ``Remove``)

dvandyk (changes requested):
Thanks @mreboud!

Requesting minor changes only: coding style, one commit message, one slight change to the name of the saturations within the BGL parametrisation.

eos/b-decays/b-to-psd-l-nu.cc
Herren, line 1: This file baffles me a bit: why is the argument here `s`? Shouldn't it be `q2`?
That would also be in better agreement with the header.

Herren (commented):
In addition to Danny's comments, I only have one more thing :)

eos/b-decays/b-to-psd-l-nu.cc
mreboud: Cf. #1148

eos/form-factors/observables.cc
mreboud: I fully agree with the comment, but we will need to do it more generally. BFW's saturations also lack a suffix. I suggest we merge it as is, since it is not ambiguous and I will take care of suffixing them later.

dvandyk: Fair enough!

dvandyk (approved):

PR #1149, opened 2026-04-27
Small changes to the figure framework

### LINKED ISSUE #1147: Plotting a single observable with `UncertaintyBandItem` in the figure framework
It is very convenient to predict multiple observables simultaneously (e.g., form factors). However, the `Figure` framework doesn't support plotting them individually. I'd like to implement that in a clean way; we really want to avoid the mess the `Plotter` framework currently is.

As Danny noted, indexing the observables in the figure is probably not the best idea, since modifying the analysis file might break backwards compatibility.

The cleanest approach would probably be to name the observable by its QualifiedName, together with options (e.g. for a differential BR to two different lepton flavours). So the main difficulty is to write this properly. What do you think @MJKirk @dvandyk ?

mreboud:
I would like these changes to appear in the next release, but I agree that all of this needs to be tested (it works with my analysis notebooks).

This PR:
- Gets rid of the `eos.plot.Plotter._weighted_quantiles`
- Adds the EOS watermark to all the subplots of the GridFigure
- Implements an `observable` option to specify which option needs to be plotted in an uncertainty band item

### DIFF
--- a/python/eos/figure/item.py
+++ b/python/eos/figure/item.py
@@ -263,6 +263,8 @@ class UncertaintyBandItem(Item):
     :type resolution: int
     :param variable: The name of the kinematic variable that is plotted on the x-axis. Defaults to the first kinematic variable recorded in the data file.
     :type variable: str | None
+    :param observable: The name of the observable that is plotted. This name can include options. Defaults to the first observable recorded in the data file.
+    :type observable: str | None
 
     Example:
 
@@ -271,7 +273,7 @@ class UncertaintyBandItem(Item):
         figure_args = '''
         plot:
           xaxis: { label: r'$q^2$', unit: r'$\\textnormal{GeV}^2$', range: [0.0, 11.63] }
-          yaxis: { label: r'$d\\mathcal{B}/dq^2$',                 range: [0.0,  5e-3] }
+          yaxis: { label: r'$d\\mathcal{B}/dq^2$',                  range: [0.0,  5e-3] }
           legend: { position: 'upper center' }
           items:
             - { type: 'uncertainty', label: r'$\\ell=\\mu$',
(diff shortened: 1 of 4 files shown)


python/eos/figure/item.py
```diff
@@ -376,23 +380,46 @@ def prepare(self, context:AnalysisFileContext=None):
...
+        self._observable = self.observable
+        if self._observable is None:
+            # assume that only one observable was predicted
+            observables = {obs.split(';')[0] for obs in list(self._datafile.lookup_table.keys())}
+            if len(observables) > 1:
+                raise ValueError(f"Data file '{self.datafile}' contains more than one predicted observable; specify 'observable' to determine which is supposed to be used")
+            if len(observables) == 0:
+                raise ValueError(f"Data file '{self.datafile}' contains no observable, check predictions")
+            self._observable = observables.pop()
+        self._observable = eos.QualifiedName(self._observable)
```
dvandyk, line 400: Does this always work? Assume ``self._observable`` is ``None``. Do we now pull in a prediction ID, which is not a valid ``QualifiedName``, since it can contain the kinematic specification?

dvandyk (changes requested):
Looks mostly good to me, except for the one comment and the missing words in the commit messages. I assume you used quotation marks ``"`` instead of apostrophes ``'`` when running ``git commit -m ``?

python/eos/figure/item.py
mreboud: From the above, `self._observable` cannot be `None`, right?
If it is not provided by the user (or provided as `None`), it will be assigned to the predicted observable. If there are multiple predicted observables, an error is raised.  If there are none, an error is raised.

dvandyk: After discussing with @mreboud, my concerns have been overcome. Line 394 ensures that line 400 will not error out when no ``observable`` option is provided.

dvandyk (approved):

PR #1152, opened 2026-05-03
[python] Fix linewidth and linestyle args of some item types

DominikSue:
`ObservableItem` previously does not use the `linewidth` and `linestyle` values from its base class for plotting.
So if you plot it using `ItemFactory`, the `linestyle`/`linewidth` argument is not used.
Added this functionality to `ObservableItem` and some others, where it is missing as well I think.


mreboud (approved):
Thanks @DominikSue!

PR #1154, opened 2026-05-12
Lc to pll

DominikSue:
I implement observables of Lambda_c -> p l l decays, including branching ratio, FL, AFB and CP-asymmetries
following the notation in 2107.13010 and my own code.
So far I only added the long-distance contribution of vector resonances modeled via a Breit-Wigner
and short-distance Wilson coefficients. Not yet the two-loop virtual corrections driven by C1,C2 by Stefan De Boer available in 1707.00988.

@dvandyk have a look and let me know what you think
Thanks in advance.

### DIFF
--- a/eos/observable.cc
+++ b/eos/observable.cc
@@ -30,6 +30,7 @@
 #include <eos/observable-impl.hh>
 #include <eos/observable.hh>
 #include <eos/rare-b-decays/observables.hh>
+#include <eos/rare-c-decays/observables.hh>
 #include <eos/s-decays/observables.hh>
 #include <eos/scattering/observables.hh>
 #include <eos/tau-decays/observables.hh>
@@ -57,16 +58,11 @@ namespace eos
     ObservableEntries::ObservableEntries() :
         _entries(&impl::observable_entries)
     {
-        std::vector<std::function<ObservableSection()>> section_makers = { make_form_factors_section,
-                                                                           make_nonlocal_form_factors_section,
-                                                                           make_nonleptonic_amplitudes_section,
-                                                                           make_b_decays_section,
-                                                                           make_c_decays_section,
-                                                                           make_rare_b_decays_section,
-                                                                           make_meson_mixing_section,
-                                                                           make_scattering_section,
-                                                                           make_s_decays_section,
-                                                                           make_tau_decays_section };
+        std::vector<std::function<ObservableSection()>> section_makers = {
+            make_form_factors_section,  make_nonlocal_form_factors_section, make_nonleptonic_amplitudes_section, make_b_decays_section,   make_c_decays_section,
+            make_rare_b_decays_section, make_rare_c_decays_section,         make_meson_mixing_section,           make_scattering_section, make_s_decays_section,
+            make_tau_decays_section
+        };
 
         for (const auto & section_maker : section_makers)
         {
@@ -133,6 +129,7 @@ namespace eos
                 _sections = std::vector<ObservableSection>({ make_b_decays_section(),
                                                              make_c_decays_section(),
                                                              make_rare_b_decays_section(),
+                                                             make_rare_c_decays_section(),
                                                              make_meson_mixing_section(),
                                                              make_nonleptonic_amplitudes_section(),
                                                              make_nonlocal_form_factors_section(),
--- a/eos/parameters/hadron-properties.yaml
+++ b/eos/parameters/hadron-properties.yaml
@@ -669,6 +669,27 @@ groups:
           unit:     's'
           latex:    '$\tau_{K_L}$'
           comments: 'cf. [PDG:2024A] p. 41'
+      'life_time::rho^0' :
+          central: +4.4655e-24
+          min:     +4.4412e-24
+          max:     +4.4897e-24
+          latex:   '$\tau_{\rho^0}$'
+          unit:     's'
+          comments: 'Extracted from width = 147.4+-0.8 MeV, cf. [PDG-Live], May 2026'
+      'life_time::omega' :
+          central: +7.5831e-23
+          min:     +7.4695e-23
+          max:     +7.6967e-23
+          latex:   '$\tau_{\omega}$'
+          unit:     's'
+          comments: 'Extracted from width = 8.68+-0.13 MeV, cf. [PDG-Live], May 2026'
+      'life_time::phi' :
+          central: +1.5491e-22
+          min:     +1.5444e-22
+          max:     +1.5538e-22
+          latex:   '$\tau_{\phi}$'
+          unit:     's'
+          comments: 'Extracted from width = 4.249+-0.013 MeV, cf. [PDG-Live], May 2026'
       'life_time::D_d' :
           central: +1.033e-12
           min:     +1.028e-12
--- a/eos/parameters/hadronic-matrix-elements.yaml
+++ b/eos/parameters/hadronic-matrix-elements.yaml
@@ -5274,6 +5274,50 @@ groups:
             - ["V", "A", "T", "T5"]
             - [0, 1, 2, 3, 4]
           comments: ''
+      # data-driven parameters for long-distance contributions to Lambda_c -> proton l l decays according to
+      # [GHM:2021A]
+      'Lambda_c->proton::res_a_rho@GHM2021':
+        central: +0.44
+        min: +0.0
+        max: +1.0
+        latex: '$a_{\rho}^{\Lambda_c\to p \ell\ell}$'
+        unit: 'GeV^2'
+        comments: ''
+      'Lambda_c->proton::res_a_omega@GHM2021':
+        central: +0.073
+        min: +0.0
+        max: +1.0
+        latex: '$a_{\omega}^{\Lambda_c\to p \ell\ell}$'
+        unit: 'GeV^2'
+        comments: ''
+      'Lambda_c->proton::res_a_phi@GHM2021':
+        central: +0.102
+        min: +0.0
+        max: +1.0
+        latex: '$a_{\phi}^{\Lambda_c\to p \ell\ell}$'
+        unit: 'GeV^2'
+        comments: ''
+      'Lambda_c->proton::res_delta_rho@GHM2021':
+        central: +0.0
+        min: -3.14
+        max: +3.14
+        latex: '$\delta_{\rho}^{\Lambda_c\to p \ell\ell}$'
+        unit: '1'
+        comments: ''
+      'Lambda_c->proton::res_delta_omega_m_rho@GHM2021':
+        central: +3.14
+        min: -3.14
+        max: +3.14
+        latex: '$\delta_{\omega}^{\Lambda_c\to p \ell\ell} - \delta_{\rho}^{\Lambda_c\to p \ell\ell}$'
+        unit: '1'
+        comments: ''
+      'Lambda_c->proton::res_delta_phi_m_rho@GHM2021':
+        central: +1.11
+        min: -3.14
+        max: +3.14
+        latex: '$\delta_{\phi}^{\Lambda_c\to p \ell\ell} - \delta_{\rho}^{\Lambda_c\to p \ell\ell}$'
+        unit: '1'
+        comments: ''
       # Form factor parameters for Lambda_b -> Lambda_c decay according to
       # [DKMR:2017A], use table 2 and supplemental material. For nominal fit all a_2 parameters are set to zero.
       # params for tensor FFs are included.
--- a/eos/rare-c-decays/Makefile.am
+++ b/eos/rare-c-decays/Makefile.am
@@ -1,0 +1,44 @@
+CLEANFILES = *~
+MAINTAINERCLEANFILES = Makefile.in
+
+AM_CXXFLAGS = @AM_CXXFLAGS@
+AM_LDFLAGS = @AM_LDFLAGS@
+
+lib_LTLIBRARIES = libeosrarecdecays.la
+libeosrarecdecays_la_SOURCES = \
+	lambda-c-to-proton-l-l.cc lambda-c-to-proton-l-l.hh \
+	observables.cc observables.hh
+libeosrarecdecays_la_CXXFLAGS = $(AM_CXXFLAGS) $(GSL_CXXFLAGS)
+libeosrarecdecays_la_LDFLAGS = $(AM_LDFLAGS) $(GSL_LDFLAGS)
+libeosrarecdecays_la_LIBADD = \
+	$(top_builddir)/eos/utils/libeosutils.la \
+	$(top_builddir)/eos/maths/libeosmaths.la \
+	$(top_builddir)/eos/models/libeosmodels.la \
+	$(top_builddir)/eos/form-factors/libeosformfactors.la \
+	-lboost_filesystem \
+	-lgsl -lgslcblas -lm
+
+include_eos_rarecdecaysdir = $(includedir)/eos/rare-c-decays
+include_eos_rarecdecays_HEADERS = \
+	lambda-c-to-proton-l-l.hh \
+	observables.hh
+
+EXTRA_DIST =
+
+AM_TESTS_ENVIRONMENT = \
+	export EOS_TESTS_PARAMETERS="$(top_srcdir)/eos/parameters";
+
+TESTS = \
+	lambda-c-to-proton-l-l_TEST
+
+LDADD = \
+	$(top_builddir)/test/libeostest.la \
+	libeosrarecdecays.la \
+	$(top_builddir)/eos/form-factors/libeosformfactors.la \
+	$(top_builddir)/eos/utils/libeosutils.la \
+	$(top_builddir)/eos/maths/libeosmaths.la \
+	$(top_builddir)/eos/models/libeosmodels.la \
+	$(top_builddir)/eos/libeos.la
+
+check_PROGRAMS = $(TESTS)
+lambda_c_to_proton_l_l_TEST_SOURCES = lambda-c-to-proton-l-l_TEST.cc
--- a/eos/rare-c-decays/lambda-c-to-proton-l-l.cc
+++ b/eos/rare-c-decays/lambda-c-to-proton-l-l.cc
@@ -1,0 +1,499 @@
+/* vim: set sw=4 sts=4 et foldmethod=syntax : */
+
+/*
+ * Copyright (c) 2019 Ahmet Kokulu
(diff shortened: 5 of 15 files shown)


eos/parameters/hadron-properties.yaml
```diff
@@ -155,20 +155,41 @@ groups:
           latex:   '$m_{\rho^0}$'
           unit:     'GeV'
           comments: 'cf. [PDG-Live], December 2020'
+      'gamma::rho^0' :
```
mreboud: I think that, until now, we have only implemented life_times rather than widths. I am fine with widths, but I would call it width.

eos/observable_TEST.cc
```diff
@@ -186,7 +186,13 @@ class ObservableTest : public TestCase
             {
                 static const std::set<std::string> allowed_kinematic_variables{ "q2",
                                                                                 "q2_min",
+                                                                                "q2_min_1",
```
mreboud: Do we really need that? The purpose of defining the allowed kinematic variables was precisely to avoid their proliferation.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
```diff
@@ -0,0 +1,502 @@
...
+                {
+                    return (2.0 * k1ss() - k1cc());
+                }
+
+                inline double
+                d2gamma(const double & c_lep) const
+                {
+                    const double c2_lep = c_lep * c_lep;
+                    const double s2_lep = 1.0 - c2_lep;
+                    const double s_lep  = std::sqrt(s2_lep);
```
mreboud: This is not used, is it?

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
```diff
@@ -0,0 +1,502 @@
...
+                using std::imag;
+                using std::norm;
+                using std::real;
+                using std::sqrt;
+
+                lambda_c_to_proton_l_l::Amplitudes result;
+
+                const auto wc = model->wilson_coefficients_uc(opt_l.value(), opt_cp_conjugate.value());
+
+                complex<double> wc_norm = model->ckm_us() * conj(model->ckm_cs()) + model->ckm_ud() * conj(model->ckm_cd());
```
mreboud: This name is a bit misleading. I would call it ckm_factor or something similar.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
```diff
@@ -0,0 +1,502 @@
...
+                };
+
+                const complex<double> c7minus  = (wc.c7() - wc.c7prime()) * wc_norm;
+                const complex<double> c7plus   = (wc.c7() + wc.c7prime()) * wc_norm;
+                const complex<double> c9minus  = (wc.c9() - wc.c9prime()) * wc_norm;
+                const complex<double> c9plus   = (wc.c9() + wc.c9prime()) * wc_norm;
+                const complex<double> c10minus = (wc.c10() - wc.c10prime()) * wc_norm;
+                const complex<double> c10plus  = (wc.c10() + wc.c10prime()) * wc_norm;
+
+                const double res_a_rho   = a_rho;
```
mreboud: You can also access them simply with `a_rho()` as it seems that you are only using them once below.

eos/rare-c-decays/observables.cc
```diff
@@ -0,0 +1,158 @@
...
+                                    Unit::InverseGeV2(),
+                                    &LambdaCToProtonLeptonLepton::differential_branching_ratio,
+                                    std::make_tuple("q2"),
+                                    Options{}),
+
+                    make_expression_observable("Lambda_c->protonll::BR1/BR2",
+                                               R"(\frac{\mathcal{B}_{[q2min1,q2max1]}}{\mathcal{B}_{[q2min2,q2max2]}}(\Lambda_c\to p \mu^+\mu^-))",
+                                               Unit::None(),
+                                               R"(
+                        <<Lambda_c->protonll::BR>>[q2_min=>q2_1_min,q2_max=>q2_1_max]
```
mreboud: Ah even stranger, you used a different naming scheme here ^^

mreboud: Why do you need to export these ratios by the way? This is rather something that will be done in the analysis files, no?

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
```diff
@@ -0,0 +1,158 @@
...
+#include <eos/observable.hh>
+#include <eos/rare-c-decays/lambda-c-to-proton-l-l.hh>
+#include <eos/utils/wilson-polynomial.hh>
+
+#include <test/test.hh>
+
+#include <array>
+#include <cmath>
+#include <fstream>
+#include <iostream>
```
mreboud: These two can probably go

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
```diff
@@ -0,0 +1,158 @@
...
+                    {        "model"_ok,       "WET" },
+                    { "form-factors"_ok, "BMRvD2022" },
+                    {            "l"_ok,        "mu" }
+                };
+
+                LambdaCToProtonLeptonLepton d(p, oo);
+
+                double eps = 5e-4;
+
+                // TEST_CHECK_RELATIVE_ERROR(d.differential_branching_ratio(0.75), 2.75291755e-7, eps);
```
mreboud: Why is this commented out?

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
```diff
@@ -0,0 +1,158 @@
...
+                    {        "model"_ok,       "WET" },
+                    { "form-factors"_ok, "BMRvD2022" },
+                    {            "l"_ok,        "mu" },
+                    { "cp-conjugate"_ok,      "true" }
+                };
+
+                LambdaCToProtonLeptonLepton dNP(p, oo);
+                LambdaCToProtonLeptonLepton dNPbar(p, oobar);
+
+                eps = 2e-3;
```
mreboud: Any reason why you had to use this value? This looks a bit too large ^^

mreboud (changes requested):

eos/parameters/hadron-properties.yaml
DominikSue: It follows PDGs usage of widths and life times. Might be easier to update it in the future this way. I will change it to widths

mreboud: Let's wait for @dvandyk's feedback :)

dvandyk: I would implement it as a life time with an entry in the ``comments`` field that it was taken from the PDG and converted from width to lifetime, stating the original value provided for the width.

eos/observable_TEST.cc
DominikSue: For Lc->p l l we have branching ratio measurements on combined bins and ratios of combined bins with no measurements available for individual bins by LHCb. So we need three `q2_min,` `q2_max` variables to implement them. `q2_min_num` and `q2_min_denom` is already implemented. I can remove `q2_min_3`
and use  `q2_min_denom` , but one more set of  `q2_min,` `q2_max` is needed.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
DominikSue: Not yet. I forgot to add it as an observable, it adds value I think.

eos/rare-c-decays/observables.cc
DominikSue: 1) I changed the naming scheme so I can use the figure methods with these kinds of observables. I forgot to change it in the other commit. Do you think it might be useful to change `q2_min_num` to `q2_num_min`
also for others at some point?

2) Do you think I shouldn't add the branching ratio measurements to EOS? (See also other comment) Should these be added better to analysis files?

eos/observable_TEST.cc
mreboud: Not sure I understand, can you send me the reference to the LHCb measurements?

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
DominikSue: For charm you have this strong CKM suppression and for the NP basis of WCs we normally didn't include a CKM factor because of this. Currently, in analog to b->sll, for the SM WCs a CKM factor gets added not in the basis but in observables. So basically NP WCs in EOS include a factor 1/ckm_factor for b->sll and now c->ull. So this large value times the CKM factor is roughly order 1. Do you have a suggestion how to do this better? I think it might not be a good implementation, but I wasn't sure how to do it better.

DominikSue:
Thanks for all the comments!

eos/observable_TEST.cc
DominikSue: https://arxiv.org/pdf/2407.11474 It is about the rho region and the upper limit on the branching ratio in the combined low-m and high-m region.

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
mreboud: I was not complaining about the WC values, but rather about the value of epsilon you used in the tests. You are testing the implementation at the 0.2% level while providing 18-digit values! If your tests are precise enough, you can reduce epsilon to 10^-5 or lower.

DominikSue: Ah ok, I thought 0.2% is enough. I am not sure if the tests are precise enough to reduce it much further. I will check.

eos/parameters/hadronic-matrix-elements.yaml
```diff
@@ -5274,6 +5274,50 @@ groups:
             - ["V", "A", "T", "T5"]
             - [0, 1, 2, 3, 4]
           comments: ''
+      # data-driven parameters for long-distance contributions to Lambda_c -> proton l l decays according to
+      # [GHM:2021A]
+      'Lambda_c->proton::res_a_rho':
```
dvandyk: Since this is specific to ``[GHM:2021A]``, please append ``@GHM2021A`` to all parameter names here.

eos/observable_TEST.cc
dvandyk: As discussed via Zoom just now, I would recommend keeping the relevant observables (and their associated kinematic variable names) to a custom/personal analysis file rather than adding them to the main EOS repository.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
```diff
@@ -0,0 +1,502 @@
...
+        struct Amplitudes
+        {
+                double aU11;
+                double aL11;
+                double aU22;
+                double aL22;
+                double aP12;
+                double aS22;
+
+                double alpha;
```
dvandyk: This is unused here, isn't it?

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
```diff
@@ -0,0 +1,502 @@
...
+
+            const complex<double>
+            normalization(const double & q2) const
+            {
+                double lam = lambda(m_Lambda_c * m_Lambda_c, m_proton * m_proton, q2);
+
+                if ((lam <= 0) || (q2 <= 4.0 * m_l * m_l))
+                {
+                    return 0.0;
+                }
```
dvandyk, line 241: I would leave a comment here, referring to DeBoer/Müller/Seidel 1606.05521, stating that there is no *overall* normalization in terms of CKM factors that applies to all Wilson coefficients. Instead, the individual factors are applied in the amplitude calculations.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
dvandyk: I agree with the sentiment. Maybe ``ckm_factor_ds``, as suggested by @DominikSue on Zoom just now.

eos/rare-c-decays/observables.cc
dvandyk: I think option 2 applies here.

eos/rare-c-decays/observables.cc
```diff
@@ -0,0 +1,158 @@
...
+                                               Unit::None(),
+                                               R"(
+                        (<<Lambda_c->protonll::BR>>[q2_min=>q2_1_min,q2_max=>q2_1_max]
+                        +
+                        <<Lambda_c->protonll::BR>>[q2_min=>q2_2_min,q2_max=>q2_2_max])
+                        /
+                        <<Lambda_c->protonll::BR>>[q2_min=>q2_3_min,q2_max=>q2_3_max]
+                        )"),
+
+                    make_observable("Lambda_c->protonll::FL(q2)",
```
dvandyk: I think we use ``F_L`` elsewhere.

eos/rare-c-decays/observables.cc
```diff
@@ -0,0 +1,158 @@
...
+                                    R"(F_{\mathrm{L}}^{\Lambda_c^+ \to p \ell^+ \ell^-}(q^2))",
+                                    Unit::None(),
+                                    &LambdaCToProtonLeptonLepton::differential_f_l,
+                                    std::make_tuple("q2"),
+                                    Options{}),
+
+                    make_observable("Lambda_c->protonll::AFB(q2)",
```
dvandyk: Similarly, I think we use ``A_FB`` elsewhere.

dvandyk (changes requested):
Thank you @DominikSue, looks very neat already. Please see my detailed change requests. I don't see any blockers!

eos/parameters/hadron-properties.yaml
dvandyk: This is resolved.

eos/observable.cc
```diff
@@ -57,16 +58,11 @@ namespace eos
...
-                                                                           make_nonlocal_form_factors_section,
-                                                                           make_nonleptonic_amplitudes_section,
-                                                                           make_b_decays_section,
-                                                                           make_c_decays_section,
-                                                                           make_rare_b_decays_section,
-                                                                           make_meson_mixing_section,
-                                                                           make_scattering_section,
-                                                                           make_s_decays_section,
-                                                                           make_tau_decays_section };
+        std::vector<std::function<ObservableSection()>> section_makers = {
```
dvandyk, line 61: Uhh. What happened here?

eos/observable_TEST.cc
dvandyk: Resolved.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
dvandyk: Resolved.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
dvandyk: Resolved.

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
dvandyk: This still requires resolution.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
dvandyk: Resolved.

eos/rare-c-decays/lambda-c-to-proton-l-l.cc
dvandyk: Resolved.

dvandyk (commented):
Most things have been resolved, except for:
- @mreboud's comment on the large tolerance of 0.2% used in (some of) the test cases;
- some snafu happening in the formatting of changes to ``eos/observables.cc``.

Fixing those, together with a rebase on top of the main branch, should render this PR mergeable. Once more, thanks @DominikSue!

eos/observable.cc
DominikSue: I checked that the clang pre-commit check wants me to have it like this. Is this not intended behavior?

eos/rare-c-decays/lambda-c-to-proton-l-l_TEST.cc
DominikSue: Increased to eps=1e-6 precision with updated checks.

DominikSue:
@dvandyk the PR should be ready now. 

1. I fixed the last typos and resolved the problem with the precision of the tests and added some more.
2. The changes to `eos/observable.cc` stem from using clang-format.
3. Copilot found a typo in `makefile.am` that might be present in other files

eos/observable.cc
dvandyk: Not fully, but if it's stable with respect to ``clang-format`` we will accept it.

eos/parameters/hadron-properties.yaml
```diff
@@ -669,6 +669,27 @@ groups:
...
+          latex:   '$\tau_{\omega}$'
+          unit:     's'
+          comments: 'Extracted from width = 8.68+-0.13 MeV, cf. [PDG-Live], May 2026'
+      'life_time::phi' :
+          central: +1.5491e-22
+          min:     +1.5444e-22
+          max:     +1.5538e-22
+          latex:   '$\tau_{\phi}$'
+          unit:     's'
+          comments: 'Extracted from width = 4.249+-0.013 MeV, cf. [PDG-Live], May 2026'
```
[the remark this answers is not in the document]
dvandyk: Dominik's version is entirely sufficient.

eos/rare-c-decays/Makefile.am
```diff
@@ -0,0 +1,44 @@
+CLEANFILES = *~
+MAINTAINERCLEANFILES = Makefile.in
+
+AM_CXXFLAGS = @AM_CXXFLAGS@
+AM_LDLAGS = @AM_LDFLAGS@
```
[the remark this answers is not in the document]
dvandyk: Will be fixed ina separate PR across the code base.

dvandyk (approved):
Thank you @DominikSue, I think this can be merged now. @mreboud Any objections?

mreboud:
> Thank you @DominikSue, I think this can be merged now. @mreboud Any objections?

Not from my side! All good, thanks @DominikSue!

PR #1163, opened 2026-06-08
Bug fix watermark

DominikSue:
Fix two small bugs about watermark option in `GridFigure`
1.  Remove watermark if plot is empty
2. Fix missing constructor in 'from_dict' if not default values are used.


DominikSue:
Fixed the grammatical error. Other suggested changes by Copilot are not necessary I think.

mreboud (approved):
Looks good. I agree that Copilot's comments are not relevant here.

dvandyk (approved):

PR #1168, opened 2026-06-11
Switch to using ``qnp::OptionValue`` in anything related to options

dvandyk:
- eae196c59ace78633f2dc3139990d1f7f3b2b8b7 Add default comparison operator for ``qnp::OptionValue``.
- 05e2a89af2626028bcac8205a06f489554728702 Add output operator for ``qnp::OptionValue`` to preserve usage of ``join()`` for containers of ``qnp::OptionValue``.
- 1ecf092ad958ed372786dc429ab0a39fd508d327 Allow additional characters ('|', '(', ')', '*') in ``qnp::OptionValue``. All of these characters are already being used in the code base.
- 92770f3f65d0a7c893545aa8b69c46562d3c6e97 Replace the character ``,`` in an option value used in the implementation of ``P->gamma`` form factor options. Replacement reads ``"partial-soft-tw-3,4" -> "partial-soft-tw-3+4"``.
- 333f7328b91437abab5d7b1e1327fab340455d35 Make the switch and use ``qnp::OptionValue`` in lieu of ``std::string`` in everything related to Options.
- c8b2fe31152320686f76ad68c99bb07af8a4ced0 Update the changelog.


eos/utils/options-impl.hh
```diff
@@ -115,11 +116,11 @@ namespace eos
 
                 if (std::find(allowed_values.begin(), allowed_values.end(), default_value) == allowed_values.end())
                 {
-                    throw InternalError("SwitchOption: The default value '" + default_value + "'is not in the list of allowed values: '"
+                    throw InternalError("SwitchOption: The default value '" + default_value.str() + "'is not in the list of allowed values: '"
                                         + join(allowed_values.begin(), allowed_values.end()) + "'");
```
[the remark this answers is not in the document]
dvandyk: Fixed.

eos/utils/options.hh
```diff
@@ -93,13 +94,13 @@ namespace eos
...
-            const std::string & operator[] (const qnp::OptionKey & key) const;
+            const qnp::OptionValue & operator[] (const qnp::OptionKey & key) const;
 
             bool has(const qnp::OptionKey & key) const;
 
-            void declare(const qnp::OptionKey & key, const std::string & value = "");
+            void declare(const qnp::OptionKey & key, const qnp::OptionValue & value);
 
-            std::string get(const qnp::OptionKey & key, const std::string & default_value = "") const;
+            qnp::OptionValue get(const qnp::OptionKey & key, const qnp::OptionValue & default_value) const;
```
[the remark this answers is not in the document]
dvandyk: This is a false positive, since ``qnp::OptionValue`` does not allow empty value. Hence, we have to get rid of the empty default value ("") used here.
This also means that I had to replace all occurrences of
```c++
std::string s = o.get("key"_ok, "");
```
with
```c++
std::string s = o.has("key"_ok) ? o["key"_ok].str() : "";
```
Marking this comment as resolved.

eos/b-decays/b-to-3l-nu.cc
```diff
@@ -68,8 +68,8 @@ namespace eos
         static const std::vector<OptionSpecification> options;
 
         Implementation(const Parameters & p, const Options & o, ParameterUser & u) :
-            model(Model::make(o.get("model"_ok, "SM"), p, o)),
-            form_factors(FormFactorFactory<PToGammaOffShell>::create("B->gamma^*::" + o.get("form-factors"_ok, "KKvDZ2022"), p, o)),
+            model(Model::make(o.get("model"_ok, "SM"_ov).str(), p, o)),
```
mreboud: Why is this needed? Shouldn't the model constructor take an option value as input?

eos/b-decays/b-to-3l-nu.cc
```diff
@@ -864,8 +864,8 @@ namespace eos
     {
         Model::option_specification(),
         FormFactorFactory<PToGammaOffShell>::option_specification(),
-        { "lprime"_ok, { "e", "mu", "tau" }, "mu" },
-        { "l"_ok, { "e", "mu", "tau" }, "e" },
+        { "lprime"_ok, { "e"_ov, "mu"_ov, "tau"_ov }, "mu"_ov },
```
mreboud, line 867: Can you remind me why we don't use an implicit cast from strings in the option constructors?

eos/utils/qualified-name.cc
```diff
@@ -137,11 +137,11 @@ namespace eos
                 throw QualifiedNameSyntaxError("A qualified name's option value part must not be empty");
             }
 
-            // VALUE := ['a'-'z', 'A'-'Z', '0'-'9', '+', '-', '/', '.', '^', '_']
+            // VALUE := ['a'-'z', 'A'-'Z', '0'-'9', '+', '-', '/', '.', '^', '_', '|', '(', ')', '*']
```
mreboud, line 140: Why is star needed? Aren't we using `star` on purpose?

mreboud (approved):
I couldn't read all of the changes, but it looks good overall. I am just a bit afraid of having to add `_ov` and `.str()` in so many places.

eos/b-decays/b-to-3l-nu.cc
dvandyk: This ensure that all option values created at startup are valid option values. Put differently, as soon as we introduce one invalid option value, e.g.
```c++
{
  { "lprime"_ok, { "e,"_ov } }
}
```
we will see a run-time error when loading the library. No test case needed.

eos/utils/qualified-name.cc
dvandyk: https://github.com/eos/eos/blob/c8b2fe31152320686f76ad68c99bb07af8a4ced0/eos/b-decays/b-to-vec-l-nu.cc#L362

mreboud: Ah yes of course we use *, I am an idiot sorry

eos/b-decays/b-to-3l-nu.cc
dvandyk: Hm, right now it does not. It would limit what we could pass to it. I am not opposed to it, but I would prefer to discuss it first and then potentially adapt to that in a separate PR.

dvandyk:
> I couldn't read all of the changes, but it looks good overall. I am just a bit afraid of having to add `_ov` and `.str()` in so many places.

Using
```
std::string value = o["somekey"_ok].str()
```
Does not incur any overhead. An ``OptionValue`` is only a wrapper around ``std::string``. Using ``ov.str()``, you obtain a ``const std::string &`` to the underlying string.

mreboud:
> > I couldn't read all of the changes, but it looks good overall. I am just a bit afraid of having to add `_ov` and `.str()` in so many places.
> 
> Using
> 
> ```
> std::string value = o["somekey"_ok].str()
> ```
> 
> Does not incur any overhead. An `OptionValue` is only a wrapper around `std::string`. Using `ov.str()`, you obtain a `const std::string &` to the underlying string.

Yes, yes, I know. But it is still less transparent than just providing strings for a fresh developer.

dvandyk:
> > > I couldn't read all of the changes, but it looks good overall. I am just a bit afraid of having to add `_ov` and `.str()` in so many places.
> > 
> > Using
> > ```
> > std::string value = o["somekey"_ok].str()
> > ```> >     
> >   
> > Does not incur any overhead. An `OptionValue` is only a wrapper around `std::string`. Using `ov.str()`, you obtain a `const std::string &` to the underlying string.
> 
> Yes, yes, I know. But it is still less transparent than just providing strings for a fresh developer.

I have modified the PR. Now ``Model::make()`` uses a ``qnp::OptionValue`` as its first arguments. Moreover, the internal table has been converted to use``qnp::OptionValue`` as key as well. This avoids problems such as not being able to create models whose keys are not valid option values.

PR #1175, opened 2026-06-17
Alter times of coverage and weekly reports

### LINKED ISSUE #1173: Delayed scheduled workflows
The scheduled workflows (coverage @ 5:05am, weekly builds @ 6:05am, weekly reports @ 7:05am) are frequently not run until much after the scheduled time - previously 5-10 minutes, but recently an hour or more late.

This is apparently a known issue - see discussion https://github.com/orgs/community/discussions/156282, particularly https://github.com/orgs/community/discussions/156282#discussioncomment-17177966, basically Github is overwhelmed.

For now all we can do is move everything earlier (and there may also be some small benefit to picking an even more off peak time than 5 past the hour) - say 1:xx am and 4:xx am for coverage and weekly report respectively? Weekly builds not so necessary as these we don't need for the dev meeting.

MJKirk:
Shift coverage and weekly report workflows earlier so they are hopefully always done by the weekly meeting.

Closes #1173

### DIFF
--- a/.github/workflows/coverage.yaml
+++ b/.github/workflows/coverage.yaml
@@ -1,8 +1,8 @@
 on:
-  # Run once weekly, 2h before the ``Report Activity`` workflow
-  # (which runs at '5 7 * * TUE').
+  # Run once weekly, 3h before the ``Report Activity`` workflow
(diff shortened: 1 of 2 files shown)


dvandyk (approved):
Looks good!

PR #1201, opened 2026-07-11
Improve user experience when using analysis files

### LINKED ISSUE #1181: ``eos-analysis run``: emit warning rather than errors when figure files are missing
Scenario: analysis file with included figures, which use the outcome of ``find-mode``.

The ``AnalysisFile`` loader eagerly validates every figure's ``fixed_parameters_from_file`` at startup. So this file must already exist for ``eos-analysis run`` to even load the analysis file. This is counterproductive...

dvandyk:
- 772445ad [python] Add a ``format_version`` field to the analysis file format. The default format is now ``1``.
- 610195fc [python] Move ``AnalysisFileContext`` into its own module. This breaks the potential for circular dependencies.
- 6bcb66ff [python] Add ``AnalysisFileDescription`` and make use of it in ``AnalysisFile.__init__``. Unknown top-level keys now raise an exception.
- 5b8e9d9b [python] Fix ``GridFigure.draw`` to forward context to each plot. The previous behaviour deviated from other figures and prohibited the easy solution to #1181 in the next commit.
- 15b2294f [python] Defer figure parameter-file loading to draw time. Resolves #1181.
- e13a26d8 [eos] Update changelog for PR #1201.

### DIFF
--- a/python/eos/analysis_file.py
+++ b/python/eos/analysis_file.py
@@ -24,11 +24,8 @@ import yaml
 import inspect
 from collections import Counter
 from dataclasses import asdict
-from eos.analysis_file_description import PriorComponent, LikelihoodComponent, PosteriorDescription, \
-                                       PredictionDescription, ObservableComponent, ParameterComponent, \
-                                       StepComponent, PriorDescription, MaskComponent, MaskExpressionComponent, MaskNamedComponent, \
-                                       MetadataDescription
-from eos.figure import FigureFactory
+from eos.analysis_file_description import AnalysisFileDescription, PriorDescription, \
+                                       MaskExpressionComponent, MaskNamedComponent
 
 # The highest analysis file format version understood by this version of EOS. Increment this
 # whenever a change to the schema alters how existing files are interpreted. Files that omit the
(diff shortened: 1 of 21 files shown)


python/eos/analysis_file.py
```diff
@@ -51,26 +53,35 @@ def __init__(self, analysis_file):
...
-        else:
-            self._metadata = MetadataDescription()
+        # The top-level 'format_version' records the schema version of the analysis file. Files that
+        # omit it predate format versioning and are treated as version 1. Reject files that declare a
+        # newer version than this EOS understands, rather than silently misinterpreting them.
+        self.format_version = self.input_data.get('format_version', 1)
+        if self.format_version > SUPPORTED_FORMAT_VERSION:
+            raise RuntimeError(f'Cannot load analysis file: it declares format version '
+                               f'{self.format_version}, but this version of EOS supports at most '
+                               f'{SUPPORTED_FORMAT_VERSION}; please update EOS')
```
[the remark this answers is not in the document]
dvandyk: Fixed.

python/eos/analysis_file_context.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/python
```
mreboud, line 1: Hmm do we need these added comments?

mreboud (approved):
Looks good :)

python/eos/analysis_file_context.py
dvandyk: The shebang at the top of the file is not needed but recommended.

PR #1218, opened 2026-08-03
Adding options to figure framework

DominikSue:
I add two optional arguments to items of the figure framework.

1. Level option to UncertaintyBandItem to change the credibility interval that is plotted
2. Band option to BinnedUncertaintyItem similar to the band option in UncertaintyBandItem

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -90,6 +90,8 @@
 - Detect newly introduced C++ compiler warnings in the CI/CD workflows using ``g++``'s SARIF diagnostics output, providing native code-scanning annotations on pull requests and failing the build on any warning not present in a committed baseline (D. van Dyk)
 - Add ``eos.LogPrior.External``, which wraps a user-supplied Python factory (providing ``evaluate()``, ``sample()``, and ``compute_cdf()`` callables) as a prior (D. van Dyk)
 - Add the means to ensure a minimum effective sample size for the final samples produced by ``sample-nested``, exposed through the ``eos-analysis`` command-line interface (D. van Dyk)
+- Add an optional ``levels`` field to ``UncertaintyBandItem`` and ``BinnedUncertaintyItem`` items that modifies the plotted credibility interval (D. Suelmann)
+- Add an optional ``band`` field to ``BinnedUncertaintyItem`` items to plot different parts of the band separately (D. Suelmann)
 
 ### Deprecated
 
--- a/python/eos/figure/item.py
+++ b/python/eos/figure/item.py
@@ -841,6 +841,9 @@ class UncertaintyBandItem(Item):
             if band_type not in ['area', 'outer', 'median']:
                 raise ValueError(f"Unrecognized band type '{band_type}'; must be one of 'area', 'outer', or 'median'")
 
+        if len(self.band) == 0:
+            raise ValueError("Parameter 'band' must contain at least one of 'area', 'outer', or 'median'")
+
         if self.interpolation not in ['linear', 'cubic']:
             raise ValueError(f"Unrecognized interpolation type '{self.interpolation}'; must be either 'linear' or 'cubic'")
 
@@ -990,6 +993,11 @@ class BinnedUncertaintyItem(Item):
 
     This routine expects the uncertainty propagation to have produced a data file.
 
+    :param band: A set of strings that determines which parts of the uncertainty band to draw. Can be any combination of ``'area'``, ``'outer'``, or ``'median'``.
+        Defaults to ``{'area', 'outer', 'median'}``.
+        ``'area'`` fills the area between the lower and upper bounds of the uncertainty band,
+        ``'outer'`` draws the outer lines of the uncertainty band, and ``'median'`` draws the median line of the uncertainty band.
+    :type band: set[str] | list[str] | str
     :param datafile: The path to an existing data file of type :class:`eos.data.Prediction` that contains the uncertainty estimates.
     :type datafile: str
     :param levels: The credibility levels that shall be visualized in percent (optional) ordered from lowest to highest. Defaults to [68.27].
@@ -1020,6 +1028,7 @@ class BinnedUncertaintyItem(Item):
         figure.draw()
     """
 
+    band:set[str]|list[str]|str=field(default_factory=lambda : {'area', 'outer', 'median'})
     datafile:str
     levels:list[float]=field(default_factory=lambda: [68.27])
     range:tuple[float, float]|None=field(default=None)
@@ -1028,6 +1037,20 @@ class BinnedUncertaintyItem(Item):
 
     def __post_init__(self):
         super().__post_init__()
+
+        if isinstance(self.band, str):
+            self.band = {self.band}
+        elif isinstance(self.band, list):
+            self.band = set(self.band)
+        elif not isinstance(self.band, set):
+            raise TypeError(f"Parameter 'band' must be a string, list of strings, or a set of strings, not {type(self.band).__name__}")
+
+        for band_type in self.band:
+            if band_type not in ['area', 'outer', 'median']:
+                raise ValueError(f"Unrecognized band type '{band_type}'; must be one of 'area', 'outer', or 'median'")
+
+        if len(self.band) == 0:
+            raise ValueError("Parameter 'band' must contain at least one of 'area', 'outer', or 'median'")
 
         for level in self.levels:
             if level <= 0 or level >= 100:
@@ -1065,8 +1088,8 @@ class BinnedUncertaintyItem(Item):
     def draw(self, ax):
         """Draw the binned uncertainty band on the provided axes.
 
-        For each bin, the central 68% interval and the median are drawn as a filled box with
-        outer and median lines, optionally rescaled by the inverse of the bin width.
+        Depending on the ``band`` setting, fills the area between the lower and upper bounds,
+        draws the outer lines, and/or draws the median line, optionally all rescaled by the inverse of the bin width.
 
         :param ax: The matplotlib axes onto which the band is drawn.
         :type ax: matplotlib.axes.Axes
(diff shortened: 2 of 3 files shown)


python/eos/figure/item.py
```diff
@@ -496,6 +496,7 @@ class UncertaintyBandItem(Item):
     band:set[str]|list[str]|str=field(default_factory=lambda : {'area', 'outer', 'median'})
     datafile:str
     interpolation:str=field(default='linear')
+    level:float|None=field(default=0.6827)
```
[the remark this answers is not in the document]
dvandyk: Same line: please drop ``|None``. You do not test for it, and it providing ``level=None`` actually crashes the code.

python/eos/figure/item.py
```diff
@@ -496,6 +496,7 @@ class UncertaintyBandItem(Item):
     band:set[str]|list[str]|str=field(default_factory=lambda : {'area', 'outer', 'median'})
     datafile:str
     interpolation:str=field(default='linear')
+    level:float|None=field(default=68.27)
```
dvandyk: Drop ``|None``. Code crashes when this is provided.

python/eos/figure/item.py
```diff
@@ -522,6 +523,7 @@ class UncertaintyBandItem(Item):
         If ``'median'`` is provided, the band's median line is drawn.
         Defaults to ``{'area', 'outer', 'median'}``.
         * ``interpolation`` (*str*) -- The type of interpolation to be used for the band. Can be either ``linear`` (default) or ``cubic``.
+        * ``level`` (*float*) -- The credibility level that shall be visualized in percent (optional). Defaults to 68.27%.
```
dvandyk: Uff. This ``_api_doc`` block needs to go. You added it absolutely correctly, but in the wrong position. That's on me / the existing code base. Sorry. Can you move it to line 464 instead?

python/eos/figure/item.py
```diff
@@ -639,7 +643,7 @@ def prepare(self, context:AnalysisFileContext=None):
         _ovalues_lower   = []
         _ovalues_central = []
         _ovalues_higher  = []
-        INTERVAL = [0.15865, 0.5, 0.84135]  # central 68% interval
+        INTERVAL = [0.5-self.level/100/2, 0.5, 0.5+self.level/100/2]  # central 68% interval
```
dvandyk: The comment is now incorrect. Please remove it.

python/eos/figure/item.py
dvandyk: Also, add your name to the copyright statement at the top of the file please.

python/eos/figure/item.py
```diff
@@ -780,10 +800,16 @@ def draw(self, ax):
...
+                label = None # do not label anything else if we fill the band area
+            if 'outer' in self.band:
+                ax.plot([xmin, xmax], [olo,      olo],                              alpha=self.alpha, color=self.color, label=label, lw=self.linewidth, ls=self.linestyle)
+                ax.plot([xmin, xmax], [ohi,      ohi],                              alpha=self.alpha, color=self.color,              lw=self.linewidth, ls=self.linestyle)
+                label = None # do not label anything else if we plot the outer lines
+            if 'median' in self.band:
+                ax.plot([xmin, xmax], [ocentral, ocentral],                         alpha=self.alpha, color=self.color, label=label, lw=self.linewidth, ls=self.linestyle)
             label = None
 
     def legend(self):
```
dvandyk, line 1128: You did not adjust ``legend``. It should provide a handle matching what you plot. So:
- a rectangle with border and a central line (if all 3 options are present),
- a borderless rectangle (if only ``area`` is provided),
- only a line (if only ``median`` is provided).

Happy to discuss this in more detail via Discord.

python/eos/figure/item.py
dvandyk: This docstring is now incorrect.

python/eos/figure/item.py
```diff
@@ -724,11 +729,26 @@ class BinnedUncertaintyItem(Item):
...
+        super().__post_init__()
+
+        if isinstance(self.band, str):
+            self.band = {self.band}
+        elif isinstance(self.band, list):
+            self.band = set(self.band)
+        elif not isinstance(self.band, set):
+            raise TypeError(f"Parameter 'band' must be a string, list of strings, or a set of strings, not {type(self.band).__name__}")
+
+        for band_type in self.band:
```
dvandyk, line 1048: You do not test for ``self.band`` being the empty set. Please do.

python/eos/figure/item.py
```diff
@@ -724,11 +729,26 @@ class BinnedUncertaintyItem(Item):
         figure.draw()
     """
 
+    band:set[str]|list[str]|str=field(default_factory=lambda : {'area', 'outer', 'median'})
     datafile:str
```
dvandyk, line 1032: The binned item does not provide level. Would it make sense to modify the first commit to add ``level`` to both  item types.

python/eos/figure/item.py
```diff
@@ -696,6 +696,11 @@ class BinnedUncertaintyItem(Item):
 
     This routine expects the uncertainty propagation to have produced a data file.
 
+    :param band: A set of strings that determines which parts of the uncertainty band to draw. Can be any combination of ``'area'``, ``'outer'``, or ``'median'``.
```
dvandyk, line 996: This commit is attributed to ``DominikSue`` instead of ``Dominik Suelmann``. Worthwhile to check your global git config and ``commit --amend --reset-author``.

python/eos/figure/item_TEST.py
```diff
@@ -212,6 +212,7 @@ def test_full(self):
             type: uncertainty
             label: '$\\ell=\\mu$'
             variable: 'q2'
+            level: 95
```
dvandyk: I would prefer if the contents of this commit were distributed across the first two commits in this PR. Let me know if this is too difficult though. It's a preference, not a hard requirement.

python/eos/figure/item_TEST.py
```diff
@@ -1163,6 +1168,30 @@ def test_invalid(self):
 
 class BinnedUncertaintyItemValidationTests(unittest.TestCase):
 
+    def test_band_normalization(self):
```
dvandyk, line 1427: Worthwhile testing that an empty set and an empty list is rejected with a ``ValueError``.

python/eos/figure/item_TEST.py
```diff
@@ -1153,6 +1154,10 @@ def test_invalid(self):
         with self.assertRaises(ValueError):
             eos.figure.ItemFactory.from_yaml("type: uncertainty\ndatafile: 'x'\ninterpolation: 'quadratic'")
 
+        # a level value larger than 100 is rejected
+        with self.assertRaises(ValueError):
```
dvandyk: ``assertRaisesRegex(..., "not in the interval")`` is a more specific test. There is a large number of possible ``ValueError`` exceptions that can be raised in this code path.

python/eos/figure/item_TEST.py
```diff
@@ -1163,6 +1168,30 @@ def test_invalid(self):
...
+
+        # a list of band types is normalized to a set
+        item = eos.figure.ItemFactory.from_yaml("type: uncertainty-binned\nvariable: 'nonexistent'\n"
+                                                "datafile: 'eos/data/prediction_TEST.d/predictions-binned'\nband: ['area', 'outer']")
+        self.assertEqual(item.band, {'area', 'outer'})
+
+    def test_invalid(self):
+
+        # a 'band' that is neither a string, list, nor set is rejected
+        with self.assertRaises(ValueError):
```
dvandyk: ``assertRaisesRegex(ValueError, "must be a string, list of strings")``

python/eos/figure/item_TEST.py
```diff
@@ -1163,6 +1168,30 @@ def test_invalid(self):
...
 
+    def test_band_normalization(self):
+
+        # a single band type given as a string is normalized to a set
+        item = eos.figure.ItemFactory.from_yaml("type: uncertainty-binned\nvariable: 'nonexistent'\n"
+                                                "datafile: 'eos/data/prediction_TEST.d/predictions-binned'\nband: 'median'")
+        self.assertEqual(item.band, {'median'})
+
+        # a list of band types is normalized to a set
+        item = eos.figure.ItemFactory.from_yaml("type: uncertainty-binned\nvariable: 'nonexistent'\n"
```
dvandyk, line 1435: Your modified test case does not check the full construct -> prepare -> draw pipeline for a non-default ``band`` argument.

Changelog.md
```diff
@@ -89,6 +89,8 @@
 - Detect newly introduced C++ compiler warnings in the CI/CD workflows using ``g++``'s SARIF diagnostics output, providing native code-scanning annotations on pull requests and failing the build on any warning not present in a committed baseline (D. van Dyk)
 - Add ``eos.LogPrior.External``, which wraps a user-supplied Python factory (providing ``evaluate()``, ``sample()``, and ``compute_cdf()`` callables) as a prior (D. van Dyk)
 - Add the means to ensure a minimum effective sample size for the final samples produced by ``sample-nested``, exposed through the ``eos-analysis`` command-line interface (D. van Dyk)
+- Add an optional ``level`` field to ``UncertaintyBandItem`` items that modifies the plotted credibility interval (D. Suelmann)
```
dvandyk: As mentioned before, ``level`` would be a nice addition to ``BinnedUncertaintItem`` as well.

dvandyk (changes requested):
@DominikSue Thank you. Despite all of my niggling comments: this looks like a good addition. Happy to discuss further via Discord.

python/eos/figure/item.py
DominikSue: I think this should be the only somewhat open comment. Only the first point is not fulfilled currently as no central line is drawn. I think for this I need to modify CompositeRegionHandler right? Should I do this in another commit? Also not yet completely sure how to do this.

dvandyk:
Two test case failures are due to a bug in how we handle PRs from outside repos. Merging now.

PR #1235, opened 2026-09-01
Fix various multithreading problems and race conditions in nested C++/Python operations

dvandyk:
- 13d8a7dd [utils] Add ``ThreadPool::WaitGuard`` and use it in multithreaded paths.
- b524f53c [python] Make the Python bindings thread-safe when using nested Python/C++ calls.
- a7b382d2 [utils] Fix a dangling capture in the log test.
- 293d1ce0 [utils] Notify the log's callbacks without holding its lock.
- 91ed66c7 [python] Own the log callbacks' references and notify them under the GIL.
- f778e628 [utils] Bounds-check the id that a parameter name maps to.
- d4d5bd58 [utils] Keep the parameter that a repeated declaration names.
- e6d7407e [utils] Fix the double-checked locking in the singleton accessor.
- 9350a993 [python] Acquire the GIL for the external prior and likelihood block.
- 8dd496c6 [utils] Drop needless captures in the log test.
- 5a2e7830 [utils] Synchronise the log test's message collector.
- da28fc87 [eos] Update changelog for PR #1235.

### DIFF
--- a/eos/utils/log_TEST.cc
+++ b/eos/utils/log_TEST.cc
@@ -22,6 +22,11 @@
 
 #include <test/test.hh>
 
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+#include <thread>
+
 using namespace test;
 using namespace eos;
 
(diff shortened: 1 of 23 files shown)


eos/utils/observable_cache.cc
```diff
@@ -227,6 +227,9 @@ namespace eos
     void
     ObservableCache::update()
     {
+        // an observable may call back into the embedding runtime from one of the pool's threads
+        const auto guard = ThreadPool::instance()->wait_guard();
+
```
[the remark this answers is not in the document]
dvandyk: This is only true (at least in ``g++`` and ``clang``) if the variable has a trivial destructor, explicitly exempting objectes used for RAII (like here).

eos/utils/log_TEST.cc
```diff
@@ -150,14 +155,16 @@ class LogOneTimeMessageTest : public TestCase
...
         run() const
         {
-            std::vector<std::tuple<std::string, LogLevel, std::string>> messages;
+            // the callback outlives this call, so it must not capture anything local to it
 
             // register callback
             std::function<void(const std::string &, const LogLevel &, const std::string &)> callback =
-                    [&messages](const std::string & id, const LogLevel & level, const std::string & message) { messages.push_back(std::make_tuple(id, level, message)); };
+                    [this](const std::string & id, const LogLevel & level, const std::string & message) { messages.push_back(std::make_tuple(id, level, message)); };
             Log::instance()->register_callback(callback);
```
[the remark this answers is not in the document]
dvandyk: Valid criticism. Fixed.

eos/utils/parameters.cc
```diff
@@ -303,6 +303,12 @@ namespace eos
 
             std::vector<ParameterSection> _sections;
 
+            static bool
+            _identical_declarations(const Parameter::Template & lhs, const Parameter::Template & rhs)
+            {
+                return (lhs.name == rhs.name) && (lhs.min == rhs.min) && (lhs.central == rhs.central) && (lhs.max == rhs.max) && (lhs.latex == rhs.latex) && (lhs.unit == rhs.unit);
```
mreboud, line 309: I understand that accessing all the attributes one by one with `&&` is faster, but it might be hard to maintain if we modify the `Template` structure. Wouldn't it be worth iterating over the attributes instead?

mreboud (commented):
LGTM, I only have one comment :)

eos/utils/parameters.cc
dvandyk: As in
```C++
if (lhs.name != rhs.name)
    return false;
if (lhs.min != rhs.min)
    return false;
// ...
```
?

dvandyk: @mreboud was referring to reflections to synthesize an ``operator==``. However, that is only available / possible starting with ``-std=c++26``, not our ``-std=c++20``. Keeping this as is.

PR #1238, opened 2026-09-11
Remove outputs from existing example notebooks and prohibit future commits to store stale outputs

dvandyk:
- d244b178 [examples] Clear the stale output from the inference notebook.
- 87c369d6 [examples] Clear the stored output from the external-observable notebook.
- 556488e8 [infra] Reject notebooks carrying stored output.


mreboud (approved):
Thanks @dvandyk!

PR #1244, opened 2026-09-14
Remove $ signs for parameters

### LINKED ISSUE #1110: Inconsistent use of $ for LaTeX in definition of paramers and observables
In https://eoshep.org/doc/reference/analysis-file-format.html#observables we give this example:
```yaml
observables:
  'B->pi::f_+(q2)/f_0(q2)':
    latex: '$\frac{f_+}{f_0}$'
    unit: '1'
    options: {}
    expression:
      '<<B->pi::f_+(q2)>> / <<B->pi::f_0(q2)>>'
```
but for observables EOS expects to add the dollar signs for latex itself so we shouldn't put them here.
This affects e.g. plotting predictions of these custom observables because the label is wrong

mreboud:
Closes #1110


mreboud:
I wanted to add an error when parameters or observables are provided with invalid '$' signs, but encountered another bug when doing so: this line is actually causing a segfault when caught...
https://github.com/eos/eos/blob/4b0a4ea097599fc3bcfef52eea57a6c883ab164b/eos/utils/parameters.cc#L541-L544

eos/utils/parameters_TEST.d/bad-node.yaml
```diff
@@ -0,0 +1,8 @@
+# Fixture for ParametersTest: an incorrect parameter node.
+# Parameters::override_from_file() rethrows the resulting node error as a
+# ParameterInputFileNodeError.
+test::bad:
+    min: 1.0
+    max: 2.0
+    central: 1.5
+    latex: ["e^{i\\pi}"]
```
dvandyk, line 8: Please move this to a separate commit.

eos/utils/parameters_TEST.d/bad-node.yaml
```diff
@@ -0,0 +1,8 @@
+# Fixture for ParametersTest: an incorrect parameter node.
+# Parameters::override_from_file() rethrows the resulting node error as a
+# ParameterInputFileNodeError.
```
dvandyk, line 3: It raises ParameterFileParseError, wrapping a ParameterFileNodeError.

dvandyk (changes requested):
Rebase plus two nitpicky things, otherwise approved! Merci bien @mreboud !

dvandyk (approved):

PR #1251, opened 2026-09-15
Implement a generation index for `eos.Parameters`

### LINKED ISSUE #1241: Generation index for `eos.Parameters`
Currently, two observables that differ in their kinematics but share common (potentially hard-to-compute) objects are recomputed from scratch. Having a way of checking if the Parameters were changed would allow us not to reevaluate these objects and therefore optimise evaluation.

mreboud:
Github: Closes #1241


dvandyk:
@mreboud approved verbally after hours; merging.

PR #1272, opened 2026-09-18
Use more sensible watermarking for grid figures

dvandyk:
- 94724078 [python] Stamp the watermark on a single plot of `GridFigure` by default.
- e59ab164 [python] Accept the watermark options in a `CornerFigure`.
- 69998d72 [eos] Update changelog for PR #1272.


mreboud:
Looks good. Could you also add this option to `CornerFigure`?
Thanks a lot!

MJKirk (approved):

PR #1275, opened 2026-09-18
Remove the polarized Lambda_c saturation observables

dvandyk:
Breaking: the Lambda_c->Lambda and Lambda_c->neutron ``Saturation[...,0|perp|para]`` observables are gone; use the polarization-summed ``Saturation[...]`` instead, as the mesonic transitions and the Lambda_b ones already do.


mreboud (approved):
oops, thanks!

PR #1278, opened 2026-09-21
Add further checks on latex -> html conversion

### LINKED ISSUE #1027: LaTeX ->HTML for reference titles
Somewhere in the conversion chain latex -> reST -> html, some of the maths in reference titles gets messed up: do a ctrl-f for `:math:` on https://eoshep.org/doc/reference/bibliography.html.
I think at least one issue is that in reST math markup, the first and/or last character inside the backtick shouldn't be a space (from my reading of https://docutils.sourceforge.io/docs/ref/rst/roles.html and https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules)
It also doesn't seem to like "free" superscripts e.g. `$^{*}$` in the latex, although I can't see a rule that says this.

It would be nice to have some way of testing for these issues - at the latest we can just search for any remaining `:math:` in the HTML documentation, but maybe something earlier is possible.

Also it should be possible to test for some of the problems in 5cde3a5, e.g. checking for an even number of $ in the string, an equal number of opening and closing braces

MJKirk:
Adds checks on the reference title latex, and checks the generated documentation for any lingering reST `:math:` markup that signals a missed failure.

In 0d59299190a7bc222e7867963efc0483d9c52f14, I made the tests collect all failures and output a list. As noted, I cut the full latex error output. I think this is a worthwhile trade-off to get all failures at once, but this could be reverted.

Currently we have two errors, related to the parameters `Lambda_b::polarisation@unpolarised` and `Lambda_b::polarisation@LHCb`, as these don't have latex representations. The failure is actually a side-effect of PR #1244, since we now force the parameter latex to be a `$...$` string, and the reST -> html conversion doesn't deal well with empty ```:math:`` ``` notation.

Otherwise I think this is satisfactory to close #1027 

Another note - these latex tests contribute a non-trivial amount of time (around 8 minutes) to the checks, since we are individually rendering almost 6000 individual latex strings. If you are doing it locally, it is a lot faster for future instances since matplotlib caches these, but obviously this doesn't help with the github checks.


mreboud:
> Currently we have two errors, related to the parameters `Lambda_b::polarisation@unpolarised` and `Lambda_b::polarisation@LHCb`, as these don't have latex representations. The failure is actually a side-effect of PR #1244, since we now force the parameter latex to be a `$...$` string, and the reST -> html conversion doesn't deal well with empty `` :math:`` `` notation.

Yes, that's not super satisfactory. I wouldn't like to enforce non-empty latex strings for Observables and Parameters, but we can always catch those when testing or using them. This is what I have done here https://github.com/eos/eos/blob/03db721aa7154380ea62d046c94f28965e533c2e/doc/doc_TEST.py#L24-L26

MJKirk:
> Yes, that's not super satisfactory. I wouldn't like to enforce non-empty latex strings for Observables and Parameters

I would agree with this.

In which case I suggest I add a check in: 
https://github.com/eos/eos/blob/03db721aa7154380ea62d046c94f28965e533c2e/doc/reference/parameters.rst.py#L35
(~and equivalent in the `observables.rst.py` file~ - not needed as this already deals with this special case correctly) so that if the latex string is empty, we don't add put an empty `:math:` markup into the reST.

doc/doc_TEST.py
```diff
@@ -41,7 +44,45 @@ def test_001_Observables(self):
...
+                    texmanager.get_text_width_height_descent(latex_string, fontsize=12)
+                except Exception as e:
+                    bad_references.append(ref_id)
+        if bad_references:
+            self.fail(f"Cannot compile latex representation of the following reference titles: {[str(ref_id) for ref_id in bad_references]}")
+
+    def test_003_math_conversion(self):
+        """Check for any lingering ':math:' in the generated html documentation"""
+        html_directory = os.path.join(os.environ.get("BUILDDIR"), 'html')
+        unconverted_math = []
```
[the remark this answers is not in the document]
MJKirk: Tests are always run through the Makefile, which sets `BUILDDIR` directly even if not set on the command line.

mreboud (approved):
Looks good!

doc/doc_TEST.py
```diff
@@ -43,6 +43,20 @@ def test_001_Observables(self):
                         except Exception as e:
                             self.fail(f"Cannot compile latex representation of observable {qn}, caught exception of type {type(e).__name__}: {e}")
 
+    def test_002_References(self):
+        """ Check the latex representation of all reference titles"""
```
dvandyk: Spurious whitespace at the beginning of the docstring.

doc/doc_TEST.py
```diff
@@ -57,6 +57,25 @@ def test_002_References(self):
                 except Exception as e:
                     self.fail(f"Cannot compile latex representation of reference {ref_id} title, caught exception of type {type(e).__name__}: {e}")
 
+    def test_003_math_conversion(self):
+        """Check for any lingering ':math:' in the generated html documentation"""
+        html_directory = os.path.join(os.environ.get("BUILDDIR"), 'html')
+        unconverted_math = []
+
+        for root, _, files in os.walk(html_directory):
+            for filename in files:
```
dvandyk, line 81: Can you check that this is actually iterating over anything? Maybe
```python
checked_files = []
for root, _, files in os.walk(html_directory):
    for filename in files:
        # perform the checks
    checked_files.append(filename)
# ...
if not checked_files:
    self.fail('Found no HTML files in the documentation build')
```

This is unlikely to happen in the CI workflow, but might happen if I am stupid and run the test without running ``make -C doc html`` first.

doc/doc_TEST.py
```diff
@@ -24,14 +25,17 @@ def test_000_Parameters(self):
                         try:
                             texmanager.get_text_width_height_descent('$' + latex_string + '$', fontsize=12)
                         except Exception as e:
-                            self.fail(f"Cannot compile latex representation of parameter {parameter.name()}, caught exception of type {type(e).__name__}: {e}")
+                            bad_parameters.append(parameter)
+        if bad_parameters:
+            self.fail(f"Cannot compile latex representation of the following parameters: {[p.name() for p in bad_parameters]}")
```
dvandyk: I like it but would also keep the error messages, to avoid having to re-run the ``doc-check`` locally, possibly in a docker container if it is otherwise not reproducable:
```python
                        except Exception as e:
                            bad_parameters.append((parameter.name(), f'{type(e).__name__}: {e}'))
        if bad_parameters:
            self.fail('Cannot compile latex representation of the following parameters:\n'
                      + '\n'.join(f'  {name}: {reason}' for name, reason in bad_parameters))
```

Same should probably be applied to ``test_001_Observables`` and ``test_002_References``.

dvandyk (commented):
Thanks @MJKirk. I left a few comments.

doc/doc_TEST.py
MJKirk: Agreed: write out full LaTeX error, and then fail with the summary list

dvandyk (approved):
Approved, thanks @MJKirk!

